PR #5, opened 2015-01-04
fixed mutation let-aliasing bug

andmkent:
let-aliasing initially forgot to check if a variable that might act as
an alias was ever mutated. fixed by disallowing a mutated variable to
act as an alias.


samth:
Looks good to commit.

Issue #6, opened 2015-01-07
typed/rackunit doesn't work with test cases or test suites


lexi-lambda:
More or less exactly what it says on the tin. The `test-case` and `test-suite` functions don't work properly, giving type errors.

Issue #7, opened 2015-01-07
Needs implementation of typed/rackunit/docs-complete


lexi-lambda:
The [`rackunit/docs-complete`](http://docs.racket-lang.org/rackunit/Testing_Utilities.html#%28mod-path._rackunit%2Fdocs-complete%29) package has no equivalent typed version, and it doesn't work with `require/typed` due to some internal macro issues.

PR #8, opened 2015-01-08
Improve rackunit support with test-case, test-suite, and check-docs

lexi-lambda:
This fixes issues #6 and #7.

The `test-case` macro was extremely simple to reimplement in typed racket, but `test-suite` is a different beast. This reimplements the `test-suite` macro entirely within typed racket, something that required reimplementing the `current-seed` parameter and the `test-suite-test-case-around` and `test-suite-check-around` functions.

The precise type of `Seed` is somewhat unclear from the rackunit code, since it appears to be simply threaded through the code, and it can be anything at all, in theory. However, rackunit itself uses a monad type it defines itself in the `run-tests` function from `rackunit/text-ui`, so this code assumes seeds will be of that type.

### DIFF
--- a/typed-racket-more/typed/rackunit/main.rkt
+++ b/typed-racket-more/typed/rackunit/main.rkt
@@ -1,7 +1,8 @@
 #lang typed/racket
 (require typed/private/utils
          typed/private/rewriter
-         "type-env-ext.rkt")
+         "type-env-ext.rkt"
+         (for-syntax syntax/parse))
 
 (define-type check-ish-ty
   (case-lambda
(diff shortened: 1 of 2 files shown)


samth:
Awesome! Did you check if the Monad type works with the rackunit gui as well?

samth:
@rmculpepper Can you have a look at this as well?

typed-racket-more/typed/rackunit/main.rkt
```diff
@@ -93,22 +94,108 @@
...
 (require/opaque-type TestCase test-case? rackunit)
 (provide TestCase test-case?)
 
+(require/typed
+ rackunit/private/monad
+ [#:opaque monad monad?])
+(define-type Seed (Option monad))
 
+(define-type test-suite-handler-down
+  (rackunit-test-suite (Option String) (Thunk Any) (Thunk Any) Seed -> Seed))
```
samth: Do these Thunks need to accomodate returning multiple values -- ie, using `AnyValues` instead of `Any`?

lexi-lambda:
@samth It appears it does *not* work with the GUI runner, but it doesn't look too terribly difficult to resolve. I'll take a look and try to push a fix soon.

typed-racket-more/typed/rackunit/main.rkt
lexi-lambda: @samth Hmm, that's a good point; probably wouldn't hurt. The rest of the existing code used `(Thunk Any)` everywhere, so I didn't think about it, but `AnyValues` would probably be better.

typed-racket-more/typed/rackunit/main.rkt
```diff
@@ -93,22 +94,108 @@
...
+ [apply-test-suite (rackunit-test-suite
+                    test-suite-handler-down
+                    test-suite-handler-up
+                    test-suite-handler-here
+                    Seed -> Seed)])
+
+(define current-seed : (Parameter Seed)
+  (make-parameter #f))
+
+(: test-suite-test-case-around (test-suite-handler-here -> ((Thunk Any) -> Any)))
```
samth: I would type these as producing `Void`.

typed-racket-more/typed/rackunit/main.rkt
```diff
@@ -93,22 +94,108 @@
...
+
+(: test-suite-check-around (test-suite-handler-here -> ((Thunk Any) -> Any)))
+(define (test-suite-check-around fhere)
+  (lambda (thunk)
+    (let* ([name #f]
+           [test (make-rackunit-test-case name thunk)]
+           [seed (current-seed)])
+      (current-seed (fhere test name thunk seed)))))
+
+(define-syntax (test-suite stx)
```
samth: Add a comment saying that this is copy-pasted, so we know what to refactor later.

lexi-lambda:
@samth Updated with your feedback. I didn't include the `AnyValues` change since I'm unsure how to actually do that, since `AnyValues` doesn't seem to be a valid type at the moment.

The `rackunit/gui` support should work with the change I've made in racket/rackunit#2.

samth:
Merged!

PR #10, opened 2015-01-10
path-type allow non-subtype struct acc

andmkent:
fixes odd corner case internal errors where an alias's path doesn't check as a subtype of the type of the variable.

1) For example error, download tar:
https://www.dropbox.com/s/m0lj8n82z2xwogr/funky-typed.tar?dl=0

2) untar
3) racket funky-typed/synth/synth.rkt

With this commit, the internal error this currently triggers disappears and instead typechecking errors with a type mismatch on Array (this file does some non-standard usage of some of the private internals of the math/array library that doesn't work out). So both break, but this commit ensures it is a user level error for this case (i.e. a readable typechecker error) instead of a misleading internal path-mismatch error.

### DIFF
--- a/typed-racket-lib/typed-racket/env/lexical-env.rkt
+++ b/typed-racket-lib/typed-racket/env/lexical-env.rkt
@@ -7,15 +7,13 @@
 ;; but split here for performance
 
 (require "../utils/utils.rkt"
-         racket/keyword-transform racket/list racket/match racket/set
+         racket/keyword-transform racket/list
          (for-syntax syntax/parse racket/base)
-         (contract-req)
+         (contract-req)                       
          (env type-env-structs global-env mvar-env)
-         (rep object-rep type-rep rep-utils)
          (utils tc-utils)
          (only-in (rep type-rep) Type/c)
          (typecheck renamer)
-         (types subtype resolve union)
          (except-in (types utils abbrev kw-types) -> ->* one-of/c))
 
 (provide lexical-env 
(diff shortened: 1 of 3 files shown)


samth:
Can we make a test for this?

Also, can you use #:when instead of nevermind?

andmkent:
Okay, this initial fix is not right (I'll update the pull req in a sec if my next approach passes all the TR tests).

Here are two files which together produce the error when you run main.rkt:
foo.rkt https://gist.github.com/andmkent/eba07d62b322e2fac3e2
main.rkt (requires foo.rkt) https://gist.github.com/andmkent/11b14797f3b6cb03f717

Someone who is better at modules could probably turn this into a test right quick. I can probably figure out how to with some more tinkering...

Basically, this code (which is a simpler version of the funkytown code) involves a module ("foo.rkt") which exports a macro ("in-foo") which references a struct ("foo") defined in that module also, however that struct ("foo") is *not* exported. When another module ("main.rkt") imports and uses that macro ("in-foo"), it is typechecking code which is doing type-tests and unsafe-struct accesses on a struct it doesn't even have a type for. This would normally error with a type checking error (user level) -- let aliasing got involved because it made some aliases to struct-path objects before the "foo is an unbound type" errors kicked in and instead it hit that internal error (which was annoying! Why didn't that "foo unbound" error get thrown sooner? Then this (probably) wouldn't have been an issue at all!).

andmkent:
So, because the internal error in path-type was getting called before the proper typechecking errors were being reported, I tried returning the "Error" type if path-type is unable to extract a type -- this worked, causing all the code which was triggering the internal error to instead give the proper user-level typechecking errors. All TR test cases are passing.

PR #13, opened 2015-01-20
Improve check-below errors for classes/objects

takikawa:
This PR improves the error messages when check-below reports a subtyping error with `Class` or `Instance` types, which are often very big.

Question to code reviewers: should I move `class-mismatch` to a separate module? Suggestions on which one or where to put the new one?


samth:
I think you should make `expected-but-got` and `class-mismatch` into a new error message module.

takikawa:
Okay, pushed a new commit that moves them. Look ok?

samth:
Looks good.

takikawa:
Merged.

PR #14, opened 2015-01-22
Added occurrence typing for private fields.

takikawa:
Adds occurrence typing for private fields. This has been implemented on an experimental branch for a while. The new thing is that it now comes with mutation checking in the class body to disable occurrence typing when necessary.

Code reviewers: my main concern is that the use of `FieldPE` path elements on the object of the accessor function is too hacky (see code in `tc-envops.rkt`). I can't think of any other way to implement it though.

### DIFF
--- a/typed-racket-lib/typed-racket/rep/object-rep.rkt
+++ b/typed-racket-lib/typed-racket/rep/object-rep.rkt
@@ -16,6 +16,7 @@
 (def-pathelem StructPE ([t Type?] [idx natural-number/c])
   [#:frees (λ (f) (f t))]
   [#:fold-rhs (*StructPE (type-rec-id t) idx)])
+(def-pathelem FieldPE () [#:fold-rhs #:base])
 
 (def-object Empty () [#:fold-rhs #:base])
 
--- a/typed-racket-lib/typed-racket/typecheck/check-class-unit.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/check-class-unit.rkt
@@ -804,7 +804,7 @@
     (for/list ([m (in-set method-names)])
       (define external (dict-ref internal-external-mapping m))
       (define maybe-type (dict-ref type-map external #f))
-      (->* (list (make-Univ))
+      (->* (list Univ)
            (cond [(and maybe-type
                        (not (equal? (car maybe-type) top-func))
                        (not inner?))
@@ -813,7 +813,7 @@
                        (not (equal? (car maybe-type) top-func)))
                   (Un (-val #f)
                       (function->method (car maybe-type) self-type))]
-                 [else (make-Univ)]))))
+                 [else Univ]))))
 
   (define method-types
     (make-method-types (hash-ref parse-info 'method-internals) methods))
@@ -835,10 +835,8 @@
       (define external (dict-ref internal-external-mapping f))
       (define maybe-type (dict-ref type-map external #f))
       (values
-       (-> (make-Univ) (or (and maybe-type (car maybe-type))
-                           (make-Univ)))
-       (-> (make-Univ) (or (and maybe-type (car maybe-type))
-                           -Bottom)
+       (-> Univ (or (and maybe-type (car maybe-type)) Univ))
+       (-> Univ (or (and maybe-type (car maybe-type)) -Bottom)
            -Void))))
 
   (define (make-private-field-types field-names getter-ids type-map)
@@ -850,14 +848,13 @@
         ;; This case is more complicated than for public fields because private
         ;; fields support occurrence typing. The object is set as the field's
         ;; accessor id, so that *its* range type is refined for occurrence typing.
-        (list (make-arr* (list (make-Univ))
+        (list (make-arr* (list Univ)
                          (or (and maybe-type (car maybe-type))
-                             (make-Univ))
+                             Univ)
                          #:filters -no-filter
                          #:object
                          (make-Path (list (make-FieldPE)) getter-id))))
-       (-> (make-Univ) (or (and maybe-type (car maybe-type))
-                           -Bottom)
+       (-> Univ (or (and maybe-type (car maybe-type)) -Bottom)
            -Void))))
 
   (define-values (field-get-types field-set-types)
@@ -878,7 +875,7 @@
       (or (and maybe-type
                (not (equal? maybe-type top-func))
                (function->method maybe-type self-type))
-          (make-Univ))))
+          Univ)))
 
   (define private-method-types
     (make-private-like-types (hash-ref parse-info 'private-names)
@@ -942,7 +939,7 @@
           (append all-types
                   init-types
                   init-rest-type
-                  (list self-type (make-Univ)))))
+                  (list self-type Univ))))
 
 ;; check-methods : Listof<Symbol> Listof<Syntax> Dict<Symbol, Symbol> Dict Type
 ;;                 -> Dict<Symbol, Type>
--- a/typed-racket-lib/typed-racket/typecheck/check-class-unit.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/check-class-unit.rkt
@@ -12,11 +12,11 @@
          syntax/stx
          "signatures.rkt"
          (private parse-type syntax-properties)
-         (env lexical-env tvar-env global-env type-alias-helper)
+         (env lexical-env tvar-env global-env type-alias-helper mvar-env)
          (types utils abbrev union subtype resolve generalize)
          (typecheck check-below internal-forms)
-         (utils tc-utils)
-         (rep type-rep)
+         (utils tc-utils mutated-vars)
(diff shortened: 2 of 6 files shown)


typed-racket-lib/typed-racket/rep/object-rep.rkt
```diff
@@ -16,6 +16,7 @@
 (def-pathelem StructPE ([t Type?] [idx natural-number/c])
   [#:frees (λ (f) (f t))]
   [#:fold-rhs (*StructPE (type-rec-id t) idx)])
+(def-pathelem FieldPE () [#:fold-rhs #:base])
```
andmkent: Can you let-alias an expression with this sort of path? ex: (let ([x exp]) ...) where the object of exp has a FieldPE in its path?

If so, I think a change similar to that in update would need to be added to 
https://github.com/takikawa/typed-racket/blob/priv-field-occ-type/typed-racket-lib/typed-racket/types/path-type.rkt

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
```diff
@@ -56,6 +56,18 @@
         [else (let ([flds* (append lhs (cons (make-fld ty* acc-id #f) (cdr rhs)))])
                 (make-Struct nm par flds* proc poly pred))]))]
 
+    ;; class field ops
+    ;;
+    ;; This is a bit of an abuse of filters/objects. Normally, the object in
+    ;; an expression like (car? x) is the `x' identifier. However, for an
+    ;; application of a field accessor (accessor obj), the id `accessor` is
+    ;; the object. We need this so that we can refine the *range* of the
+    ;; accessor, not the object's type.
```
andmkent: This comment confuses me -- maybe I'm reading it wrong. 

'car?' should be 'car', right? and so we're saying 'car' is the new path element into the object 'x'?

Is `accessor' the object? or the path element added to the object of 'obj'?

takikawa: Yeah, `car?` is a typo. The `accessor` is the name reference in the `Path` object, yes. The path element that's added is `FieldPE`. That's for the object that you get from the application (via the latent object on `accessor`).

typed-racket-lib/typed-racket/rep/object-rep.rkt
takikawa: Yes, I think if you do `(let ([x f]) ...)` where `f` refers to a private field inside the class. I think you're right that I need to add a case there.

takikawa:
I think I've fixed this for let-aliasing now. Thanks for the feedback. It turned out that recent versions of `plot` actually depend on this let-aliasing to work correctly with this PR (which confused me for a while since the `let` only existed in the expansion of a `match` expression).

typed-racket-lib/typed-racket/typecheck/check-class-unit.rkt
```diff
@@ -723,12 +749,31 @@
...
+        ;; fields support occurrence typing. The object is set as the field's
+        ;; accessor id, so that *its* range type is refined for occurrence typing.
+        (list (make-arr* (list (make-Univ))
+                         (or (and maybe-type (car maybe-type))
+                             (make-Univ))
+                         #:filters -no-filter
+                         #:object
+                         (make-Path (list (make-FieldPE)) getter-id))))
+       (-> (make-Univ) (or (and maybe-type (car maybe-type))
+                           -Bottom)
```
samth: When is this `-Bottom`?

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
```diff
@@ -56,6 +56,18 @@
         [else (let ([flds* (append lhs (cons (make-fld ty* acc-id #f) (cdr rhs)))])
                 (make-Struct nm par flds* proc poly pred))]))]
 
+    ;; class field ops
+    ;;
+    ;; This is a bit of an abuse of filters/objects. Normally, the object in
+    ;; an expression like (car x) is the `x' identifier. However, for an
+    ;; application of a field accessor (accessor obj), the id `accessor` is
+    ;; the object. We need this so that we can refine the *range* of the
+    ;; accessor, not the object's type.
```
samth: Just to make sure I understand what's going on here, what appears to be a refinement of the type of a private field is really a refinement of the return type of the accessor function for that field. We can't turn it into a refinement on the type of the argument to the accessor, since that will have some object type that by definition doesn't mention the private field type. Therefore, `FieldPE` refinements are just a marker saying "the thing being refined is a field represented by this accessor function. Furthermore, this is ok because a private field can't be the result of some other accessor (such as `car`).

Is that all correct?

typed-racket-lib/typed-racket/types/path-type.rkt
```diff
@@ -57,6 +57,11 @@
     [((PolyDots: _ body-t) _) (path-type path body-t resolved)]
     [((PolyRow: _ _ body-t) _) (path-type path body-t resolved)]
     
+    ;; for private fields in classes
+    [((Function: (list (arr: doms (Values: (list (Result: rng _ _))) _ _ _)))
```
samth: Don't we ave a simpler pattern expander for this?

typed-racket-lib/typed-racket/utils/mutated-vars.rkt
```diff
@@ -17,8 +19,18 @@
         (loop stx tbl)))
     (syntax-parse stx
       #:literal-sets (kernel-literals)
+      ;; let us care about custom syntax classes
+      [form
```
samth: Why do we need this additional case? It seemed from the earlier discussion like skipping some `set!` cases would be all that's needed here.

samth:
This is in pretty good shape, just a few comments.

typed-racket-lib/typed-racket/typecheck/check-class-unit.rkt
takikawa: If there isn't a type annotation for the field, in which case we don't know how we can mutate the field so it's disallowed. (we can't necessarily just synthesize the field type first, because the field initial value expression may be a method call)

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
takikawa: Yes, that sounds correct. Note that this still works if you have something like `car` around the field access. But you won't have the other way around, like you said.

typed-racket-lib/typed-racket/types/path-type.rkt
takikawa: I don't know off the top of my head, but I will look for it.

typed-racket-lib/typed-racket/utils/mutated-vars.rkt
takikawa: Because we want to know when the field mutator function is called and invalidate occurrence typing in that case.

typed-racket-lib/typed-racket/typecheck/check-class-unit.rkt
```diff
@@ -723,12 +749,31 @@
...
+        ;; This case is more complicated than for public fields because private
+        ;; fields support occurrence typing. The object is set as the field's
+        ;; accessor id, so that *its* range type is refined for occurrence typing.
+        (list (make-arr* (list (make-Univ))
+                         (or (and maybe-type (car maybe-type))
+                             (make-Univ))
+                         #:filters -no-filter
+                         #:object
+                         (make-Path (list (make-FieldPE)) getter-id))))
+       (-> (make-Univ) (or (and maybe-type (car maybe-type))
```
samth: Use `-Univ` instead of allocating and hashing every time with `(make-Univ)`.

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
samth: Right, makes sense. Can you expand the above comment a little along these lines?

typed-racket-lib/typed-racket/utils/mutated-vars.rkt
samth: Invalidate occurrence typing on what? Are we using this to track function calls as well as mutable variables?

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
takikawa: Sure, will do.

typed-racket-lib/typed-racket/utils/mutated-vars.rkt
takikawa: Yes, this is used to add the name of the accessor to the mutated variable environment. Since the case the extra `FieldPE:` case above looks at uses of the accessor name, that prevents refinement of the accessor result.

PR #15, opened 2015-01-22
Improve occurrence typing on Listof types

takikawa:
This PR improves occurrence typing to use `car`/`cdr` information for `Listof` types. It also simplifies the types of operations like `car` and improves type inference a little bit.

Code review concerns: was there any reason this wasn't done before? (unsound maybe?) The new type for `car` seems simpler (only one case instead of two) but is it harder to understand for users?

Also, I haven't actually adjusted the types for all of the `car` variants yet (the changes are straightforward). I'll do that if code review checks out.

[Review on Reviewable](https://reviewable.io/reviews/racket/typed-racket/15)

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -20,7 +20,7 @@
  "base-structs.rkt"
  racket/file
  (only-in racket/private/pre-base new-apply-proc)
- (only-in (types abbrev) [-Boolean B] [-Symbol Sym] -Flat)
+ (only-in (types abbrev) [-Boolean B] [-Symbol Sym] -Flat -lst**)
  (only-in (types numeric-tower) [-Number N])
  (only-in (rep type-rep)
           make-ClassTop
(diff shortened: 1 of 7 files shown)


takikawa:
It turns out the special case that I added initially for this PR was unnecessary. Changing the base environment types is enough.

Unfortunately, the tests don't pass yet because the change to the type of `first` causes this type error:

````
unit-tests/typecheck-tests.rkt:3081:22: Type Checker: Polymorphic function `first' could not be applied to arguments:
Argument 1:
  Expected: (U '() (Pairof a (Listof b)))
  Given:    (List g4803 ... B)
````

which boils down to this inference test failure (a new test that I added):

````
--------------------
All inference tests > Tests for infer > ((U '() (Pairof X (Listof Y)))) ((List A ... B))
((U '() (Pairof X (Listof Y)))) ((List A ... B))
FAILURE
location:   <pkgs>/typed-racket-test/unit-tests/infer-tests.rkt:274:4
Could not infer a substitution
--------------------
````

It seems like this case should succeed though, so I'm not sure what's going on. I'll look into it, but if anyone else has any ideas let me know.

samth:
@takikawa what's the status here?

takikawa:
@samth the status is that I'm stuck on the inference problem above. I'll update this PR when I fix it.

shekari:
Note that this is a backwards incompatible change, because instantiating the functions with `A Any ...` no longer extracts an `A` in many cases. If we are going to break backwards compatibility in this tiny case we should at least make the functions a lot simpler by making the types like `cadr : (All (a) (Pair Any (Pair a Any)))` with the appropriate `Null`s.

samth:
Is this worth keeping open at this point?

pnwamk:
This PR and the associated discussion may be useful to reference in the future, but leaving it as "open" seems incorrect at this point.

PR #17, opened 2015-01-26
Move contracted defns to a submodule.

samth:
I'll squash this and write a real commit message once it works.
Which it doesn't yet.

- [x] pass unit tests
- [x] pass integration tests
- [x] build math and plot
- [x] write a comment explaining this whole thing
- [x] check for excess dependencies
- [x] write a decent commit message

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -16,6 +16,7 @@
  racket/format
  racket/dict
  unstable/list
+ syntax/flatten-begin
  (only-in (types abbrev) -Bottom)
  (static-contracts instantiate optimize structures combinators)
  ;; TODO make this from contract-req
@@ -133,10 +134,13 @@
                                         #,(syntax-position orig-id)
                                         #,(syntax-span orig-id))))])]))
 
+;; The below requires are needed since they provide identifiers that
+;; may appear in the residual program.
+
+;; TODO: It would be better to have individual contracts specify which
+;; modules should be required, but for now this is just all of them.
 (define extra-requires
   #'(require
-      ;; the below requires are needed since they provide identifiers
-      ;; that may appear in the residual program
       (submod typed-racket/private/type-contract predicates)
       typed-racket/utils/utils
       (for-syntax typed-racket/utils/utils)
@@ -144,7 +148,10 @@
       typed-racket/utils/evt-contract
       unstable/contract racket/contract/parametric))
 
-;; should the above requires be included in the output?
+;; Should the above requires be included in the output?
+;;   This box is only used for contracts generated for `require/typed`
+;;   and `cast`, contracts for `provides go into the `#%contract-defs`
+;;   submodule, which always has the above `require`s.
 (define include-extra-requires? (box #f))
 
 (define (change-contract-fixups forms)
@@ -157,15 +164,26 @@
          (begin (set-box! include-extra-requires? #t)
                 (generate-contract-def e ctc-cache sc-cache))))))
 
-(define (change-provide-fixups forms)
-  (define ctc-cache (make-hash))
-  (define sc-cache (make-hash))
+;; TODO: These are probably all in a specific place, which could avoid
+;;       the big traversal
+(define (change-provide-fixups forms  [ctc-cache (make-hash)] [sc-cache (make-hash)])
   (with-new-name-tables
    (for/list ([form (in-list forms)])
-     (cond [(contract-def/provide-property form)
-            (set-box! include-extra-requires? #t)
-            (generate-contract-def/provide form ctc-cache sc-cache)]
-           [else form]))))
+     (syntax-parse form #:literal-sets (kernel-literals)
+       [_
+        #:when (contract-def/provide-property form)
+        (generate-contract-def/provide form ctc-cache sc-cache)]
+       [(module* name #f forms ...)
+        (quasisyntax/loc form
+          (module* name #f 
+            #,@(change-provide-fixups (syntax->list #'(forms ...))
+                                      ctc-cache sc-cache)))]
+       [((~literal #%plain-module-begin) forms ...)
+        (quasisyntax/loc form
+          (#%plain-module-begin
+           #,@(change-provide-fixups (flatten-all-begins #'(begin forms ...))
+                                     ctc-cache sc-cache)))]
+       [_ form]))))
 
 ;; To avoid misspellings
 (define impersonator-sym 'impersonator)
@@ -233,9 +251,6 @@
      fail
      kind
      #:cache cache)))
-
-
-
 
 (define any-wrap/sc (chaperone/sc #'any-wrap/c))
 
--- a/typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
@@ -26,35 +26,44 @@
 (define (freshen-id id)
   ((make-syntax-introducer) id))
 
-;; generate-prov : dict[id -> def-binding] dict[id -> list[id]] id
+;; generate-prov : dict[id -> def-binding] dict[id -> list[id]] id id
 ;;                 -> (values listof[syntax] listof[listof[list[id id]]])
 ;; defs: defines in this module
 ;; provs: provides in this module
 ;; pos-blame-id: a #%variable-reference for the module
+;; mk-redirect-id: the name of a definition created by `make-make-redirect-to-contract`
 
 ;; The first returned value is a syntax object of definitions that defines the
 ;; contracted versions of the provided identifiers, and the corresponding
 ;; provides.
 ;;
 ;; The second value is a list of two element lists, which are type name aliases.
-(define (generate-prov defs provs pos-blame-id)
+(define (generate-prov defs provs pos-blame-id mk-redirect-id)
   ;; maps ids defined in this module to an identifier which is the possibly-contracted version of the key
   (define mapping (make-free-id-table))
 
-  ;; triple/c in the signatures corresponds to three values:
-  ;; (values syntax? identfier? (listof (list/c identifier? identifier?))
-  ;; First return value is a syntax object of definitions
-  ;; Second is the id to export
-  ;; Third is a list of two element lists representing type aliases
+  ;; quad/c in the signatures corresponds to four values:
+  ;; (values syntax? syntax? identfier? (listof (list/c identifier? identifier?))
+  ;; First return value is a syntax object of definitions, which will go in
+  ;;    the #%contract-defs submodule
+  ;; Second is a syntax object of definitions to go in the main module, including 
+  ;;    the defintion to be exported
+  ;; Third is the id to export
+  ;; Fourth is a list of two element lists representing type aliases
 
-  ;; mk : id -> triple/c
+
+  ;; mk-ignored-quad : identifier -> quad/c
+  (define (mk-ignored-quad i) (values #'(begin) #'(begin) i null))
+
+  ;; mk : id -> quad/c
   ;;
-  ;; internal-id : the id being provided
-  ;; if `internal-id' is defined in this module, we will produce a (begin def ... provide) block
+  ;; internal-id : the id being provided. If `internal-id' is defined
+  ;; in this module, we will produce a (begin def ... provide) block
   ;; and a name to provide instead of internal-id.
   ;;
-  ;; Anything already recorded in the mapping is given an empty (begin) and the already-recorded id
-  ;; otherwise, we will map internal-id to the fresh id in `mapping'
+  ;; Anything already recorded in the mapping is given an empty
+  ;; (begin) and the already-recorded id otherwise, we will map
+  ;; internal-id to the fresh id in `mapping'
   (define (mk internal-id)
     (define new-id (freshen-id internal-id))
     (cond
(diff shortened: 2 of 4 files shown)


samth:
This is now in very close to final form. The remaining problem is that static struct info doesn't provide the correct version of the struct identifiers, but everything else appears to build.

@takikawa If you want to start building on this, it's probably reasonable now.

andmkent:
Looks fine to me. (caveat: I've done no hacking in this part of TR yet)

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -157,15 +160,21 @@
...
-     (cond [(contract-def/provide-property form)
-            (set-box! include-extra-requires? #t)
-            (generate-contract-def/provide form ctc-cache sc-cache)]
-           [else form]))))
+     (syntax-parse form #:literal-sets (kernel-literals)
+       [_
+        #:when (contract-def/provide-property form)
+        (generate-contract-def/provide form ctc-cache sc-cache)]
+       [(module* name #f forms ...)
+        #`(module* name #f
```
takikawa: Would preserving syntax locations via `syntax/loc` matter for either of these cases?

typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
```diff
@@ -26,7 +26,7 @@
 (define (freshen-id id)
   ((make-syntax-introducer) id))
 
-;; generate-prov : dict[id -> def-binding] dict[id -> list[id]] id
+;; generate-prov : dict[id -> def-binding] dict[id -> list[id]] id id
 ;;                 -> (values listof[syntax] listof[listof[list[id id]]])
 ;; defs: defines in this module
 ;; provs: provides in this module
```
takikawa: Can you add a description of the `mk-redirect-id` parameter too?

typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
```diff
@@ -100,11 +101,15 @@
 
     (with-syntax* ([id internal-id]
                    [export-id new-id]
-                   [protected-id (freshen-id #'id)])
+                   [protected-id (freshen-id #'id)]
+                   [local-untyped-id (freshen-id #'id)])
```
takikawa: Is this identifier unused?

typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
```diff
@@ -112,29 +117,31 @@
...
+  (define (mk-syntax-quad internal-id new-id)
     (with-syntax* ([id internal-id]
                    [export-id new-id]
-                   [untyped-id (freshen-id #'id)])
-      (define/with-syntax def
-        #`(define-syntax untyped-id
-            (lambda (stx)
-              (tc-error/stx stx "Macro ~a from typed module used in untyped code" 'untyped-id))))
+                   [untyped-id (freshen-id #'id)]
+                   [local-untyped-id (freshen-id #'id)])
```
takikawa: Ditto, unused?

typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
```diff
@@ -37,17 +37,18 @@
...
+(define (generate-prov defs provs pos-blame-id mk-redirect-id)
   ;; maps ids defined in this module to an identifier which is the possibly-contracted version of the key
   (define mapping (make-free-id-table))
 
-  ;; triple/c in the signatures corresponds to three values:
-  ;; (values syntax? identfier? (listof (list/c identifier? identifier?))
-  ;; First return value is a syntax object of definitions
-  ;; Second is the id to export
-  ;; Third is a list of two element lists representing type aliases
+  ;; quad/c in the signatures corresponds to four values:
```
takikawa: Just want to make sure I understand this change. So we need to split up the definition vs. just the id to export so that the definitions can be relocated into a submodule right?

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -6,15 +6,15 @@
...
-         (private parse-type type-annotation syntax-properties)
+         (private parse-type type-annotation syntax-properties type-contract)
          (env global-env init-envs type-name-env type-alias-env
               lexical-env env-req mvar-env scoped-tvar-env
               type-alias-helper)
          (utils tc-utils)
-         (typecheck provide-handling def-binding tc-structs
-                    typechecker internal-forms)
+         "provide-handling.rkt" "def-binding.rkt" "tc-structs.rkt"
+         "typechecker.rkt" "internal-forms.rkt"
```
takikawa: Is there any reason to make these relative paths here? (and if so, maybe a separate commit?)

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -6,15 +6,15 @@
...
          syntax/location
-         racket/format
+         racket/format typed-racket/utils/redirect-contract
```
takikawa: Can this go in the `(utils ...)` require above?

typed-racket-lib/typed-racket/utils/redirect-contract.rkt
```diff
@@ -0,0 +1,25 @@
+#lang racket/base
+
+(require syntax/modcollapse (for-template racket/base))
+(provide make-make-redirect-to-contract)
+
+(define (((make-make-redirect-to-contract varref) id) stx)
```
takikawa: I think this module/function could use purpose statements. Especially since `make-make-redirect-to-contract` is a scary function name.

typed-racket-lib/typed-racket/private/type-contract.rkt
samth: I don't think it can matter, but I'll do it.

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
samth: Relative paths work better with check syntax in drracket.

typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
samth: Yes, that's right. I've improved the comment, hopefully.

typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
samth: Done.

samth:
I've addressed most of the comments, but I still need to write a big doc comment for @stamourv about what the submodule is doing. More feedback before I get that done is also welcome.

samth:
@stamourv let me know what you think about that explanation

samth:
Turns out that this won't entirely remove the dependencies it should, because `prims.rkt` depends on too much stuff. For example, the implementation of `require/typed` has a direct dependency on `racket/contract/base`, which we should remove. I'll open a separate issue for reducing prims dependencies generally.

samth:
This is now done and ready to merge from my perspective.

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -157,15 +164,25 @@
          (begin (set-box! include-extra-requires? #t)
                 (generate-contract-def e ctc-cache sc-cache))))))
 
-(define (change-provide-fixups forms)
-  (define ctc-cache (make-hash))
-  (define sc-cache (make-hash))
+;; These are probably all in a specific place, which could avoid the big traversal
```
shekari: This should be marked as a TODO.

typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
```diff
@@ -62,35 +66,38 @@
       [(dict-ref mapping internal-id
                  ;; if it wasn't there, put it in, and skip this case
                  (λ () (dict-set! mapping internal-id new-id) #f))
-       => (λ (mapped-id) (values #'(begin) mapped-id null))]
+       => (λ (mapped-id) (values #'(begin) #'(begin) mapped-id null))]
```
shekari: This seem like it is the same as line 80. Maybe add a function: `(define (mk-ignored-quad id) (values #'(begin) #'begin) id null))`. Not sure if that would be cleaner though.

typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
```diff
@@ -62,35 +66,38 @@
...
     (define type-is-constructor? #t) ;Conservative estimate (provide/contract does the same)
     (match-define (list type-desc constr pred (list accs ...) muts super) (extract-struct-info si))
-    (define-values (defns new-ids aliases)
-      (map/values 3
+    (define-values (defns export-defns new-ids aliases)
+      (map/values 4
                   (lambda (e) (if (identifier? e)
                                   (mk e)
-                                  (values #'(begin) e null)))
+                                  (values #'(begin) #'(begin) e null)))
```
shekari: Could reuse `mk-ignored-quad` here as well if it supported `#f`.

typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
```diff
@@ -112,47 +132,58 @@
...
-  (for/lists (l l*) ([(internal-id external-ids) (in-dict provs)])
-    (define-values (defs id alias) (mk internal-id))
+  (for/lists (defs export-defs provides aliases) ([(internal-id external-ids) (in-dict provs)])
+    (define-values (defs export-def id alias) (mk internal-id))
     (define provide-forms
       (for/list ([external-id (in-list external-ids)])
-        #`(rename-out [#,id #,external-id])))
-    (when (pair? external-ids) (set-box! include-extra-requires? #t))
-    (values #`(begin #,defs (provide #,@provide-forms))
+        (if (free-identifier=? id external-id)
```
shekari: Why do we need to check here? Is this just an optimization?

typed-racket-lib/typed-racket/utils/redirect-contract.rkt
```diff
@@ -0,0 +1,45 @@
+#lang racket/base
+
+(require syntax/modcollapse (for-template racket/base))
+(provide make-make-redirect-to-contract)
+
+;; This produces an identifier that expands to a local-require.  It's
```
shekari: This sentence confused me, since nothing here returns an identifier. It seems that make-make-redirect-to-contract is a curried function that takes `varref` and `id` and returns a syntax transformer that expands to a local require.

typed-racket-lib/typed-racket/utils/redirect-contract.rkt
```diff
@@ -0,0 +1,45 @@
...
+
+;; varref: a variable reference to the typed module that has the
+;;         appropriate submodule in it
+
+;; id: the name of the export from the submodule which will be
+;;     redirected-to by the local-require
+
+;; stx: the syntax object that's the argument to the macro (that is,
+;;      the stx object that's the reference to the typed identifier in
+;;      a untyped module). The funny eta-expansion with `redirect` is
```
shekari: I don't see why you need the eta expansion, just remove the `stx` argument and return `redirect`. This also makes it clearer that this is returning a syntax transformer.

typed-racket-lib/typed-racket/utils/redirect-contract.rkt
```diff
@@ -0,0 +1,45 @@
...
+    (cond
+     [(identifier? stx)
+      (with-syntax ([mp (collapse-module-path-index
+                           (module-path-index-join
+                            '(submod "." #%contract-defs)
+                            (variable-reference->module-path-index
+                             varref)))]
+                    [i (datum->syntax id (syntax-e id) stx stx)])
+        #`(let ()
+            (local-require (only-in mp [#,(datum->syntax #'mp (syntax-e #'i)) i]))
```
shekari: I'm confused by the context argument to datum->syntax here. Can you explain whats going on? I thought module exports were symbols not identifiers, and so this should be unnecessary.

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -350,63 +362,130 @@
...
-                (dict-update h #'in (lambda (tail) (cons #'out tail)) '())]
-               [(name:unknown-provide-form . _)
-                (tc-error "provide: ~a not supported by Typed Racket" (syntax-e #'name.name))]
-               [_ (int-err "unknown provide form")])))]
+           (syntax-parse f
+             [i:id
+              (dict-update h #'i (lambda (tail) (cons #'i tail)) '())]
+             [((~datum rename) in out)
+              (dict-update h #'in (lambda (tail) (cons #'out tail)) '())]
+             [(name:unknown-provide-form . _)
```
shekari: Should these be unsupported provide form instead of unknown?

shekari:
LGTM.

typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
samth: It just simplifies the expansion.

typed-racket-lib/typed-racket/utils/redirect-contract.rkt
samth: Module exports are symbols, but `require` and `local-require` bind based on identifiers. This makes sure that the name binds the reference on the next line.

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
samth: You mean rename the syntax class? There isn't anything currently called `unsupported-provide-form`.

stamourv:
Sorry for the late reply, was traveling.
Comments look good. We need more like that.

PR #18, opened 2015-01-27
Improve the type for directory-list

lexi-lambda:
Really simple change. Just adds support for the `#:build?` keyword argument for `directory-list`.


stamourv:
Looks good to me. Could you add a test?

lexi-lambda:
@stamourv Sure. Where should the test for this go?

stamourv:
I think a unit test would make most sense, in which case it would go in `typed-racket-test/unit-tests/typecheck-tests.rkt`.

lexi-lambda:
@stamourv Thanks, I added some tests.

PR #19, opened 2015-01-28
Remove uses of srfi/1 in Typed Racket

takikawa:


samth:
should `unzipN` be added to `racket/list`?

Otherwise, r+

PR #22, opened 2015-02-01
Add typed/srfi/19 to typed-racket-more

lexi-lambda:
This should include accurate types for all the bindings exported from `srfi/19`. (I'd add tests, but I don't think there are any tests for typed-racket-more at the moment. Let me know if there are, and I'll add them.)

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/libraries.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/libraries.scrbl
@@ -77,6 +77,7 @@ The following libraries are included with Typed Racket in the
 @defmodule/incl[typed/racket/system]
 @defmodule/incl[typed/rackunit]
 @defmodule/incl[typed/srfi/14]
+@defmodule/incl[typed/srfi/19]
 @defmodule/incl[typed/syntax/stx]
 @defmodule/incl[typed/json]
 
--- a/typed-racket-more/typed/srfi/19.rkt
+++ b/typed-racket-more/typed/srfi/19.rkt
@@ -1,0 +1,104 @@
+#lang typed/racket/base
+
+(require/typed/provide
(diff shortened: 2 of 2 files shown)


lexi-lambda:
@samth Can I get an update on this?

samth:
@lexi-lambda Sorry about the delay with the review here -- this all looks quite good. Do the error messages show the fake struct you create ever?

lexi-lambda:
@samth I've tested it in various situations, and I haven't managed to make TR print the fake struct in any error messages. The only possible problem is that there isn't actually way, as far as I can tell, to introspect what the types of the aliases actually are.

For example, evaluating `(:print-type make-time)` simply prints this:

    (-> Time-Type Nanosecond Second Time)

It doesn't actually print what, say, `Nanosecond` actually is, and there doesn't appear to be any way to find out. It's actually an `Index`, but short of looking at the `srfi/19` documentation, I don't think there would be any way to be aware of that.

samth:
Ok, seems reasonable. @takikawa @stamourv @shekari Any thoughts on the alias printing issue?

lexi-lambda:
Actually, I've thought of another problem which makes me reconsider this hack. As much as I like its elegance, I don't think it's a good idea to use in actual code. This is why:

    (define x : Index (time-nanosecond (current-time)))

The above code fails to typecheck.

samth:
Yeah, that's not ok, especially since just changing `Index` to `Number` will also not type check.

I think the right solution is to just not use the type aliases, given the error message issues.

/me should do something more sensible with error messages.

lexi-lambda:
@samth Fixed. It would be nice to have better error messages, though, so that something like that is at least possible.

samth:
Ok, great, I'll merge soon.

PR #24, opened 2015-02-10
Handle expansion of internal forms inside splicing-syntax-parameterize

lexi-lambda:
See PR 14973.


lexi-lambda:
@samth I updated this with your suggestion.

PR #25, opened 2015-02-11
Add initial support for prefab struct types

takikawa:
The main thing that is missing from this PR is contract generation support. It turns out that the contract system doesn't provide a suitable contract combinator for prefab structs, so we'll have to make one. I'm planning to follow up with that in a separate PR.

The first few commits and 04ac72e are misc commits that are related (came up while developing) but aren't directly about prefabs.

Other missing features: individually mutable fields, support for auto-values, and typechecking for functions like `make-prefab-struct`.

For the first two, I'm planning to generally improve struct support (not just prefabs) in a separate PR.

### DIFF
--- a/typed-racket-lib/typed-racket/core.rkt
+++ b/typed-racket-lib/typed-racket/core.rkt
@@ -113,7 +113,8 @@
                        (cond [(andmap equal? tgs tcs) ""]
                              [indented?
                               (format "\n[more precisely: ~a]"
-                                      (pretty-format-type (make-Values tcs) #:indent 17))]
+                                      (pretty-format-type (make-Values (map -result tcs))
+                                                          #:indent 17))]
                              [else (format " [more precisely: ~a]" (cons 'Values tcs))])
                        ;; did any get pruned?
                        (cond [(andmap equal? t tcs) ""]
--- a/typed-racket-lib/typed-racket/infer/infer-unit.rkt
+++ b/typed-racket-lib/typed-racket/infer/infer-unit.rkt
@@ -547,6 +547,12 @@
                         [else empty])])
              (% cset-meet proc-c (cgen/flds context flds flds*)))]
 
+          ;; two prefab structs with the same key
+          [((Prefab: k flds) (Prefab: k* flds*))
+           #:when (equal? k k*)
+           ;; FIXME: should account for mutable fields
+           (cgen/list context flds flds*)]
+
           ;; two struct names, need to resolve b/c one could be a parent
           [((Name: n _ #t) (Name: n* _ #t))
            (if (free-identifier=? n n*)
--- a/typed-racket-lib/typed-racket/infer/infer-unit.rkt
+++ b/typed-racket-lib/typed-racket/infer/infer-unit.rkt
@@ -13,7 +13,7 @@
            (utils tc-utils)
            (rep free-variance type-rep filter-rep object-rep rep-utils)
            (types utils abbrev numeric-tower union subtype resolve
-                  substitute generalize)
+                  substitute generalize prefab)
            (env index-env tvar-env))
           make-env -> ->* one-of/c)
          "constraint-structs.rkt"
@@ -548,10 +548,16 @@
              (% cset-meet proc-c (cgen/flds context flds flds*)))]
 
           ;; two prefab structs with the same key
-          [((Prefab: k flds) (Prefab: k* flds*))
-           #:when (equal? k k*)
-           ;; FIXME: should account for mutable fields
-           (cgen/list context flds flds*)]
+          [((Prefab: k ss) (Prefab: k* ts))
+           #:when (and (prefab-key-subtype? k k*)
+                       (>= (length ss) (length ts)))
+           (% cset-meet*
+              (for/list/fail ([s (in-list (take ss (length ts)))]
+                              [t (in-list ts)]
+                              [mut? (in-list (prefab-key->field-mutability k*))])
+                (if mut?
+                    (cgen/inv context s t)
+                    (cgen context s t))))]
 
           ;; two struct names, need to resolve b/c one could be a parent
           [((Name: n _ #t) (Name: n* _ #t))
--- a/typed-racket-lib/typed-racket/infer/infer-unit.rkt
+++ b/typed-racket-lib/typed-racket/infer/infer-unit.rkt
@@ -552,7 +552,7 @@
            #:when (and (prefab-key-subtype? k k*)
                        (>= (length ss) (length ts)))
            (% cset-meet*
-              (for/list/fail ([s (in-list (take ss (length ts)))]
+              (for/list/fail ([s (in-list ss)]
                               [t (in-list ts)]
                               [mut? (in-list (prefab-key->field-mutability k*))])
                 (if mut?
--- a/typed-racket-lib/typed-racket/rep/type-rep.rkt
+++ b/typed-racket-lib/typed-racket/rep/type-rep.rkt
@@ -402,9 +402,18 @@
   ;; This should eventually be based on understanding of struct properties.
   [#:key '(struct procedure)])
 
+;; Represents prefab structs
+;; key  : prefab key encoding mutability, auto-fields, etc.
+;; flds : the types of all of the prefab fields
+(def-type Prefab ([key prefab-key?]
+                  [flds (listof Type/c)])
+  [#:frees (λ (f) (combine-frees (map f flds)))]
+  [#:fold-rhs (*Prefab key (map type-rec-id flds))]
+  [#:key 'prefab])
+
 ;; A structure type descriptor
 (def-type StructTypeTop () [#:fold-rhs #:base] [#:key 'struct-type])
-(def-type StructType ([s (or/c F? B? Struct?)]) [#:key 'struct-type])
+(def-type StructType ([s (or/c F? B? Struct? Prefab?)]) [#:key 'struct-type])
 
 ;; the supertype of all of these values
 (def-type BoxTop () [#:fold-rhs #:base] [#:key 'box])
--- a/typed-racket-lib/typed-racket/typecheck/tc-literal.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-literal.rkt
@@ -133,7 +133,7 @@
 (define (tc-prefab struct-inst expected)
   (define expected-ts
     (match (and expected (resolve expected))
-      [(Prefab: _ ts) (in-list ts)]
+      [(Prefab: _ ts) (in-sequence-forever (in-list ts) #f)]
       [_ (in-cycle (in-value #f))]))
   (define key (prefab-struct-key struct-inst))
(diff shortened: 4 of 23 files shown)


typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -225,9 +228,17 @@
 (define (refine-struct-variance! parsed-structs)
   (define stys (map parsed-struct-sty parsed-structs))
   (define tvarss (map (compose struct-desc-tvars parsed-struct-desc) parsed-structs))
-  (define names (map Struct-name stys))
+  (define names
+    (for/list ([parsed-struct (in-list parsed-structs)])
+      (struct-names-type-name (parsed-struct-names parsed-struct))))
   (refine-variance! names stys tvarss))
 
+;; get-struct-name : (U Struct Prefab) -> Symbol
```
takikawa: Whoops, I forgot that I added this function. I'm planning to delete it since it's unused now.

typed-racket-lib/typed-racket/infer/infer-unit.rkt
```diff
@@ -547,6 +547,12 @@
                         [else empty])])
              (% cset-meet proc-c (cgen/flds context flds flds*)))]
 
+          ;; two prefab structs with the same key
+          [((Prefab: k flds) (Prefab: k* flds*))
+           #:when (equal? k k*)
+           ;; FIXME: should account for mutable fields
```
samth: We can't merge this without either a more-conservative approach or a correct approach here.

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -402,9 +402,18 @@
...
   [#:key '(struct procedure)])
 
+;; Represents prefab structs
+;; key  : prefab key encoding mutability, auto-fields, etc.
+;; flds : the types of all of the prefab fields
+(def-type Prefab ([key prefab-key?]
+                  [flds (listof Type/c)])
+  [#:frees (λ (f) (combine-frees (map f flds)))]
+  [#:fold-rhs (*Prefab key (map type-rec-id flds))]
+  [#:key 'prefab])
```
samth: Can prefabs be `procedure?`?

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -272,6 +283,49 @@
 
   (parsed-struct sty names desc (struct-info-property nm/par) type-only))
 
+;; check and register types for a prefab struct
+(define (tc/prefab vars nm/par fld-names tys
```
samth: This needs a signature. Also, do we want to support `#:type-only?` here?

typed-racket-lib/typed-racket/types/prefab.rkt
```diff
@@ -0,0 +1,112 @@
...
+                        (-> prefab-key? integer? prefab-key?)]
+                       [prefab-key->field-count
+                        (-> prefab-key? integer?)]
+                       [abbreviate-prefab-key
+                        (-> prefab-key? prefab-key?)]
+                       [prefab-key-subtype?
+                        (-> prefab-key? prefab-key? any)])
+
+;; Convert a prefab key to its expanded version
+(define (normalize-prefab-key key field-length)
```
samth: I feel like this function must exist somewhere already.

typed-racket-lib/typed-racket/types/prefab.rkt
```diff
@@ -0,0 +1,112 @@
...
+  (let loop ([key key] [count 0])
+    (if (null? key)
+        count
+        (loop (cddddr key)
+              (+ (cadr key)
+                 (car (caddr key))
+                 count)))))
+
+;; Convert a prefab key to a shortened version
+(define (abbreviate-prefab-key key)
```
samth: Similarly, this function feels like it belongs somewhere else.

typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -93,25 +93,31 @@
                #:expected golden
                (string-append base-message extra-message1 extra-message2)))))
 
-  ;; test: syntax? tc-results? [(option/c tc-results?)] -> void?
+  ;; test: syntax? tc-results? [(option/c tc-results?)]
+  ;;       [(listof (list id type))] -> void?
```
samth: Need more explanation of the changes to this file.

typed-racket-lib/typed-racket/rep/type-rep.rkt
takikawa: No, because prefab struct type declarations cannot use the `#:property` option.

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
takikawa: I don't think we need `#:type-only?` here because it's only used for `require/typed`. Since prefab structs can be declared anywhere, there's no need to do a special `require/typed` for them.

typed-racket-test/unit-tests/typecheck-tests.rkt
takikawa: Do you mean in the commit message, in comments, or elsewhere?

samth: Mostly in the file, but the commit message would be good too.

typed-racket-lib/typed-racket/types/prefab.rkt
maxsnew: `cddddr` but not `caaddr`?

takikawa:
I've now fixed the inference/subtyping for mutable prefab fields. I think I've also addressed the other remaining comments. My plan is to add contract generation separately and push that simultaneously with the `prefab/c` pull request for the core.

typed-racket-lib/typed-racket/core.rkt
```diff
@@ -113,7 +113,8 @@
                        (cond [(andmap equal? tgs tcs) ""]
                              [indented?
                               (format "\n[more precisely: ~a]"
-                                      (pretty-format-type (make-Values tcs) #:indent 17))]
+                                      (pretty-format-type (make-Values (map -result tcs))
+                                                          #:indent 17))]
```
samth: Is 17 computed from something?

takikawa: It's computed from the length of the "more precisely" string above in order to get the type to indent to the right position.

typed-racket-lib/typed-racket/infer/infer-unit.rkt
```diff
@@ -547,6 +547,18 @@
                         [else empty])])
              (% cset-meet proc-c (cgen/flds context flds flds*)))]
 
+          ;; two prefab structs with the same key
+          [((Prefab: k ss) (Prefab: k* ts))
+           #:when (and (prefab-key-subtype? k k*)
+                       (>= (length ss) (length ts)))
+           (% cset-meet*
+              (for/list/fail ([s (in-list (take ss (length ts)))]
```
samth: Why is this `take` call needed? The `for` loop will already stop at the length of `ts`.

typed-racket-lib/typed-racket/typecheck/tc-literal.rkt
```diff
@@ -124,8 +125,21 @@
...
 
-
-
-
+;; Typecheck a prefab struct literal
+(define (tc-prefab struct-inst expected)
+  (define expected-ts
+    (match (and expected (resolve expected))
+      [(Prefab: _ ts) (in-sequence-forever (in-list ts) #f)]
+      [_ (in-cycle (in-value #f))]))
```
samth: This just throws away the expected type when it's wrong. Can't that miss errors?

takikawa: The result of `tc-literal` is checked against the expected type with `check-below` in `tc-expr/check`, so it should not be possible.

typed-racket-lib/typed-racket/infer/infer-unit.rkt
takikawa: Good catch, it probably isn't needed.

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -33,10 +33,17 @@
...
+           (tc/prefab (attribute t.tvars)
+                      #'t.nm
+                      (syntax->list #'(t.fields ...))
+                      (syntax->list #'(t.types ...))
+                      #:maker (attribute t.maker)
+                      #:mutable (attribute t.mutable))
+           (tc/struct (attribute t.tvars) #'t.nm (syntax->list #'(t.fields ...)) (syntax->list #'(t.types ...))
+                      #:mutable (attribute t.mutable)
+                      #:maker (attribute t.maker)
+                      #:type-only (attribute t.type-only)))]
```
samth: I still feel like these two calls could be combined into one.

typed-racket-lib/typed-racket/types/prefab.rkt
```diff
@@ -0,0 +1,127 @@
+#lang racket/base
+
+;; Utilities for dealing with prefab struct types
+
+(require "../utils/utils.rkt"
+         (contract-req)
+         racket/list
+         racket/match)
+
+(provide/cond-contract [normalize-prefab-key
```
samth: This code should all go somewhere in core Racket -- see discussion w/ @mflatt elsewhere. But that's fine to do later, once we figure out what we want.

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -554,6 +554,20 @@
          ;; subtyping on structs follows the declared hierarchy
          [((Struct: nm (? Type/c? parent) _ _ _ _) other)
           (subtype* A0 parent other)]
+         [((Prefab: k1 ss) (Prefab: k2 ts))
+          (and (prefab-key-subtype? k1 k2)
+               (and (>= (length ss) (length ts))
+                    (for/fold ([A A0])
+                              ([s (in-list (take ss (length ts)))]
```
samth: Again w/ the extra `take`.

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
takikawa: I'll can try that refactoring and see if it improves the code.

takikawa:
I removed the redundant `take`s and merged the `tc/struct` and `tc/prefab` functions now.

typed-racket-lib/typed-racket/core.rkt
samth: Should we have a (computed) constant somewhere for this?

samth:
Looks good.

takikawa:
Merged.

Issue #26, opened 2015-02-12
Typechecking on mutually recursive types is buggy when used in a union with Symbol


lexi-lambda:
> Note: This bug is so strange that it's rather difficult to even reproduce. I'll try and describe exactly what circumstances allow me to reproduce it.

Take the following trivial program. Call it `bug.rkt`.

```racket
#lang typed/racket/base

(define-type Bson-Document (HashTable Symbol Bson-Value))
(define-type Bson-Value (U Bson-Document Symbol))

(define (make-document) : Bson-Value
  (define log-data : Bson-Document (make-immutable-hasheq `()))
  log-data)
```

For me, if I run `raco setup`, then running `racket bug.rkt` typechecks just fine, as it should! However, now I run the following command:

    raco setup -c --pkgs typed-racket-more

This shouldn't have any affect on any program, much less this one, which uses absolutely *zero* functionality from `typed-racket-more`. Nevertheless, now if I run `racket bug.rkt`, I get the following type error:

```
/tmp/bug.rkt:8:2: Type Checker: type mismatch
  expected: Bson-Value
  given: Bson-Document
  in: log-data
  context...:
   /Applications/Racket v6.1.1.8/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:284:0: type-check
   /Applications/Racket v6.1.1.8/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:494:0: tc-module
   /Applications/Racket v6.1.1.8/share/pkgs/typed-racket-lib/typed-racket/tc-setup.rkt:39:0: tc-setup
   /Applications/Racket v6.1.1.8/share/pkgs/typed-racket-lib/typed-racket/typed-racket.rkt:18:4
   standard-module-name-resolver
```

Don't worry, it gets weirder. I now tried modifying the definition for `Bson-Value`. I changed it to `(U Bson-Document String)`—anything that didn't include `Symbol` worked. Then I ran `racket bug.rkt` again. No error!

Okay... now I changed `Bson-Value` back to `(U Bson-Document Symbol)`. This time, though, I reversed the order of the type type declarations, so lines 3 and 4 looked like this:

```racket
(define-type Bson-Value (U Bson-Document Symbol))
(define-type Bson-Document (HashTable Symbol Bson-Value))
```

[...]

PR #28, opened 2015-02-15
Prevent type alias placeholders from collapsing in unions with symbols

lexi-lambda:
This implements the solution proposed by @takikawa for issue #26. It also includes his example as an integration test. All existing tests pass with this change.


takikawa:
Thanks! Merged (commit 619d6945c3263d21e6439a40286cde90e6afd585).

PR #30, opened 2015-02-17
Adjust the fix in 619d6945c3263d21e

takikawa:
For unions of multiple type aliases the new placeholder would cause union collapsing incorrectly. Put an uninterned symbol in the placeholder types to avoid this.

@samth @lexi-lambda any feedback?


samth:
This looks good to me.

takikawa:
BTW, I think in the long-term the `remove-type-name` function is not necessary since I can probably avoid registering it to begin with for non-recursive aliases. But I don't want this short-term fix to depend on that re-write.

PR #34, opened 2015-02-26
adjustments to accommodate a different macro system

mflatt:
These changes make TR work with both the current macro expander and one that has a different underlying model of binding (but is only slightly different on the outside).


samth:
This looks good to me -- feel free to merge it.

PR #35, opened 2015-02-27
`define-runtime-module-path-index` for reference to a contract submodule

mflatt:
This change is meant to fix problems with `raco exe` --- along with a repair in `raco exe`.


samth:
This looks good to me.

samth:
This was merged.

PR #39, opened 2015-03-03
move contract submodules to start of enclosing module

mflatt:
Putting the contract submodules early enough allows them to be used
for `module*` submodules.

This change is intended to repair the problem that Alexander reported,


samth:
Can you add a link to Alexander's report in the commit message? Also, can you add a test (just a file in the succeed directory in the tests should be enough)?

mflatt:
I've added the test and the pointer, but this change causes the "succeed/even-odd-recursive-type.rkt" test to fail, so the change is not right. In any case, the ordering of submodules is the issue for Alexander's example.

samth:
Is the right solution some kind of sorting?

mflatt:
@samth I don't know; I'm not clear on what goes wrong with the even-odd test

samth:
Merged with fix to some other code.

PR #40, opened 2015-03-03
Add syntax location information to cast party

takikawa:
This adds more syntax information to cast blame parties. Example:

````
3: broke its contract
  promised: String
  produced: 3
  in: String
  contract from: 
      (cast
       /home/asumu/plt/racket-git-2/extra-pkgs/typed-racket/typed-racket-test/fail/cast-mod1.rkt
       6:0)
  blaming: (cast /home/asumu/plt/racket-git-2/extra-pkgs/typed-racket/typed-racket-test/fail/cast-mod1.rkt 6:0)
````

(the "contract from:" bit changed)

Any opinions on the format of the blame party?

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims.rkt
@@ -316,7 +316,11 @@ This file defines two sorts of primitives. All of them are provided into any mod
                        'cast
                        'typed-world
                        val
-                       (quote-syntax #,stx)))
+                       (srcloc #,(syntax-source stx)
+                               #,(syntax-line stx)
+                               #,(syntax-column stx)
+                               #,(syntax-position stx)
+                               #,(syntax-span stx))))
                     'feature-profile:TR-dynamic-check #t))
              #'ty)))
 
--- a/typed-racket-test/fail/cast-mod1.rkt
+++ b/typed-racket-test/fail/cast-mod1.rkt
@@ -1,5 +1,5 @@
 #;
-(exn-pred exn:fail:contract? #rx".*produced: 3" #rx".*promised: String.*" )
+(exn-pred exn:fail:contract? #rx".*produced: 3" #rx".*promised: String.*" #rx"6\\.0")
 
 #lang typed/racket/base
 

### CODE REVIEW

**Comment by rfindler** [comment]
Seems like the "blaming" line and the "contract from" line have the same
information. Can maybe the blame line say "the cast" or something like that?

Robby

**Comment by takikawa** [comment]
@rfindler is there a simple way to adjust the "contract from" line separately from the "blaming" line using the `contract` form? I just supplied the extra info for the positive blame expression, but is there a better way?

**Comment by rfindler** [comment]
Are you using the 6 argument version of `contract`?

Robby

**Comment by takikawa** [comment]
Yes. With my commit it looks like this:

````
(contract #,ctc-expr
          val
          '(cast #,(syntax-source stx)
                 #,(format "~a:~a" (syntax-line stx) (syntax-column stx)))
          'typed-world
          val
          (quote-syntax #,stx)))
````

**Comment by rfindler** [comment]
If you change the third argument to just 'the-cast, do the error message
look better?

Robby

**Comment by takikawa** [comment]
So that's what it used to be before my commit. I realized that the actual problem wasn't what the blame party was set to, it's that the last expression `(quote-syntax #,stx)` seems to be failing to provide line/column information in certain cases for TR casts. I'll try to fix that instead.

**Comment by rfindler** [comment]
Ah, good!

Robby

**Comment by rfindler** [comment]
I think that last argument can be a srcloc. Probably that's a better thing
to do than sticking a quoted syntax object in there anyway.

**Comment by takikawa** [comment]
Ok, this should provide better syntax locations now without any change to the blame label.

File: typed-racket-test/fail/cast-mod1.rkt [hunk 1]
```diff
@@ -1,5 +1,5 @@
 #;
-(exn-pred exn:fail:contract? #rx".*produced: 3" #rx".*promised: String.*" )
+(exn-pred exn:fail:contract? #rx".*produced: 3" #rx".*promised: String.*" #rx"6\\.0")
```
samth: What is the number 6 doing here?

typed-racket-lib/typed-racket/base-env/prims.rkt
```diff
@@ -316,7 +316,11 @@ This file defines two sorts of primitives. All of them are provided into any mod
                        'cast
                        'typed-world
                        val
-                       (quote-syntax #,stx)))
+                       (srcloc #,(syntax-source stx)
```
samth: I think there's a form for doing this -- looks like it's `(quote-srcloc stx)`

typed-racket-test/fail/cast-mod1.rkt
takikawa: That's the line/column number. It's testing that the source location is intact. (it was intact for this case before too, but there are some situations in which it's missing)

typed-racket-lib/typed-racket/base-env/prims.rkt
takikawa: That adds another library dependency though. Is that fine?

samth: I'm pretty sure we depend on that elsewhere.

typed-racket-test/fail/cast-mod1.rkt
samth: Ok, makes sense.

takikawa:
Merged.

PR #44, opened 2015-03-04
Adjust named-let to use the return type annotation whenever possible

### LINKED ISSUE #43: Return type annotations on named-let forms aren't considered within the body unless all bindings are explicitly annotated
This fails to typecheck if it is run at the top-level (within a module):

```racket
(let loop : Integer ([n 10])
  (cond
    [(= n 5) n]
    [else (loop (sub1 n))]))
```

The return type of `loop` is considered to be `AnyValues` within the loop body (inferred from the surrounding context), ignoring the annotation. However, if `n` is explicitly annotated, for some reason it typechecks:

```racket
(let loop : Integer ([n : Integer 10])
  (cond
    [(= n 5) n]
    [else (loop (sub1 n))]))
```

lexi-lambda:
Fixes #43. This just adds an explicit annotation around the entire `let` expression to enforce the return type while still maintaining type inference on the bindings.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims.rkt
@@ -439,26 +439,29 @@ This file defines two sorts of primitives. All of them are provided into any mod
            (~and (~seq (~optional (~seq : ret-ty))
                        (bs:optionally-annotated-binding ...) body ...)
                  (~seq rest ...)))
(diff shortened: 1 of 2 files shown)


samth:
looks good to me

lexi-lambda:
Merged in c79b61a7515f0a24e1d3d1dd37f6b5398d28d8ac.

PR #45, opened 2015-03-06
Vanilla Dependent Types Preview

andmkent:
### Summary
I've added refinement types of the form: {x : τ | ψ} where
* x is an identifier (actually these get turned into DeBruijn indices)
   and is in scope for both τ and ψ
* τ is a type
* ψ is an arbitrary proposition

In #lang typed/racket these types are written like this:
```racket
(Refine [x : τ] ψ)
```

This means that the following program will typecheck:
```racket
#lang typed/racket
(define-type-alias Num Number)
(define-type-alias Bool Boolean)

(: negate/dep (->i [x : (U Num Bool)] 
                   (Refine [z : (U Num Bool)]
                           (or (and (Num @ z) (Num @ x))
                               (and (Bool @ z) (Bool @ x))))))
(define (negate a)
  (cond
    [(number? a) (* -1 a)]
    [else (not a)]))

(add1 (negate -41))

(: foo (-> (U Num Bool) Num))
(define (foo x)
  (let ([y (negate x)])
    (if (number? y)
        x
        42)))

(foo -3)
```
### To Do's:
- [ ]  Add propositions than can talk about comparisons of linear expressions
   (e.g.  (≤ (+ (car x) y) z))
- [ ]  Add linear expressions as Objects
- [ ]  Add a path element LengthPE that can talk about the length of vectors
   (maybe later make it able to talk about list length as well since that's
    static?)
- [x]  fix 1 last bug (an identifier is not getting it's type put in the type table,
   which is causing an error while building the math lib)
- [ ]  generalize dependent function constructor syntax (right now it's hard coded
   to support 1, 2, or 3 arguments, doh! lol)
- [ ]  a little tidying (as I'm writing this, I'm seeing a change or two that way back
   when I thought was needed and not it's just useless extra arguments w/ default
   values (e.g. env in (define (check-below tr1 expected #:env [env #f] #:obj [obj #f]) ...))

#### Notes for files with many/complex changes:
* .../env/lexical-env.rkt (CHANGED)
   most of this was moved to .../typecheck/tc-envops.rkt and
   expanded w/ the necc. extra logic

* .../env/lookup.rkt (NEW)
   I added this file so all the logic for looking up types could be
   together and some code that explicitly reasons about aliasing details
   could just call functions in here

* .../env/type-env-structs.rkt (CHANGED)
   moved the more complicated logic in here elsewhere so now its just the
   simple raw operations

* ...logic/prop-ops.rkt (NEW)
   logic for extracting propositions from types/propositions. In other words
   we don't really want to keep nested info like ({y : Num | (Num @ z)} @ x) around,
   we want to break it down into it's atomic facts: (and (Num @ x) (Num @ z))

* .../logic/proves.rkt (NEW)
   Contains the judgement Γ |- ψ and related logic. It's built assuming goals
   will usually be small and environments will more often be large.

* .../logic/type-update.rkt (NEW)
   Logic for updating a type in a given environment. For example, if the return
   type of a function is {x : (U Num Str) | (or (and (Num @ x) (Num @ y))
                                                (and (Str @ x) (Str @ y)))}
   and y (a variable local to the function) is known to be a Num, we can update
   the return type to Num before we have to substitute the empty object in for y.

* .../rep/type-rep.rkt (CHANGED)
   new type & logic for abstracting/instantiating idents (turning
   identifiers into DeBruijns and back again)

* .../typecheck/tc-app/tc-app-main.rkt (CHANGED)
   if a function being applied is dependent, calculate the *actual* inferrable
   types of the arguments instead of just checking if it is a subtype of the range.
   This allows the refinement of dependent return types based on their argument type

* .../typecheck/tc-envops.rkt (CHANGED)
   anytime we add *any* new info to the type environment, we need to flatten/extract
   any nested type information (see ...logic/prop-ops.rkt) and then update
   with all the flattened info. Also, some of the logic that was deep down
   inside of env/type-env-structs.rkt is now here. (NOTE: there was one spot (I think
   in the OO code?) where I wasn't sure exactly what to do about potential nested props,
   so I left one old-school version in there called "with-lexical-env/naive-extend-types"

* .../typecheck/tc-funapp.rkt (CHANGED)
    dependent functions now have their types updated w/ the incoming argument types

* .../typecheck/tc-lambda-unit.rkt (CHANGED)
    instantiate DeBruijns in function types if we're using a dependent type

* .../typecheck/tc-let-unit.rkt (CHANGED)
    really just made a helper function to clean things up a little, it looks like
    a big change but it's really not

* .../typecheck/tc-metafunctions.rkt (CHANGED)
    cleaned up messy/redundant code

* .../typecheck/tc-subst.rkt (CHANGED)
    fixed substitution to correctly handle and recur into dependent types

* .../types/filter-ops.rkt (CHANGED)
    added a "-simple-or" that tries to do a little less of an overhaul of the input
    so that when a user gives refinement propositions the types more closely resemble
    what the user gave... this may or may not be worth it
    Oh, and cleaning.

* .../types/subtype.rkt (CHANGED)
    subtyping must have the current context to reason about dependent types --
    thus the lexical environment and the current object are now explicit arguments
    to all the subtyping functions now.
    Also, function subtyping on dependent functions has to reason about the actual
    domain types and how they affect each other and the range when doing subtyping.


samth:
Can you reformat the PR description so that it uses markdown for code blocks and for checklists?

typed-racket-lib/typed-racket/logic/proves.rkt
```diff
@@ -0,0 +1,140 @@
...
+
+;;returns a list of the remaining goals to be proved
+;; only proves based on type-env lookups
+;; A env obj goal -> filter w/ proven facts removed
+(define/cond-contract (logical-reduce A env obj goal)
+  (c:-> any/c env? Path? Filter/c
+        (listof Filter/c))
+  (match goal
+  
+    [(Bot:) `(,goal)]
```
samth: This is `(list goal)`.

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-main.rkt
```diff
@@ -78,16 +78,22 @@
 
        (define arg-tys
          (match f-ty
+           ;; TODO(AMK) support multi-arr functions w/ dependent types
+           [(Function: (list (arr: doms rng rest drest kws #t)))
+            (define arg-ts (map (curryr tc-expr/check? #f) args*))
```
samth: I think an explicit `for/list` would be cleaner here.

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
```diff
@@ -97,12 +122,104 @@
     (pattern (~seq #:unreachable form:expr))
     (pattern (~seq) #:with form #'(begin)))
   (syntax-parse stx
-    [(_ ps:expr u:unreachable? . b)
-     #'(let-values ([(new-env atoms) (env+ (lexical-env) ps)])
+    [(_ ps:expr u:unreachable? body ...)
```
samth: why the switch to `...`?

andmkent: I thought 'body ...' might be clearer/harder to mess up (I introduced a bug I had to find when I forgot b was a list). Maybe changing 'b' to 'bodies' and leaving the '.' would be the clarification I _really_ wanted.

AlexKnauth:
I tried to build drracket with this, but I got a bunch of errors, starting with this:
in drracket/private/tooltip.rkt lines 95-100:
```racket
    (define/override (show on?)
      (when timer
        (cond
          [on? (send timer start 200 #f)]
          [else (send timer stop)]))
      (super show on?))
```
Gives these errors:
```
/Applications/Racket v6.1.1.8 2/Racket v6.1.1.8/share/pkgs/drracket/drracket/private/tooltip.rkt:98:15: Type Checker: send: type mismatch
  expected: an object
  given: (U False (Instance Timer%))
  in: (send timer start 200 #f)
  context...:
   f44
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:275:0: type-check
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:415:0: tc-module
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/tc-setup.rkt:39:0: tc-setup
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/typed-racket.rkt:18:4
   standard-module-name-resolver
/Applications/Racket v6.1.1.8 2/Racket v6.1.1.8/share/pkgs/drracket/drracket/private/tooltip.rkt:99:16: Type Checker: send: type mismatch
  expected: an object
  given: (U False (Instance Timer%))
  in: (send timer stop)
  context...:
   f44
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:275:0: type-check
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:415:0: tc-module
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/tc-setup.rkt:39:0: tc-setup
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/typed-racket.rkt:18:4
   standard-module-name-resolver
Type Checker: Summary: 2 errors encountered
  context...:
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:275:0: type-check
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:415:0: tc-module
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/tc-setup.rkt:39:0: tc-setup
   /Users/Alex/Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/typed-racket.rkt:18:4
   standard-module-name-resolver
```
When it should work fine because of occurrence typing.

lexi-lambda:
I actually got some of those errors, too, though I wasn’t trying this PR. I think it has to do with some change made to racket (possibly the object? change?) that raco isn’t picking up on.

Try running raco setup -c and rebuilding everything from scratch; that worked for me.

andmkent:
Interesting.

I'm still doing a little debugging, so it wouldn't surprise me at all if
the error was due to my changes. The one error I'm looking into right now
could be related... but with Alexis' comment I'm thinking (hoping? ha ha)
maybe it's not. Either way, when I fix the bug I'm currently looking into
I'll push it to this PR with an update if there's any other bugs I discover.

It should be noted, I'm working on a racket checkout from 2015-01-31 (the
TR checkout is more recent)... so it's also possible I'll discover some
errors whenever I pull (I'm not in any rush though, I'm trying to make
progress on this branch first before I introduce more variables into the
mix).

AlexKnauth:
I tried raco setup -c followed by raco setup, and the error is still there, but I think that's because this was fixed here: https://github.com/racket/typed-racket/commit/62c86d5ddd246d7f6da77f6f52a2560d725bde24
and drracket/private/tooltip.rkt was converted to typed racket after that.  So it isn't due to you're changes, but it's something that was broken before you started working on this, and was fixed since then.

PR #47, opened 2015-03-13
Fix the types for pin-over and pin-under in typed/pict

lexi-lambda:
Self-explanatory; these types were just missing a parameter for each case.


samth:
Looks good to me.

PR #49, opened 2015-03-15
Make order-of-magnitude more specific

mromyers:
(order-of-magnitude n) is only negative if n < 1. Hence it is non-negative for all positive integers.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -87,7 +87,7 @@
 [real->floating-point-bytes (->opt -Real (one-of/c  4 8) [Univ -Bytes -Nat] -Bytes)]
 [system-big-endian? (-> B)]
 
-[order-of-magnitude (-> -PosReal -Int)]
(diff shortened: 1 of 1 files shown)


lexi-lambda:
I think the syntax for `cl->` is wrong here. I believe it should be this:

```racket
(cl-> [(-PosInt) -Nat]
      [(-PosReal) -Int])
```

Using `cl->*` seems more common, though, in which case it would be written like this:

```racket
(cl->*
 (-> -PosInt -Nat)
 (-> -PosReal -Int))
```

I'm not sure which is preferred, though.

TwoPoint7182818:
Oops, you're right:
```
(define-syntax cl->                                                                                                                                                       
  (syntax-parser                                                                                                                                                          
   [(_ [(dom ...) rng] ...)                                                                                                                                               
    #'(cl->* (dom ... . -> . rng) ...)]))
```

lexi-lambda:
This looks good to me. @samth?

stamourv:
Please squash the two commits. Otherwise, this looks good.

mromyers:
Commits squashed.

PR #52, opened 2015-03-19
Include more specific types for integers passed to exact->inexact, real->double-flonum, and real->single-flonum

lexi-lambda:
This just adds some additional cases to `exact->inexact` for `Positive-Integer` and `Negative-Integer`, which now yield `Positive-Flonum` and `Negative-Flonum`, respectively.


lexi-lambda:
I've updated this to also include more specific types for `real->double-flonum` and `real->single-flonum`.

stamourv:
That looks good. Could you run the random tester, say, 1M times before pushing? Just to be sure.

lexi-lambda:
I've updated this a little bit more. In either case, I think this is sound, but in running the random checker, I ran across a different but (I think) unrelated bug, #53.

PR #57, opened 2015-03-22
Modernize typed/net/url

lexi-lambda:
This updates `typed/net/url` with the following changes:

- Uses more modern TR forms instead of the older, legacy forms.
- Functions previously declared using `case->` now use `->*` when appropriate.
- Includes some additional functions from `net/url` that were missing.
- The `typed/net/url-structs` and `typed/net/url-connect` modules have been spun off to be equivalent to their untyped counterparts.
- The type alias `URL` has been added for the `url` struct to fit with naming conventions.

### DIFF
--- a/typed-racket-more/typed/net/url.rkt
+++ b/typed-racket-more/typed/net/url.rkt
@@ -1,86 +1,77 @@
 #lang typed/racket/base
 
-(require typed/private/utils)
+;; net/url-structs
+(require typed/net/url-structs)
+(provide (all-from-out typed/net/url-structs))
 
-(require-typed-struct/provide path/param ([path : (U String 'up 'same)] [param : (Listof String)]) net/url)
+;; opaque types
+(require/typed/provide
+ net/url
+ [#:opaque URL-Exception url-exception?]
+ [#:opaque HTTP-Connection http-connection?])
 
-(require-typed-struct/provide
- url ([scheme : (Option String)]
-      [user : (Option String)]
-      [host : (Option String)]
-      [port : (Option Integer)]
-      [path-absolute? : Boolean]
-      [path : (Listof path/param)]
-      [query : (Listof (Pair Symbol (Option String)))]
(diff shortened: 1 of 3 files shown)


lexi-lambda:
@samth Can I merge this, or do you have any comments?

typed-racket-more/typed/net/url.rkt
```diff
@@ -1,86 +1,70 @@
...
-      [fragment : (Option String)])
- net/url)
-
-(require/opaque-type URL-Exception url-exception? net/url)
-(provide URL-Exception url-exception?)
-
-(require/opaque-type HTTP-Connection http-connection? net/url)
-(provide HTTP-Connection http-connection?)
-
-(define-type-alias PortT (case-lambda (url -> Input-Port) (url (Listof String)-> Input-Port)))
```
samth: Why not keep these abstractions? At a minimum we should keep providing them, but they look useful for the subsequent definitions as well.

samth:
Other than that, looks good.

typed-racket-more/typed/net/url.rkt
lexi-lambda: This is why:

```racket
Welcome to Racket v6.1.1.
> (require typed/net/url)
> (:print-type get-impure-port)
PortT
>
```

As long as type alias printing works the way it does, providing type aliases needs to be done with a lot of caution. They effectively make the types opaque. As far as I know, there's no way to print the "underlying" type the alias refers to.

samth: You can use `:type` for this:

```
> (define-type S (-> Integer Integer))
> (:type S)
(-> Integer Integer)
[can expand further: Integer]
```

But the larger issue is that we can't just drop exported identifiers, since that may break people's programs.

lexi-lambda:
Merged in d2745f89eb7c67da0057b05d685caa0fa2adb099.

PR #59, opened 2015-03-23
More dependency reduction.

samth:
This is now implemented, and is roughly ready for review. Still to-do:
 - [x] run tests
 - [x] run `raco setup`

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
@@ -12,9 +12,8 @@
 ;; submodule is again to avoid dependency.
 
 (require
-  syntax/parse
-  (for-syntax racket/base racket/syntax
-              syntax/parse syntax/parse/experimental/template)
+  syntax/parse/pre
+  (for-syntax racket/base racket/syntax syntax/parse/pre)
   (for-label racket/base) ;; used for identifier comparison only
   (for-template racket/base))
 
@@ -42,7 +41,7 @@
     (begin
       (provide nms ...)
       (module* literal-set #f
-        (require syntax/parse)
+        (require syntax/parse/pre)
         (provide set-name)
         (define-literal-set set-name (nms ...)))
       (define-syntax (nms stx)
@@ -111,17 +110,18 @@
 
   (syntax-parse stx
     [(_ :clause ...)
-     (template
-       (begin
-         (begin
-           (define-syntax-class name
-             #:auto-nested-attributes
-             #:literal-sets ((internal-literals #:at name))
-             (pattern i:internal^ #:with (lit . body) #'i.value))
-           (define pred
-             (syntax-parser
-               [(~var _ name) #t]
-               [_ #f]))) ...))]))
+     (syntax
+      (begin 
+        (begin
+          (define-syntax-class name
+            #:auto-nested-attributes
+            #:literal-sets ((internal-literals #:at name))
+            (pattern i:internal^ #:with (lit . body) #'i.value))
+          (define pred
+            (syntax-parser
+              [(~var _ name) #t]
+              [_ #f])))
+        ...))]))
 
 
 (define-internal-classes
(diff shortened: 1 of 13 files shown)


stamourv:
LGTM.

FWIW, `syntax/parse/pre` seems to be undocumented.

samth:
@stamourv I checked with @rmculpepper and he said I can rely on it, but yeah, docs would be nice.

typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
```diff
@@ -111,17 +110,18 @@
...
-           (define-syntax-class name
-             #:auto-nested-attributes
-             #:literal-sets ((internal-literals #:at name))
-             (pattern i:internal^ #:with (lit . body) #'i.value))
-           (define pred
-             (syntax-parser
-               [(~var _ name) #t]
-               [_ #f]))) ...))]))
+     (syntax
+      (begin
```
takikawa: Can we retain the spacing here to reduce the diff? I think aside from `template` -> `syntax` this is just indentation.

typed-racket-lib/typed-racket/typed-racket.rkt
```diff
@@ -4,7 +4,7 @@
  (for-syntax racket/base racket/lazy-require
              "standard-inits.rkt")
  ;; these need to be available to the generated code
- "typecheck/renamer.rkt"
+ "typecheck/renamer.rkt" syntax/location
```
takikawa: Why is this import necessary? (nothing else in the file seemed to change?)

typed-racket-lib/typed-racket/utils/redirect-contract.rkt
```diff
@@ -1,6 +1,6 @@
 #lang racket/base
 
-(require syntax/modcollapse (for-template racket/base))
+(require syntax/private/modcollapse-noctc (for-template racket/base))
```
takikawa: How reliable is this require of a private module?

typed-racket-lib/typed-racket/typed-racket.rkt
samth: Because we produce generated code (for the blame object for provided identifiers) that uses `syntax/location` -- that's why it's here with the comment. Previously we got away without this because something else depended on `syntax/location`.

typed-racket-lib/typed-racket/utils/tc-utils.rkt
```diff
@@ -6,9 +6,10 @@ don't depend on any other portion of the system
 |#
 
 (require syntax/source-syntax "disappeared-use.rkt"
-         racket/promise racket/string
-         syntax/parse/pre (for-syntax racket/base syntax/parse/pre)
-         (only-in unstable/sequence in-slice))
+         racket/promise racket/string racket/lazy-require
+         syntax/parse/pre (for-syntax racket/base syntax/parse/pre))
+
+(lazy-require [unstable/sequence (in-slice)])
```
takikawa: It shouldn't be too hard to re-write `tc-error/fields` to just avoid this dependency entirely, I think. Though I guess it's not a big deal if it's only required on error?

typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
samth: Well, really we should fix `template`'s indentation.

typed-racket-lib/typed-racket/utils/tc-utils.rkt
samth: The point of this commit is for it not to be `require`d at runtime, so I don't think there's a problem here. I expect that the typechecker itself depends on `unstable/syntax` in a variety of other ways too.

The better solution would be to split `tc-utils` into a module needed only in the type checker and one that gets used by other modules.

typed-racket-lib/typed-racket/utils/redirect-contract.rkt
samth: Currently, `scribble/srcdoc` uses it as well.

PR #62, opened 2015-03-25
Fix expansion of keyword types in kw-convert

endobson:
I believe that this is correct but expensive. The tests seemed to pass except it didn't finish and was churning away, so I think this caused one test to take much longer. That is highly likely but I don't see a way around checking the body with all the types it could be called with.


samth:
Can you figure out which test it was that took a long time?

pnwamk:
moved to https://github.com/racket/typed-racket/pull/485

PR #63, opened 2015-03-25
Strengthen the types for member, memv, memq, and memf

### LINKED ISSUE #61: return type for `memf` too generous?
```racket
> memf
- : (All (a) (-> (-> a Any) (Listof a) (U False (Listof a))))
```

This return type allows `Empty` as a possible result, because `Empty` is a subtype of every `Listof ...`. Right? 

But if `memf` isn’t `False`, then it returns a positive-length list tail. Right? 

If so, perhaps there’s no need to permit `Empty`, and `memf` could use a tighter type that reflects this, e.g.:

```racket
- : (All (a) (-> (-> a Any) (Listof a) (U False (List* a (Listof a))))
```

lexi-lambda:
Fixes #61.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -624,13 +624,13 @@
 [kernel:reverse (-poly (a) (-> (-lst a) (-lst a)))]
 [append (-poly (a) (->* (list) (-lst a) (-lst a)))]
 [length (-poly (a) (-> (-lst a) -Index))]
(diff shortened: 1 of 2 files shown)


samth:
LGTM.

samth:
By the way, if you rebase and push to this branch before pushing to `racket/typed-racket:master`, then it will close this PR as merged.

lexi-lambda:
Well, I can push to `racket:master` right now without rebasing, so it'll mark it as merged anyway, but that's a cool trick. Nice to know, thanks!

PR #65, opened 2015-03-27
Initial version of improved types for numeric comparisons

shekari:
So looking at old bug reports I came across [pr14896](http://bugs.racket-lang.org/query/?cmd=view%20audit-trail&pr=14896), which showed that some idioms didn't work all that well because filter support wasn't that great.

Given that filters can now do intersections we can clean up a bunch of these cases. In particular we don't need a case for each sized positive numeric type. So putting this out here for general comments before going to deep into the cleanup.


stamourv:
That looks reasonable.
Could you add test cases that correspond to the old filters that you removed? That way we can make sure that the new ones subsume them.

shekari:
This has turned into many commits that I'll submit separately.

PR #68, opened 2015-03-28
Strengthen infinity checks in comparison operations.

shekari:
This adds checks for infinity when comparing against reals and accounts
for NaNs.


stamourv:
LGTM

PR #69, opened 2015-03-28
Make Place-Channel <: (Evtof Any)

takikawa:
This PR makes place channels a subtype of `(Evtof Any)`. There's probably no reason this wasn't done before except that I didn't know that place channels were valid synchronizable events.

One issue is that maybe the types of `place-channel-get` and `place-channel-put` should be more precise (to match the set of `place-message-allowed?` things) instead of `Any`. That would also change the event type.

That's not totally straightforward because place channels allow C pointer values (we don't have any types for those) and also immutable prefabs (we don't have a way to quantify over all immutable prefabs).

We should be able to adjust this type later though because `(Evtof Any)` is more conservative than the actual type.


samth:
LGTM.

takikawa:
Merged.

PR #72, opened 2015-03-28
Make zero? sound with regards to NaN.

shekari:


lexi-lambda:
Why does `-RealZero` contain `NaN`?

shekari:
@lexi-lambda I'm not sure behind the reasoning, but all flonum types conatin `NaN`.

samth:
Basically, very few closure properties hold on sets of floats that don't include `NaN`

stamourv:
What @samth said.

This PR looks good to me.

PR #74, opened 2015-04-01
Add typed/rnrs/sorting-6

lexi-lambda:
This is a small addition to `typed-racket-more`, adding a `typed/rnrs/sorting-6` module. Here's my reasoning for adding it:

- For some reason, Racket does not provide built-in `vector-sort` and `vector-sort!` functions. This does, and they operate on plain Racket vectors.
- Including this polymorphically with `require/typed` is impossible, since the wrapped value causes contract errors (this might actually be a bug with polymorphic vectors; I should look into that more).
- Even if it were possible (or if you require it with a more specific, non-polymorphic type), the generated contracts make these functions considerably more expensive than they should be.

Even if it's uncommonly used, I figure it can't hurt to stick in the library. I found myself wanting it, at least.

### DIFF
--- a/typed-racket-more/typed/rnrs/sorting-6.rkt
+++ b/typed-racket-more/typed/rnrs/sorting-6.rkt
@@ -1,0 +1,8 @@
+#lang s-exp typed-racket/base-env/extra-env-lang
+
+(require rnrs/sorting-6)
+
+(type-environment
+ [list-sort (-poly (a) (-> (-> a a Univ) (-lst a) (-lst a)))]
+ [vector-sort (-poly (a) (-> (-> a a Univ) (-vec a) (-vec a)))]
+ [vector-sort! (-poly (a) (-> (-> a a Univ) (-vec a) -Void))])


samth:
I think I'd prefer to add `vector-sort` and `vector-sort!` to `racket/vector`.

lexi-lambda:
Sure, I would, too. Do you think that's a good enough reason to omit this?

samth:
I think that if we have that in `racket/vector` (and in the appropriate type environment) then we shouldn't make `typed/rnrs/*` trusted.

lexi-lambda:
Actually, you're right. The implementation of `vector-sort!` *sucks*.

```racket
(define (vector-sort < v)
  (list->vector (sort (vector->list v) <)))

(define (vector-sort! < v)
  (let ([v2 (vector-sort < v)])
    (vector-copy! v 0 v2)))
```

That's more or less worthless, which I didn't really realize before trying this. I'll look into adding a better implementation to `racket/vector`.

samth:
I think `sort` actually has a `vector-sort!` implementation inside.

PR #77, opened 2015-04-01
Fix type restriction in values->tc-results

takikawa:
This PR should fix the current test failures on DrDr.

This also includes a subtyping change that I needed to make for the tests to pass.


andmkent:
LGTM. 

I wonder if at some point we don't want to just expand function subtyping to be as expressive as it is in the formalism so we don't have to have special cases like TypeFilter and NotTypeFilter in subtyping.

shekari:
LGTM.

I'm not sure what it is in the formalism, but we already have this function it more or less in `implied-atomic?`.

PR #78, opened 2015-04-02
Add comments for changes to substitution.

shekari:
@samth These look good?


samth:
Restriction is spelled wrong. ;-)  Otherwise looks good.

PR #79, opened 2015-04-02
Simplify the type for zero

shekari:
With the first commit I'm not sure how we would want to organize the code. Do we want to try and make the lazy-require not necessary?


samth:
@shekari I think the `lazy-require` is probably necessary -- `subst-type` ends up using `subtype`.

@andmkent can you review this?

stamourv:
The numeric parts LGTM.

shekari:
@andmkent any comments?

andmkent:
(sorry for not seeing this till now)

Looks great! 

I actually did something similar *just* for dependent function subtyping (this approach looks nicer than what I did). This should help reduce special casing for dependent functions.

PR #80, opened 2015-04-02
Add sealing contracts for row polymorphic types

takikawa:
This enables contract generation for row polymorphic types for values from untyped code.

Depends on the class sealing/unsealing operations that are in a PR for core Racket: https://github.com/plt/racket/pull/912

The implementation for static contracts is pretty close to `parametric->/sc`. The `utils/sealing-contract.rkt` file does the heavy lifting and implements a contract combinator that is related to `parametric->/c`.

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -147,6 +147,7 @@
       typed-racket/utils/any-wrap typed-racket/utils/struct-type-c
       typed-racket/utils/opaque-object
       typed-racket/utils/evt-contract
+      typed-racket/utils/sealing-contract
       unstable/contract racket/contract/parametric))
 
 ;; Should the above requires be included in the output?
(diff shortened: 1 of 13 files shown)


typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -472,28 +473,46 @@
...
-                          field-names (map t->sc/both field-types)))
-                   absents)]
+         ;; add a seal/unseal if there was a row variable and the
+         ;; row polymorphic function type was in negative position
+         (define seal/sc
+           (and (F? row-var)
+                (not (lookup-row-constraints (F-n row-var)))
+                (triple-lookup
+                 (hash-ref recursive-values (F-n row-var)
+                           (λ () (error 'type->static-contract
```
samth: This is an internal error, so it should use `int-err`.

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -472,28 +473,46 @@
...
+                            pubment-names (map t->sc/meth pubment-types))
+                       (map (λ (n sc) (member-spec 'inner n sc))
+                            augment-names (map t->sc/meth augment-types))
+                       (map (λ (n sc) (member-spec 'init n sc))
+                            init-names (map t->sc/neg init-types))
+                       (map (λ (n sc) (member-spec 'field n sc))
+                            field-names (map t->sc/both field-types)))
+                     absents))
+         (if seal/sc
+             (and/sc seal/sc sc-for-class)
```
samth: Is this new contract going to break existing code?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -695,8 +714,25 @@
...
+          (let loop ([ty b])
+            (match (resolve ty)
+              [(Function: _) #t]
+              [(Union: elems) (andmap loop elems)]
+              [(Poly: _ body) (loop body)]
+              [(PolyDots: _ body) (loop body)]
+              [_ #f])))
+        (unless function-type?
+          (fail #:reason "cannot generate contract for non-function polymorphic type"))
+        (let ((temporaries (generate-temporaries vs-nm)))
```
samth: `[]`

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -695,8 +714,25 @@
...
+          (define rv (for/fold ((rv recursive-values)) ((temp temporaries)
+                                                        (v-nm vs-nm))
+                       (hash-set rv v-nm (same (sealing-var/sc temp)))))
+          (parameterize ([bound-names (append (bound-names) vs-nm)])
+            (sealing->/sc temporaries (take constraints 3)
```
samth: Why 3?

typed-racket-lib/typed-racket/utils/sealing-contract.rkt
```diff
@@ -0,0 +1,127 @@
+#lang racket/base
+
+;; This module provides a sealing->/c which operates like the parametric->/c
```
samth: Should this be in `racket/contract`?

typed-racket-lib/typed-racket/static-contracts/combinators/parametric.rkt
```diff
@@ -54,3 +62,38 @@
...
+
+;; combinator for sealing-> contracts for row polymorphism
+(struct sealing-combinator combinator (vars members)
+  #:transparent
+  #:property prop:combinator-name "sealing->/sc"
+  #:methods gen:sc
+    [(define (sc-map v f)
+       (match v
+        [(sealing-combinator (list arg) vars members)
+         (sealing-combinator (list (f arg 'covariant)) vars members)]))
```
samth: Why are these covariant?

typed-racket-lib/typed-racket/private/type-contract.rkt
takikawa: It refuses to generate a contract right now, so it shouldn't.

typed-racket-lib/typed-racket/utils/sealing-contract.rkt
takikawa: I don't think it's likely to be that useful outside of Typed Racket.

typed-racket-lib/typed-racket/private/type-contract.rkt
takikawa: I should probably add a comment here. I think (but I'm not 100% sure yet) that the constraints for the row variable for augmenting methods are not actually necessary for the contract because it's non-sensical for a class type to have an augment method type without a corresponding public method type. So only the first three sets of constraints are needed.

PR #81, opened 2015-04-03
Simplify the type of =

shekari:
This PR removes a bunch of cases of = that are just restrictions that fall out from the more general cases.

First two commits are from an earlier pr but need the `RealZeroNoNan` simplification so was easier to leave them in.


stamourv:
Aside from the one comment, LGTM.

Did you notice any impact on typechecking time from these kinds of changes? Smaller types should hopefully make some things faster, though the extra time spent doing intersections may reduce those gains.

PR #85, opened 2015-04-06
Stop trying to represent rest objects because they are currently broken.

### LINKED ISSUE #83: wrong object when returning dotted terms?
This unit test will fail (it's not currently in the test suite):
```racket
[tc-e/t (lambda: ([x : One])
          (let ([f (plambda: (a ...) [w : a ... a] w)])
            (f x "hello" #\c)))
        (t:-> -One (-lst* -One -String -Char))]
```

with error message:
```racket
Typed Racket Tests > Unit Tests > Typechecker tests > tc-expr tests > 1112 
(lambda: ([x : One])
  (let ([f (plambda: (a ...) [w : a ... a] w)])
    (f x "hello" #\c)))
FAILURE
location:   <pkgs>/typed-racket-test/unit-tests/typecheck-tests.rkt:1112:8
actual:     (tc-results
 (list
  (tc-result
   (-> One (List One String Char) : (Top | Bot) ((0 0)))
   (Top | Bot)
   -))
 #f)
expected:   (tc-results
 (list (tc-result (-> One (List One String Char)) (Top | Bot) -))
 #f)
tc-expr did not return the expected value. It returned a more precise value. The types don't match.
--------------------
```

So, object of this function is incorrect. It should be empty, but is instead the first argument of the function.

shekari:
Closes #83.

This currently has a slight regression in the filters on functions that return the rest argument. I don't think this is a real issue though, because those types must always be non-`#f`, and so any conditional later will figure that out.

I'm still looking at what would be required to support rest args as objects, to see if we should just go straight there.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
@@ -238,11 +238,9 @@
       [(null? (syntax-e s)) (formals (reverse acc) #f stx)]
       [else (formals (reverse acc) s stx)])))
 
+;; Currently no support for objects representing the rest argument
 (define (formals->objects formals)
-  (for/list ([i (in-list (append (formals-positional formals)
-                                 (if (formals-rest formals)
-                                     (list (formals-rest formals))
-                                     empty)))])
+  (for/list ([i (in-list (formals-positional formals))])
     (make-Path null i)))
 
 
--- a/typed-racket-lib/typed-racket/typecheck/tc-metafunctions.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-metafunctions.rkt
@@ -13,16 +13,18 @@
          merge-tc-results
          tc-results->values)
 
-
+;; Objects representing the rest argument are currently not supported
 (define/cond-contract (abstract-results results arg-names #:rest-id [rest-id #f])
   ((tc-results/c (listof identifier?)) (#:rest-id (or/c #f identifier?))
    . ->* . SomeValues/c)
-  (define arg-names* (append arg-names (if rest-id (list rest-id) null)))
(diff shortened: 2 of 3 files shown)


shekari:
Ok actually supporting rest arguments is non-trivial. One of the underlying bugs is instantiating a poly-dotted function would need to remove the object from the result as would replicating a normal rest arg. So I think this is the best solution for now.

samth:
How could object for rest arguments make any sense? An object should describe a particular reference to something in the environment. A rest argument, by definition, is a newly-allocated list and so a property of it can't be a property of something already around.

Am I missing something here?

shekari:
They can alias existing things for example: `((lambda a (first a)) x)` should return the object `x`. I dont think this is very useful, but it does make sense.

samth:
That does make sense, but it would require a whole new kind of object in the system, which could represent data structure creation, instead of traversal. So I don't think we should worry about this until we see a real need for it.

andmkent:
As a side note, adding linear expressions as objects similarly breaks the
association between objects and locations in memory.

On Mon, Apr 6, 2015 at 11:13 AM, Sam Tobin-Hochstadt <

> That does make sense, but it would require a whole new kind of object in
> the system, which could represent data structure creation, instead of
> traversal. So I don't think we should worry about this until we see a real
> need for it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/racket/typed-racket/pull/85#issuecomment-90097719>.
>



-- 
-Andrew

samth:
@andmkent Yes, and we could do that same with a `cons` object, which is what would be needed for @shekari's example. I'm just skeptical that it would be useful.

shekari:
Any thoughts on the actual change?

andmkent:
Looks good to me.

samth:
Looks good.

PR #88, opened 2015-04-07
Simplify type for abs

shekari:
This also contains commits for changing tc-subst so that the simplification works.

@andmkent Can you review the first, and @stamourv can you review the second?

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env-numeric.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env-numeric.rkt
@@ -665,19 +665,16 @@
 
   (define abs-cases ; used both for abs and magnitude
     (list
-     (map unop (list -Zero -One -PosByte -Byte -PosIndex -Index -PosFixnum -NonNegFixnum))
+     (-NonNegReal . -> . -NonNegReal : -true-filter : (-arg-path 0))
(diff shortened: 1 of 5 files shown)


stamourv:
LGTM

andmkent:
LGTM -- I just had that one question about the usage of restrict.

shekari:
I think in the case where the function returns an argument at a type which cannot be restricted, the application expression should have the type that corresponds to the functions return value not the original argument type (which I believe is in agreement with what you said). This does not require 'orig because the type of the argument is the first argument to restrict.

andmkent:
@shekari Ah! That makes sense.

I was a little confused by the wording of the comment and did not think closely enough about the argument order it seems. The comment in the new code in this request describes what is going on as "a restriction of the returned type to the arguments type", but the definition of restrict describes (restrict t1 t2) as "restrict t1 to be a subtype of t2". Maybe this is just confusing to me and no one else?

Would a description like this be potentially less ambiguous but still accurate:

```racket

;; Substitution of objects into a tc-result
;; This is a combination of the other substitutions, where
;; additionally if the returned object corresponds to an argument's
;; object then the return type takes into account the argument's type.

```

shekari:
Well I think its `restrict` which is confusing. Other than the polymorphic case it is more or less symetrical and tries to be `intersect`. Improving the docs in that way is fine by me though.

PR #89, opened 2015-04-08
Numeric ops

endobson:
_Description: this PR primarily simplifies a bunch of the types for numeric operations by indicating that they're the identity function using the function type's latent object._

So these pass all the existing TR checks, but break the math library.

The issue is something like `(ann (map conjugate (list 1 2 3)) (Listof Integer))` and with the new type of conjugate it cannot infer the right type. But with either `(inst map Integer Integer)` or `(ann conjugate (-> Integer Integer))` (with another upcoming change) this works. But that is not a solution in the math scenario because it actually needs to work for multiple types at once (i.e `Flonum`, `SingleFlonum`, `InexactReal`).

Thoughts on how to proceed? I don't think without doing some tricky logic in infer will it be possible to get this to infer correctly, but I may be wrong.


samth:
I think the trickiness in `infer` will be required. Can you give a short example that we can put in the unit tests, though?

stamourv:
Aside from the inference regression and from my comments, LGTM.

samth:
@endobson or @pnwamk is the PR in a state that it could (or should) be revived? The internals have changed a bunch since 2015.

samth:
Closing this as likely in significant need of revision to revive.

PR #90, opened 2015-04-08
no objects for mutated private class fields

andmkent:
still need to double check and test real quick... but this should fix the bug let-aliasing helped expose by discovering mutated private fields were getting objects created for them.


samth:
1. Can you remove the whitespace changes (specifically, the new file is wrongly indented, and has extra newlines)?
2. Can we just fix `->acc` to use the smart constructors automatically, which would mean the original code would just start working?

andmkent:
I think I got both points you mentioned, @samth 

passes that new test just fine.

samth:
Looks good.

PR #116, opened 2015-04-11
Make flonum and ext-flonum types use the same definitions.

shekari:
This avoids problems where one could be changed to fix a bug and the
other would be missed.


stamourv:
Aside from my comments, LGTM.

PR #117, opened 2015-04-16
Make subtype use hash of hashs.

endobson:
This has a 1% improvement on new-metrics.rkt by my measurements.

Does anyone have any better benchmarks?


samth:
Why order this `st -> ss -> #t` instead of the other way around?

shekari:
Because ss is usually more precise, i.e. has filters in the check below cases. That was my reasoning, I don't remember if there was data to back it up.

samth:
Ok, seems fine.

PR #120, opened 2015-04-20
Make infer/dotted instantiate the dotted variable to improve inference.

endobson:
With this we don't need to infer the length of the dotted variable,
in parts of the inference.


samth:
Can you explain what's going on here some more? Why is this an improvement for that test case?

endobson:
Without the change the test case doesn't type check.

With infer/dots we already know how many terms the polydotted variable will need, this just does that work right at the begining so that the rest of inference doesn't need to do it, because it does it poorly.

PR #123, opened 2015-04-29
Simplify more numeric ops

endobson:
These are a bunch of simplifications and fixes that don't require the improved inference changes (which aren't written yet).


stamourv:
LGTM, modulo comments.

Issue #115, opened 2015-04-10
fx>= has broken type


endobson:
This typechecks:
```
#lang typed/racket
 
(require racket/flonum)

(ann
  (let ([x : Flonum 5.0])
    (if 
      (fl>= x (ann -4.0 Nonpositive-Flonum))
      x
      -3.0))
  Nonpositive-Flonum)
```

shekari:
This is also true for extfl>=.

takikawa:
I think this was fixed already, possibly in commit 7e32ec93eb3b0b32a8a88097630a1527740b3a0e

rfindler:
I don't see any test cases.....

PR #124, opened 2015-05-01
Fix typing of fl>= and extfl>=

miraleung:
Addressing issue #115, based on behaviour for fl<=.

### DIFF
--- a/typed-racket-test/succeed/flonum-typecheck.rkt
+++ b/typed-racket-test/succeed/flonum-typecheck.rkt
@@ -1,0 +1,123 @@
+#lang typed/racket
+(require racket/flonum
+         typed/rackunit)
+
+
+;;Flonum typechecking tests
+;; -FlZero -Fl
+(check-eq?
+  (ann (let ([x : Flonum-Zero 0.0])
+         (if (fl>= x (ann -4.0 Flonum))
+           x -3.0))  Flonum)
+  0.0)
+(check-eq?
+  (ann (let ([x : Flonum-Zero 0.0])
+         (if (fl<= x (ann -4.0 Flonum))
+           -3.0 x))  Flonum)
+  0.0)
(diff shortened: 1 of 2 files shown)


typed-racket-test/succeed/flonum-typecheck.rkt
```diff
@@ -0,0 +1,123 @@
+#lang typed/racket
+(require racket/flonum
+         typed/rackunit)
+
+
+;;Flonum typechecking tests
+;; -FlZero -Fl
+(check-eq?
```
shekari: I'm confused at what these cases are testing. It seems like this issue was about filters but this is testing that in a very indirect manner, by making sure that the optimizer doesn't do dead code elimination on the wrong branch. Would these be better tested by the unit tests for typechecking?

miraleung: Those tests are primarily aimed at regression testing the typechecks, but they can go if they are superfluous. I couldn't find a good way of "check-expecting" for type errors in the unit tests (check-exn doesn't catch those), hence the block at the end.

It looked to me like this issue concerned fl>= typechecks that shouldn't be passing, which this commit covers.

shekari: I think the issue is that you were looking in the wrong file. This file is not a unit test (as we call it) but an integration test, as it checks the entire TR system e2e. The unittests for typechecking are located here: https://github.com/racket/typed-racket/blob/master/typed-racket-test/unit-tests/typecheck-tests.rkt. And those much more easily support the kind of test you are trying to write, and support negative tests.

miraleung:
Since the issue was open, I didn't see that it had already been dealt with. Thanks for letting me know about the tests; now moved to the correct file in a separate pull request.

PR #125, opened 2015-05-02
Unit tests for fl>= typechecking

miraleung:
Tests for issue #115 and followup to pull request #124.

### DIFF
--- a/typed-racket-test/unit-tests/typecheck-tests.rkt
+++ b/typed-racket-test/unit-tests/typecheck-tests.rkt
@@ -3634,113 +3634,107 @@
        [tc-e ((inst second Any Any Any) (list "a" "b")) -String]
        [tc-e/t (abs 4) -PosByte]
 
-       ;; PR 124: Tests for flonum typechecking
-       [tc-e
-         (ann (let ([x : Flonum-Zero 0.0])
-                (if (fl>= x (ann -4.0 Flonum))
-                  x -3.0))  Flonum)
-         -Fl]
-        [tc-e
-          (ann (let ([x : Flonum-Zero 0.0])
-                 (if (fl<= x (ann -4.0 Flonum))
-                   -3.0 x))  Flonum)
-          -Fl]
-
-        ;; -PosFl -Fl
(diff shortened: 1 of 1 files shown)


shekari:
To actually run the unit tests there are a couple of ways.
```racket typed-racket-test/run.rkt --unit``` or ```racket typed-racket-test/unit-tests/typecheck-tests.rkt``` the former runs all the unit tests where the later runs just the typecheck tests.

**Comment by miraleung** [comment]
Thanks Eric. I was using `raco test` as described [here](http://blog.racket-lang.org/2012/11/tutorial-contributing-to-racket.html), so I didn't catch those errors. The [Racket style guide](http://docs.racket-lang.org/style/Units_of_Code.html) didn't mention anything about whitespace and my editor is set to automatically remove them, but I'll make sure these don't get changed in the future. 

All the new tests in my latest push pass, although I'm not sure whether some of the `fl>=` ones should be passing, such as:

```
[tc-e/t 
  (lambda: ([x : Flonum] [y : Flonum-Zero])
    (values (fl<= x y) (fl>= x y)))
  (t:-> -Flonum -FlonumZero
        (make-Values (list
                       (-result -Boolean (-FS (-filter -NonPosFlonum 0) -top))
                       (-result -Boolean)))]
```
Should the last line be `(-result -Boolean (-FS (-filter -NonNegFlonum 0) -top))`?

endobson:
I'm confused on how those tests are passing for you because they are not for me, and don't seem like they should pass.

https://gist.github.com/endobson/b8d141a04c4a317dc696

Other than that this is looking good, you may also want to rebase this ontop of the current HEAD to pick up the changes that were made to the comparison operations.

Also minor nit, but predicate is usually in reference to one argument procedures I would refer to these as 'comparison' operators/functions.

miraleung:
Thanks Eric, those tests are fixed now. There was one that seemed to be passing for you but wasn't for me (line 3719), and since the types looked off, I updated that one as well.

samth:
Merged.

PR #126, opened 2015-05-02
Add unsafe require/provide operations

takikawa:
This adds a new `typed/racket/unsafe` library that is not loaded by default which provides unsafe operations. Right now it just provides `require/typed/unsafe` and `provide/unsafe`.

People have asked for a feature like this for a while (with a revived discussion on it yesterday), so here's a first attempt at it.

Design questions: should `#:opaque` clauses also be unsafe? Currently they are not in this PR, but they could be. The reason I didn't make them unsafe is because opaque types seem very pointless without the predicate check.

Other concerns:
  * should `typed/racket/unsafe` be reserved for a #lang that is unsafe instead?
  * should there be more kinds of unsafety provided?
  * is `unsafe-` prefix better?

CC @ntoronto as a potential major user.

[Review on Reviewable](https://reviewable.io/reviews/racket/typed-racket/126)

### DIFF
--- a/typed-racket-test/succeed/unsafe-require.rkt
+++ b/typed-racket-test/succeed/unsafe-require.rkt
@@ -1,0 +1,32 @@
+#lang racket/base
+
+;; Test unsafe require forms
+
+(module a racket/base
+  (struct foo (x y))
+  (define a-foo (foo 1 2))
+  (provide (struct-out foo) a-foo))
+
+(module b typed/racket
+  (require/typed racket/contract/combinator [#:opaque Blame exn:fail:contract:blame?])
+  (require typed/racket/unsafe)
+  (require/typed/unsafe (submod ".." a)
+                        [#:struct foo ([x : String] [y : String])]
+                        [a-foo foo])
+
+  ;; UNSAFE
+  ;; primitive error, no blame should be raised
+  (with-handlers ([(negate exn:fail:contract:blame?) void])
+    (string-append (foo-x a-foo))))
+
+(module c typed/racket
+  (require/typed racket/contract/combinator [#:opaque Blame exn:fail:contract:blame?])
+  (require typed/racket/unsafe)
+  (require/typed/unsafe racket/base
+                        [string-append (-> String String Integer)])
+
+  ;; UNSAFE
+  (with-handlers ([(negate exn:fail:contract:blame?) void])
+    (number->string (string-append "foo" "bar"))))
+
+(require 'b 'c)
--- a/typed-racket-doc/typed-racket/scribblings/reference/unsafe.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/unsafe.scrbl
@@ -1,0 +1,30 @@
+#lang scribble/manual
+
+@begin[(require (for-label (only-meta-in 0 [except-in typed/racket for])))]
+
+@title{Unsafe Typed Racket operations}
+
+@defmodule[typed/racket/unsafe]
+
+@bold{Warning}: the operations documented in this section are @emph{unsafe},
+meaning that they can circumvent the invariants of the type system. Unless the
+@racket[#:no-optimize] language option is used, this may result in unpredictable
+behavior and may even crash Typed Racket.
+
+@defform[(require/typed/unsafe m rt-clause ...)]{
+  This form requires identifiers from the module @racket[m] with the same
+  import specifications as @racket[require/typed].
+
+  Unlike @racket[require/typed], this form is unsafe and will not generate
+  contracts that correspond to the specified types to check that the values
+  actually match their types.
+}
+
+@defform[(provide/unsafe provide-spec ...)]{
+  This form declares exports from a module with the same syntax as
+  the @racket[provide] form.
+
+  Unlike @racket[provide], this form is unsafe and Typed Racket will not generate
+  any contracts that correspond to the specified types. This means that uses of the
+  exports in other modules may circumvent the type system's invariants.
+}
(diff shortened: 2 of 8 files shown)


ntoronto:
Regarding more unsafety: One of the things I find most valuable about TR is safety. I think of these new forms as necessary evils, so I'd rather minimize the number of them.

Regarding the possible bikeshed discussion: I read "/unsafe" as "with unsafe," which makes them fine to me as-is. I also thought of the more explicit `require/typed/no-contract` and `provide/no-contract`, but those don't seem scary enough.

I'm curious: Do you regard `provide/unsafe` as more dangerous than `require/typed/unsafe`?

Relatedly, what's the worst thing that each can allow to happen?

takikawa:
> I'm curious: Do you regard provide/unsafe as more dangerous than require/typed/unsafe?

I think in theory they allow the same kinds of problems, but in practice one may end up being more dangerous. Probably `provide/unsafe` because I imagine you will generally debug your own code when using `require/typed/unsafe` but the former exposes you to arbitrary clients (of course you can get this with required higher-order functions too).

> Relatedly, what's the worst thing that each can allow to happen?

Both allow arbitrary behavior with potential crashes. For example, you can use `require/typed/unsafe` to get a value like an integer and pretend it's a vector. After a dereference you may get a segfault if the optimizer is on. With `provide/unsafe`, you can have untyped clients give you inputs that would do the same thing (provides an integer argument where a vector was expected).

ntoronto:
Is there a variant of the following bad program that uses `require/typed/unsafe`?

```racket
#lang typed/racket

(provide/unsafe add-one)

(: add-one (-> Integer Integer))
;; Looks safe, but isn't because the optimizer removes the unreachable else branch
(define (add-one n)
  (if (exact-integer? n)
      (+ n 1)
      (raise-argument-error 'add-one "Integer" n)))
```

takikawa:
Yes, if going higher-order counts as a variant:

````
#lang racket/base

(module a racket/base
  (define (f g) (g "foo"))
  (provide f))

(module b typed/racket
  (require typed/racket/unsafe)
  (require/typed/unsafe (submod ".." a)
                        [f (-> (-> Integer Integer) Integer)])

  (: add-one (-> Integer Integer))
  (define (add-one n)
    (if (exact-integer? n)
        (+ n 1)
        (raise-argument-error 'add-one "Integer" n)))

  ;; Error
  (f add-one))

(require 'b)
````

**Comment by samth** [comment]
Small comments:

* I prefer `unsafe-` as a prefix.
* I think the library should have a suffix, the way `racket/unsafe/ops` does.
* The exports should be protected, as in `ffi/unsafe`.

Larger comments:

I'm sad that we need this -- instead of types on one library putting pressure on other people to type their code, instead it just creates problems for people, and thus people are working around this in worse ways. We clearly haven't figured out the right way to give people contracts that they can use effectively.

Also, I worry that people won't be nearly as careful with this as they are when you have to write `unsafe-vector-ref` everywhere to get unsafety, and thus we'll end up with more segfaults.

**Comment by lexi-lambda** [comment]
I think there are three discrete levels of safety I've seen various people looking for.

1. Contracts + the optimizer. This is what TR current provides. Assuming soundness, this provides complete safety while also ensuring maximum speed without crossing a typed-untyped boundary.
2. No contracts + the optimizer. This is what this PR provides. It's the fastest, but it can also lead to segfaults if untyped code misbehaves.
3. No contracts + no optimizer. This is in between the two options, and currently, it isn't supported. It's also the approach Matthew Butterick seems to have opted for in his recent blog post on providing both typed and untyped versions of the same library using conditional compilation.

I think having number 2 is an improvement over forcing people to use the undocumented interface, but I also wonder if it would be helpful to provide number 3 as well to prevent people from disabling contracts entirely and subsequently calling segfaults without any clarity about where the errors are coming from.

**Comment by AlexKnauth** [comment]
For option 3, would it be possible for `#lang typed/racket #:no-optimize` to communicate with the unsafe require and provide forms and have them raise syntax errors if the optimizer isn't turned off?  Or would this still create problems if there are multiple modules and one of them uses #:no-optimize and the other doesn't?  Would there be a way for this unsafe require to define the imports as forms that don't allow themselves to be used unless the optimizer is turned off? Or as forms that use contracts in that situation? Does that seem like a good solution?

**Comment by samth** [comment]
I'm not a fan of option 3. If you want to use `require/unsafe` in your program, you should think long and hard about the preconditions that you're not checking, just as you would do with `unsafe-vector-ref`.

Also, note that other downstream modules may have optimization turned on, so using `require/unsafe` means that you have to consider them as well.

**Comment by lexi-lambda** [comment]
That's a reasonable opinion (and I agree with it in principle), but it does ignore Butterick's current use case. Admittedly, that probably *should* be solved by figuring out how to generate contracts that fit his needs, though I don't believe that blog post clarified why the contract generation was insufficient.

File: typed-racket-test/succeed/unsafe-require.rkt [hunk 1]
```diff
@@ -0,0 +1,32 @@
+#lang racket/base
+
+;; Test unsafe require forms
+
+(module a racket/base
+  (struct foo (x y))
+  (define a-foo (foo 1 2))
+  (provide (struct-out foo) a-foo))
+
+(module b typed/racket
```
bennn: Note: adding a line `(provide (struct-out foo))` here raises an exception:
```
struct-out: multiple bindings (foo4 and foo3) for structure-type identifier
  at: foo4
  in: (struct-out foo)
```

Instead, you need to provide the identifiers individually. But be careful, because a typed module can't import them straight away. Adding two lines to `module c`:
```
(require (submod ".." b))
(foo-x (foo "yo" "lo"))
```

Raises some require/typed errors:
```
Type Checker: missing type for identifier;
  identifier: foo-x
```

takikawa:
Merged, also fixed the struct issue.

PR #129, opened 2015-05-09
syntax error for require/typed in untyped code

AlexKnauth:
see http://www.mail-archive.com/racket-users@googlegroups.com/msg26916.html

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -140,7 +140,7 @@
 
 
   (define ((r/t-maker legacy) stx)
-    (unless (unbox typed-context?)
+    (unless (or (unbox typed-context?) (eq? (syntax-local-context) 'module-begin))
       (raise-syntax-error #f "only allowed in a typed module" stx))
     (syntax-parse stx
       [(_ lib:expr (~var c (clause legacy #'lib)) ...)
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -140,6 +140,8 @@
 
 
   (define ((r/t-maker legacy) stx)
+    (unless (unbox typed-context?)
+      (raise-syntax-error #f "only allowed in a typed module" stx))
     (syntax-parse stx
       [(_ lib:expr (~var c (clause legacy #'lib)) ...)
        (when (zero? (syntax-length #'(c ...)))


samth:
Looks right to me, but the error message could be a little more informative.

AlexKnauth:
How? What other information would you want to provide?

samth:
I just worry that "typed context" won't be clear to everyone.

AlexKnauth:
Can you think of a better way of putting it?
Can only be used with a typed language, maybe?
And anyway, it would be clearer than not giving a syntax-error and then putting out a runtime contract violation about promising `#f`, right?

samth:
How about "in a typed module"?

AlexKnauth:
Is this good?

samth:
Looks good to me.

AlexKnauth:
Um, are you going to merge this?

samth:
Yes, soon.

samth:
Unfortunately, this fails several tests on my machine, as follows:

```
[samth@huor:~/.../extra-pkgs/typed-racket/typed-racket-test (master) plt] r succeed/pr11171.rkt
succeed/pr11171.rkt:10:2: require/typed: only allowed in a typed module
  in: (require/typed (quote UNTYPED) (struct IntTree ((elem : Integer) (left : IntTree) (right : IntTree))))
  context...:
   /home/samth/sw/plt/extra-pkgs/typed-racket/typed-racket-lib/typed-racket/base-env/prims-contract.rkt:142:2
```

AlexKnauth:
Then how did the Travis CI build pass?

samth:
The Travis tests don't run the tests, unfortunately.

AlexKnauth:
Could that test be failing because there's no #%module-begin around it?

AlexKnauth:
Why doesn't Travis run the tests?

samth:
Travis doesn't run the tests because it's non-trival to set it up right. Just installing TR and all its dependencies from a minimal install takes longer than 40 minutes on Travis, so we can't do that. The alternative is to start from a regular snapshot, but that requires trickery to install all the packages correctly, which I haven't made work yet.

AlexKnauth:
Would raco pkg install --update-deps work, or are the catalogs for snapshots not set up for that?  If I remember correctly, they aren't, but if they were, would that work?

samth:
The problem is that we need to install several packages at once, and not from the catalog, but from the local checkout.

AlexKnauth:
Oh, do you mean install `typed-racket-lib`, `,typed-racket-more`, `typed-racket-test` etc.?
Or do you mean the dependancies with the correct versions?

samth:
I mean the former. We have a checkout w/ 5 packages in it, and a racket install with different versions of those packages.

AlexKnauth:
Why wouldn't raco pkg update with the directories work then?

samth:
You need to do them all simultaneously (plus, what if a pull request adds a new package?). Anyway, if you have an idea, just creating a TR pull request is enough to test it -- I haven't found a good solution yet, but maybe you will.

samth:
Merged.

PR #131, opened 2015-05-09
Improve typecheck for private field initializers

m4burns:
This patch addresses two issues with `typed/racket/class`:

1: For multiple private fields declared with `define-values`, type information does not propagate from the values produced by the initialization expression to the declared fields. This breaks soundness of private fields: A field can be annotated with a type that does not contain the field's initial value.

This was resolved by keeping a table of temporary bindings introduced in the expansion of the initializer along with their types. The field setter's type is then checked against that of the corresponding temporary.

2: The class body typechecker assumes that the `expr` of a `define-values` clause will expand to a bare `(values vs ...)`.

This was resolved by generalizing the template for matching an expanded `define-values` initializer and extracting the type information from the `expr` instead of each element in `(vs ...)`.


takikawa:
Thanks! For the tests, can you add them as unit tests in the "class-tests.rkt" file? (https://github.com/racket/typed-racket/blob/master/typed-racket-test/unit-tests/class-tests.rkt)

Also, can you add a comment describing the purpose of the `inits-temporaries-types` variable?

m4burns:
How's this?

takikawa:
Looks good to me. The tests pass for me too, so I'll squash the commits and push. Thanks.

PR #132, opened 2015-05-11
Disable the #{} reader macro if a dispatch macro already exists on #\{

lexi-lambda:
I ran into a problem when using TR with my [curly-fn](https://github.com/lexi-lambda/racket-curly-fn) meta-language, which installs its own dispatch macro on `#\{`. Since TR already uses that for annotations, this was overridden.

This implements @samth's idea thrown out on IRC to simply disable installing the readtable if a mapping already exists for `#\{`. This should never conflict with existing code, since the only way to modify a `#lang` readtable is through meta-langs, which are fairly rare as-is.

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
@@ -497,13 +497,21 @@ for function types.
 the types @racket[t], and also provides all of the @racket[v]s.}
 
 @defform/none[#{v : t}]{ This declares that the variable @racket[v] has type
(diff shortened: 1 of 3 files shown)


samth:
This needs a test, and probably a mention in the docs. Otherwise, looks good.

lexi-lambda:
Added a test and notes in the documentation.

samth:
Modulo that one comment, looks good to me.

lexi-lambda:
Okay, I linked to the readtable term in the docs, but I don't think it's possible to link to the `'dispatch-macro` index.

samth:
Looks good, thanks!

PR #134, opened 2015-05-16
Allow the types created for structs to be specified manually instead of always generating them from the struct name

lexi-lambda:
This change affects all the struct definition forms as well as `#:struct` clauses in `require/typed`. It's something I've been wanting to change for a while, but I've only now gotten around to doing it. The change is simple and can probably best be explained in code.

```racket
(struct [point : Point] ([x : Real] [y : Real]))
```

This creates a struct called `point`, but the *type* generated for it is `Point`. The change to `require/typed` is the same.

```racket
(require/typed
 net/url-structs
 [#:struct [path/param : Path/Param]
           ([path : (U String 'up 'same)]
            [param : (Listof String)])])
```

Capitalization is, unsurprisingly, the main motivator behind this change, which lets both naming conventions be satisfied without needing any type aliases.

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
@@ -378,16 +378,19 @@ those functions.
 
 
 @section{Structure Definitions}
-@defform/subs[
+@defform/subs[#:literals (:)
 (struct maybe-type-vars name-spec ([f : t] ...) options ...)
 ([maybe-type-vars code:blank (v ...)]
- [name-spec name (code:line name parent)]
+ [name-spec name-id (code:line name-id parent)]
  [options #:transparent #:mutable #:prefab
           (code:line #:constructor-name constructor-id)
-          (code:line #:extra-constructor-name constructor-id)])]{
- Defines a @rtech{structure} with the name @racket[name], where the
+          (code:line #:extra-constructor-name constructor-id)
+          (code:line #:type-name type-id)])]{
+ Defines a @rtech{structure} with the name @racket[name-id], where the
  fields @racket[f] have types @racket[t], similar to the behavior of @|struct-id|
- from @racketmodname[racket/base].
+ from @racketmodname[racket/base]. If @racket[type-id] is specified, then it will
+ be used for the name of the type associated with instances of the declared
+ structure, otherwise @racket[name-id] will be used for both.
   When @racket[parent] is present, the
 structure is a substructure of @racket[parent].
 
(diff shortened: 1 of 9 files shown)


samth:
I like this idea, but I think it should probably be a new keyword option to `struct`, rather than adding additional syntax around the struct name.

lexi-lambda:
Sure, that works. Just add `#:type-id` or `#:type-name` as a one of the struct options?

samth:
I like `#:type-name`.

samth:
@lexi-lambda ping

lexi-lambda:
@samth I updated this to use `#:type-name` for both cases. Let me know if you have any comments—I’ll squash the commits and do a little git housekeeping once everything’s reviewed.

typed-racket-lib/typed-racket/base-env/prims-struct.rkt
```diff
@@ -81,12 +81,22 @@
...
                                               (~bind [cname #f]))))
+                        (~optional (~seq #:type-name type:id))
                         ;; FIXME: unsound, but relied on in core libraries
                         ;; #:guard ought to be supportable with some work
                         ;; #:property is harder
                         (~optional (~seq #:guard guard:expr))
                         (~seq #:property prop:expr prop-val:expr))
-                   ...)))
+                   ...)
+             #:attr untyped #`(#,@(if (attribute mutable?) #'(#:mutable) #'())
```
lexi-lambda: @samth This bit right here is sort of icky. I’d appreciate any tips on how to do this without the brittle repetition.

samth: The simplest way would be to use a parameterized splicing syntax class that has the keyword in a list as an attribute, if present, and an empty list otherwise. Also binding it to an attribute of `...` depth 1 would let you write the results as something like:

```
mutable?.kw ...
transparent?.kw ...
```
and so on.

typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
```diff
@@ -408,11 +411,13 @@ amount it needs.
...
-from @racketmodname[racket/base].
+from @racketmodname[racket/base] (with the exception of @racket[#:type-name], as
+described above).
 
-A prefab structure type declaration will bind the given @racket[name] to a
-@racket[Prefab] type. Unlike in @racketmodname[racket/base], a non-prefab
-structure type cannot extend a prefab structure type.
+A prefab structure type declaration will bind the given @racket[name-id]
+or @racket[type-id] to a @racket[Prefab] type. Unlike in
+@racketmodname[racket/base], a non-prefab structure type cannot extend
```
samth: I think this should say something like "unlike the `struct` form from `racket/base`" with the appropriate link.

typed-racket-lib/typed-racket/base-env/prims-struct.rkt
```diff
@@ -99,13 +109,25 @@
...
                                       #:property prop:procedure proc*)))]
-       [dtsi (quasisyntax/loc stx (dtsi/exec* () nm (fld ...) proc-ty))])
-      #'(begin d-s dtsi))]))
+       [stx-err-fun (if (not (free-identifier=? #'nm.name #'type))
+                        (syntax/loc stx
+                          (define-syntax type
+                            (lambda (stx)
+                              (raise-syntax-error
+                               'type-check
+                               "type name used out of context"
```
samth: Can we say what type name this is in the error message, along the lines of @jsmaniac's recent change?

typed-racket-lib/typed-racket/base-env/prims-struct.rkt
```diff
@@ -157,18 +179,28 @@
...
+                      [d-s (ignore (quasisyntax/loc stx
+                                     (struct #,@(attribute nm.new-spec) (fs.fld ...)
+                                       . opts.untyped)))]
+                      [stx-err-fun (if (not (free-identifier=? #'nm.name #'type))
+                                       (syntax/loc stx
+                                         (define-syntax type
+                                           (lambda (stx)
+                                             (raise-syntax-error
+                                              'type-check
+                                              "type name used out of context"
```
samth: Same here.

samth:

typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
lexi-lambda: This text pretty much didn’t change in this PR, but I can make that adjustment here, anyway.

typed-racket-lib/typed-racket/base-env/prims-struct.rkt
lexi-lambda: I’m not sure I understand. Could you show me a simple example?

samth: Just composing this in the text window, but something like this:

```
(define-splicing-syntax-class (opt-kw kw)
  [pattern (~seq k:keyword)
    #:when (eq? kw (syntax-e #'k))
    #:with (kw ...) #'(k)]
  [pattern (~seq)
    #:with (kw ...) #'()])
```

and then
```
(~var mutable (opt-kw '#:mutable))
```

Issue #118, opened 2015-04-17
for/last is broken and doesn't typecheck


lexi-lambda:
I cannot get a single valid use of `for/last` to properly typecheck. Even this simple code fails.

```racket
(for/last ([i (in-range 10)]) i)
```

```
Type Checker: type mismatch
  expected: False
  given: Fixnum in: (for/last ((i (in-range 10))) i)
```

shekari:
This is well documented in the definition of for/last:, but maybe we should bring it up to the user level documentation.

https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/base-env/prims.rkt#L386

The issue is that we need an annotation on the initial value to the loop so that we can correctly infer it, instead of just assuming `#f` because that is the initial value.

PR #137, opened 2015-05-18
Clarification of `for/` forms

mbutterick:
Make it explicit in the docs that  `for/vector`, `for/flvector`, `for/first`, `for/last`, and `for/and` aren't yet supported by the typechecker.


mbutterick:
Per a suggestion in issue #118.

AlexKnauth:
It seems like for/vector is supported though.

PR #139, opened 2015-05-19
change type of partition

AlexKnauth:
This change allows partition to accept a predicate that only takes the types that are in the list (or some super-set of that), and allows these predicates to return Any instead of Boolean, and it only cares about the positive side of the filter.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -703,8 +703,9 @@
                    . ->... .
                    -Index))]
 [partition
- (-poly (a b) (cl->* (-> (-> a Univ) (-lst a) (-values (list (-lst a) (-lst a))))
-                     (-> (make-pred-ty a) (-lst b) (-values (list (-lst a) (-lst b))))))]
+ (-poly (a b) (cl->*
+               (-> (asym-pred b Univ (-FS (-filter a 0) -top)) (-lst b) (-values (list (-lst a) (-lst b))))
+               (-> (-> a Univ) (-lst a) (-values (list (-lst a) (-lst a))))))]
 
 [last   (-poly (a) ((-lst a) . -> . a))]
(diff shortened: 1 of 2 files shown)


samth:
Can you add a test of the new functionality?

AlexKnauth:
I tried adding this test to typecheck-tests.rkt
```
        [tc-e (let ([pos-not-5? : [-> Real (U Positive-Real #f) : #:+ Positive-Real]
                                (λ (x) (and (positive? x) (not (= x 5)) x))])
                (partition pos-not-5? '(1 2 -3 4 5 -6)))
              (-values (list -PosReal -Real))]
```
But I got an failure including this error:
```
. . ../../../../../../../../../../Applications/Racket/2015-05-17/Racket v6.2.0.3/collects/racket/private/for.rkt:585:2: in-list: contract violation
  expected: list?
  given: (values (U 1 Byte-Larger-Than-One Positive-Index-Not-Byte Positive-Fixnum-Not-Index Positive-Integer-Not-Fixnum Positive-Rational-Not-Integer Float-Nan Positive-Float-No-NaN Single-Flonum-Nan Positive-Single-Flonum-No-Nan) (U 0 1 Byte-Larger-Than-...
```
What am I doing wrong?

AlexKnauth:
Ok, I figured out that I'm not supposed to use `-values` for this.  I also realized that I put `-PosReal` and `-Real` where I should have put `(-lst -PosReal)` and `(-lst -Real)`.

AlexKnauth:
I also had to switch the order of the cases in the type of partition to get it to infer the right type without an explicit annotation.

samth:
Looks good to me.

When committing this, try to avoid creating a merge commit (usually done by rebasing before pushing).

AlexKnauth:
I don't think rebasing first helps, does it? That's what I've done before for pull requests (and that's also what I did for this one) and it still creates the merge commit.  I can try pushing directly to the racket/typed-racket repo though.

PR #141, opened 2015-05-20
Fix the type of abs on negative zeros.

endobson:


samth:
Looks good. @stamourv what do you think?

stamourv:
LGTM

stamourv:
Merged, thanks!

PR #142, opened 2015-05-21
Initial support for typed units in typed racket.

dfeltey:
This pull request is for review only.

This pull request adds initial support for using units in typed racket.
Most unit forms are supported, including most of the "infer" forms that
infer imports/exports/linkages from the current context.

Known Issues:
`define-values/invoke-unit/infer` doesn't currently typecheck correctly,
because of the way that linkages and exports propogate seem to differ
from how `compound-unit/infer` is described in the documentation.
I think this issue may also affect `invoke-unit/infer`.


takikawa:
The travis build seems to have failed. Looks like a unit test is broken?

takikawa:
This will need docs before merging as well. A reference entry is enough for now (for both the types and additional macro forms you introduced)..

dfeltey:
I think this is probably due to my changes to unit contracts, both of the
tests pass locally, and I think the only difference would be my unit
contract implementation.

takikawa:
The syntax `(Unit (import) (export))` seems to default to a unit return type of `Void`. Is this shortcut useful in practice? It seems potentially confusing, especially if you specify that type on a unit expression that actually returns a result and TR reports that it expected `Void`.

There are some errors that TR should catch in `Unit` type parsing. For example:

````
-> (: x (Unit (import Foo) (export) String))
-> x
; Rep-stx: contract violation
;   expected: Rep?
;   given: #f
; [,bt for context]
````

Maybe a related bug?

````
Welcome to Racket v6.2.0.3.
-> (define-signature foo^ ([x : (Unit (import foo^) (export) String)]))
-> (define x (unit (import foo^) (export) "foo"))
-> (invoke-unit x (import foo^))
; Rep-stx: contract violation
;   expected: Rep?
;   given: #f
; [,bt for context]
````

Is it possible to provide a better error when a user does `invoke-unit` without the right `import` signatures? (e.g., leaves them out entirely)

**Comment by dfeltey** [comment]
1. I think the defaulting to Void is useful since most units in Racket seem
to contain only definitions, so it somewhat lightens the syntactic burden
when needing to annotate units.

2. Yes, parsing a unit type should check if the signature is in the
environment first, I will fix that. In the invoke example, though, do you
think something like that should be allowed? I mean should signatures be
allowed to contain variables with types that reference the signature? I
think this could be done, but I'm not sure if it's useful.

3. Regarding `invoke-unit` error messages, if the expression being invoked
is a unit should the error message say which imports were not provided, but
were needed? Would that be better than the type-mismatch error that is
currently given?

**Comment by takikawa** [comment]
Re 2: I am ok with that failing, but it should have a better error messages. I stumbled into this case by accident, not because I wanted the recursive reference in a real program.

Re 3: that would probably be better.

**Comment by takikawa** [comment]
I tried exporting a unit like this:

````
#lang typed/racket

(define-signature s^
  ([a : Integer] [b : String]))

(define-signature t^
  ([c : Integer] [d : String]))

(define u@
  (unit
    (import s^)
    (export t^)

    (define c (add1 a))
    (define d (string-append b "bar"))
    
    c))

(provide u@)
````

and got an error like this:

````
<pkgs>/typed-racket-lib/typed-racket/static-contracts/combinators/unit.rkt: ?: unexpected term
  at: g12
  in: ((import (s^ (a g10) (b g11))) (export (t^ (c g10) (d g11))) g12)
  context...:
   /home/asumu/plt/racket-git/racket/collects/unstable/error.rkt:65:0: raise-syntax-error*46
   /home/asumu/plt/racket-git/racket/collects/syntax/parse/private/runtime-report.rkt:31:0
````

My guess is that your contract generation is spitting out malformed `unit/c` syntax. Probably means you need more contract related tests.

File: typed-racket-lib/typed-racket/base-env/prims-contract.rkt [hunk 1]
```diff
@@ -121,6 +124,10 @@
               (~var constructor (opt-constructor legacy #'nm.nm))]
              #:with (constructor-parts ...) #'constructor.value))
 
+  (define-syntax-class signature-clause
+    #:attributes (sig-name (member 1))
```
takikawa: Nitpick: [ ] around `member 1` instead.

typed-racket-lib/typed-racket/base-env/signature-prims.rkt
```diff
@@ -0,0 +1,101 @@
...
+
+(require "../utils/utils.rkt"
+         "colon.rkt"
+         (for-syntax syntax/parse
+                     racket/base
+                     racket/list
+                     racket/syntax
+                     syntax/kerncase
+                     "../private/syntax-properties.rkt"
+                     "../typecheck/internal-forms.rkt"
```
takikawa: Use shorthand `require` macros here since you're already requiring `"../utils/utils.rkt"`. That is, the forms like `(typecheck internal-forms)`. Also repeat in other files.

typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -273,6 +273,8 @@
        (test-phase1 #,(syntax/loc #'lit (LITERAL/FAIL lit))
          (test-literal/fail (quote-syntax lit) msg.v ex.v)))]))
 
+;;check that typechecking a module fails
+
```
takikawa: Missing tests or should these two lines just be deleted?

typed-racket-lib/typed-racket/base-env/signature-prims.rkt
```diff
@@ -0,0 +1,101 @@
...
+                     racket/dict
+                     racket/unit-exptime
+                     (utils tc-utils))
+         (only-in racket/unit
+                  [define-signature untyped-define-signature]
+                  extends)
+         (for-label "colon.rkt")
+         (submod "../typecheck/internal-forms.rkt" forms)
+         (only-in "../../typed/racket/base.rkt" define-type)
+         (for-template (rep type-rep)))
```
takikawa: I'm surprised that you need this `for-template` require. What's it used for?

typed-racket-lib/typed-racket/base-env/signature-prims.rkt
```diff
@@ -0,0 +1,101 @@
...
+  
+  (define-syntax-class sig-var-form
+    #:literal-sets (colon)
+    (pattern [name:id : type]
+             #:with internal-form #'(name type)
+             #:with erased #'name))
+
+  ;; Preliminary support for type definitions in signatures
+  ;; The form is allowed in signature definitions, but currently
+  ;; is ignored
```
takikawa: I don't remember if I commented on this previously, but can we just disable them (i.e., not parse them) until we support them?

typed-racket-lib/typed-racket/base-env/signature-prims.rkt
```diff
@@ -0,0 +1,101 @@
...
+             #:with extends-id #'super
+             #:attr form #'(extends super))
+    (pattern (~seq)
+             #:with internal-form #'#f
+             #:with extends-id '()
+             #:attr form '())))
+
+
+
+(define-for-syntax (process-signature-forms forms)
```
takikawa: Purpose statement, contract.

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+                   [invoke-unit untyped-invoke-unit]
+                   [invoke-unit/infer untyped-invoke-unit/infer]
+                   [compound-unit untyped-compound-unit]
+                   [define-unit untyped-define-unit]
+                   [define-compound-unit untyped-define-compound-unit]
+                   [define-values/invoke-unit untyped-define-values/invoke-unit]
+                   [define-values/invoke-unit/infer untyped-define-values/invoke-unit/infer]
+                   [compound-unit/infer untyped-compound-unit/infer]
+                   [define-compound-unit/infer untyped-define-compound-unit/infer]
+                   [unit-from-context untyped-unit-from-context]
```
takikawa: These would probably be better as `(prefix-in (only-in ...))` and a separate `(only-in ...)` for the rest.

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+             #:with names #'(sig.sig-name ...)
+             #:attr renamers (attribute sig.rename)))
+  
+  (define-syntax-class sig-spec
+    #:literals (prefix rename)
+    (pattern sig-id:id
+             #:attr rename (lambda (id) id)
+             #:with sig-name #'sig-id)
+    (pattern (prefix p:id sig:sig-spec)
+             #:attr rename (lambda (id) (format-id #'sig.sig-name
```
takikawa: Could you put in a comment explaining the purpose of the `rename` attribute? It appears to be doing something complicated.

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+                 (define lu (member id mapping 
+                                    (lambda (x y) 
+                                      (free-identifier=? x (car y)))))
+                 (and lu (cdar lu)))
+
+               (define rn ((attribute sig.rename) id))
+               (or (lookup rn) rn))
+             #:with sig-name #'sig.sig-name))
+  
+  ;; need lexical signature vars to error with duplicate type annotations
```
takikawa: This comment confuses me since AFAICT this function below does not emit duplicate type errors. Can you clarify it? (also in general: contracts and purpose statements for helpers)

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+             #`(begin (#,name e) ...)))
+  (define-splicing-syntax-class (rest-form name arg)
+    (pattern (~seq) 
+             #:with trampoline-form
+             #`(begin))
+    (pattern (~seq e1 e2 ...)
+             #:with trampoline-form
+             #`(begin (#,name #,@arg e1)
+                      (#,name #,@arg e2) ...))))
+
```
takikawa: Can you provide a comment explaining how this abstraction works / is intended to be used?

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+              [(~var b (begin-form #'name #`(arg ...)))
+               #'b.trampoline-form]
+              case ...
+              [_ exp-e])]
+           [(_ arg ... (~var exprs (rest-form #'name #`(arg ...))))
+            #'exprs.trampoline-form]))]))
+
+
+;; Typed macro for define-values/invoke-unit
+;; This has to be handled specially because they types of
```
takikawa: Nitpick: typo

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+;; the defined values must be registered in the environment
+(define-for-syntax (imports/members sig-id)
+  (match-define-values (_ (list imp-mem ...) _ _)
+                       (signature-members sig-id sig-id))
+  #`(#,sig-id #,@(map (lambda (id)
+                        (local-expand
```
takikawa: This looks tricky, can you add a comment explaining?

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+
+(define-syntax (define-values/invoke-unit stx)
+  (syntax-parse stx 
+    #:literals (import export)
+    [(_ unit-expr
+        (import isig:sig-spec ...)
+        (export esig:sig-spec ...))
+     (define imports-stx (syntax->list #'(isig.sig-name ...)))
+     (define exports-stx (syntax->list #'(esig ...)))
+     (with-syntax ([temp (syntax-local-introduce (generate-temporary))])
```
takikawa: Since you already have `racket/syntax` imported, you may want to use `define/with-syntax` here.

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+       #`(begin
+           #,(internal (quasisyntax/loc stx
+                         (define-values/invoke-unit-internal
+                           (#,@(map imports/members imports-stx))
+                           (#,@(process-dv-exports exports-stx)))))
+           (: temp (Unit (import isig.sig-name ...)
+                         (export esig.sig-name ...)
+                         ;; FIXME this needs to be AnyValues
+                         ;; FIXME: init-depends
```
takikawa: Seems like we should implement the `init-depends` support here. Is there anything blocking that? (the other FIXME makes  sense)

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+           
+           #,(ignore (quasisyntax/loc stx
+                       (untyped-define-values/invoke-unit unit-expr
+                                                          (import isig ...)
+                                                          (export esig ...))))))]))
+(begin-for-syntax
+  (define (id-set-diff s1 s2)
+    (filter
+     (lambda (id) (not (member id s2 free-identifier=?)))
+     s1))
```
takikawa: Maybe use `syntax/id-set` instead to get these automatically via `racket/set` operations?

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+  (define-syntax-class unit-spec
+    #:literals (link)
+    (pattern unit-id:id
+             #:attr unit-ids (syntax->list #'(unit-id))
+             #:attr link-units? #f
+             #:with untyped-stx (tr:unit:invoke:expr-property 
+                                 #'(#%expression unit-id) #t)
+             #:with last-id
+             (local-expand #'unit-id (syntax-local-context) (kernel-form-identifier-list))
+             )
```
takikawa: Nitpick: weird indentation

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+             (local-expand #'uid-last (syntax-local-context) (kernel-form-identifier-list)))))
+
+;; Note: This may not correctly handle all use cases of
+;; define-values/invoke-unit/infer
+;; inferred imports and exports are handled in the following way
+;; - the exports of ALL units being linked are added to the export list
+;;   to be registered in tc-toplevel
+;; - imports are the set-difference of the union of all imports and the
+;;   union of all exports
+;; This seems to correctly handle bot recursive and non-recursive
```
takikawa: bot -> both? What doesn't this handle?

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+    (pattern (~seq)
+             #:attr untyped-import #'()
+             #:with imports #'())
+    (pattern (import sig:id ...)
+             #:attr untyped-import #'((import sig ...))
+             #:with imports #'((quote-syntax sig) ...))))
+
+
+;; need to do extra work to make this work with the existing
+;; invoke-unit machinery
```
takikawa: Can you describe the "extra work"?

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+            (void)
+            (void #,@#'imports.imports)
+            (untyped-invoke-unit
+             #,(tr:unit:invoke:expr-property
+                #'unit-expr
+                #t)
+             #,@(attribute imports.untyped-import)))))))]))
+
+;; This table implementation is going to break when only/except are allowed in
+;; typed units, the indexing strategy won't work in that case
```
takikawa: I don't understand this comment. Are you saying with those forms *none* of our approaches will work or some of them will?

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+(define-syntax (unit stx)
+  (syntax-parse stx
+    [(unit imports:unit-imports 
+       exports:unit-exports
+       init-depends:init-depend-form
+       e:unit-expr ...)
+     (ignore
+      (tr:unit
+       (quasisyntax/loc stx
+         (untyped-unit  imports
```
takikawa: Nitpick: indentation is a little weird here.

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+    [cu:compound-unit-form
+     (attribute cu.untyped-stx)]))
+
+(define-trampolining-macro (process-define-compound-unit links sigs im ex infer)
+  [(define-values (name:id ...) rhs)
+   #`(define-values (name ...)
+       #,(ignore
+          (tr:unit:compound-property
+           #`(#%expression
+              (begin
```
takikawa: I think you could just fit these on one line.

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+    #:literals (export)
+    (pattern (export ex:infer-link-export ...)
+             #:with export-links-or-sigs
+             #'(void (quote-syntax ex.link-or-sig-id) ...)))
+  
+  (define-syntax-class compound-infer-links
+    #:literals (link)
+    (pattern (link lnk:infer-linkage-decl ...)
+             ;; #:with untyped-links
+             ;; #'(link lnk.untyped-link-decl ...)
```
takikawa: Can this be deleted?

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+             #:with bound-link-ids #'(lb.link-qs ...)
+             #:with bound-sig-ids #'(lb.sig-qs ...)
+             #:with linkage-stx
+             #`((lb ...)
+                #,(tr:unit:compound:expr-property
+                   #'unit-id
+                   'infer)
+                link-id ...)
+             #:with linkage-information
+             (begin
```
takikawa: This attribute expression looks exactly the same as the one below. Abstract with a helper function.

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+                                     #`(quote-syntax #,id))
+                                   exports))
+                     (void (quote-syntax #,runtime-id))))
+                'infer)))))
+
+;; NOTE/TODO:
+;; - it seems that the docs for compound-unit/infer
+;;   suggest that imports are filled in from the 
+;;   static information bound to the unit-ids
+;;   but simple tests don't seem to confirm this
```
takikawa: What does this imply about how we implement the macro?

typed-racket-lib/typed-racket/base-env/unit-prims.rkt
```diff
@@ -0,0 +1,767 @@
...
+     (ignore
+      (tr:unit:from-context-property
+       (quasisyntax/loc stx
+         (#%expression
+          (begin
+            (void (quote-syntax sig))
+            (untyped-unit-from-context sig))))
+       #t))]))
+
+
```
takikawa: Nitpick: extraneous whitespace at EOF. In general, delete extraneous whitespace.

typed-racket-lib/typed-racket/env/signature-helper.rkt
```diff
@@ -0,0 +1,126 @@
...
+;; checks that the bindings of a signature identifier are consistent with
+;; those listed in a require/typed clause
+(define (check-signature-bindings name vars stx)
+  (match-define-values (_ inferred-vars _ _) (signature-members name name))
+  (define (id-set-diff s1 s2)
+    (filter
+     ;; is free-label-identifier=? the right thing here?
+     ;; free-template-identifier=? also seems to work
+     (lambda (id) (not (member id s2 free-label-identifier=?)))
+     s1))
```
takikawa: Again, using generic set ops seems like the right thing here.

typed-racket-lib/typed-racket/private/parse-type.rkt
```diff
@@ -364,6 +370,48 @@
                                  "given" v)
                     (make-Instance (Un)))
              (make-Instance v)))]
+      [(:Unit^ (:import^ import:id ...)
+               (:export^ export:id ...)
+               (~optional (:init-depend^ init-depend:id ...) 
+                          #:defaults ([(init-depend 1) null]))
+               (~optional result
+                          #:defaults ([result #f])))
+       (define id->sig (lambda (id)
```
takikawa: Use `(define (f ...) ...)` form.

typed-racket-lib/typed-racket/private/parse-type.rkt
```diff
@@ -364,6 +370,48 @@
...
+       (define imports (check-signatures (map id->sig (syntax->list #'(import ...)))))
+       (define exports (check-signatures (map id->sig (syntax->list #'(export ...)))))
+       (define init-depends (map id->sig (syntax->list #'(init-depend ...))))
+       (check-init-depends/imports? init-depends imports)
+       (define res (attribute result))
+       (make-Unit imports
+                  exports
+                  init-depends
+                  (if res (parse-values-type res) (-values (list -Void))))]
+
```
takikawa: Nitpick: delete extra linebreak.

typed-racket-lib/typed-racket/typecheck/check-unit-unit.rkt
```diff
@@ -0,0 +1,736 @@
...
+         (types utils abbrev union subtype resolve generalize signatures)
+         (typecheck check-below internal-forms)
+         (utils tc-utils)
+         (rep type-rep)
+         (for-syntax racket/base racket/unit-exptime)
+         (for-template racket/base
+                       (private unit-literals)
+                       (submod "internal-forms.rkt" forms)))
+
+(import tc-if^ tc-lambda^ tc-app^ tc-let^ tc-expr^)
```
takikawa: Do you actually need all these imports? It seems like `tc-if^` for example isn't used.

typed-racket-lib/typed-racket/typecheck/check-unit-unit.rkt
```diff
@@ -0,0 +1,736 @@
...
+(export check-unit^)
+
+;; id-set-diff : (listof id) (listof id) -> (listof id)
+;; returns the set difference of two lists of identifers
+(define (id-set-diff s1 s2)
+    (filter
+     (lambda (id) (not (member id s2 free-identifier=?)))
+     s1))
```
takikawa: See above about gen sets.

typed-racket-lib/typed-racket/typecheck/check-unit-unit.rkt
```diff
@@ -0,0 +1,736 @@
...
+
+;; Syntax class deifnitions
```
takikawa: Typo.

typed-racket-lib/typed-racket/typecheck/check-unit-unit.rkt
```diff
@@ -0,0 +1,736 @@
...
+             [exprs #`()]
+             [unannotated-exports (map car export-mapping)])
+            ([a/d (in-list ann/defs)])
+    (syntax-parse a/d
+      [a:unit-body-annotation
+       (define name (attribute a.name))
+       ;; TODO:
+       ;; Duplicate annotations from imports
+       ;; are not currently detected due to a bug
+       ;; in tc/letrec-values
```
takikawa: If there's a bug number/URL can you link it?

typed-racket-lib/typed-racket/typecheck/check-unit-unit.rkt
```diff
@@ -0,0 +1,736 @@
...
+    [t:signature-index-table
+     (values (attribute t.imports)
+             (attribute t.exports)
+             (attribute t.init-depends))]))
+
+(define-syntax-class unit-expansion
+  #:literals (let-values letrec-syntaxes+values #%plain-app quote)
+  #:attributes (;imports
+                ;exports
+                ;init-depends
```
takikawa: Delete?

typed-racket-lib/typed-racket/typecheck/check-unit-unit.rkt
```diff
@@ -0,0 +1,736 @@
...
+            import-vector:expr
+            export-vector:expr
+            list-dep:expr
+            unit-body:expr)
+           #:with body-stx #'unit-body))
+
+
+;; Compound Unit syntax classes
+(define-syntax-class compound-unit-expansion
+  #:literals (#%expression #%plain-app void quote-syntax begin)
```
takikawa: Use literal sets instead with kernel-literals (for all of these).

dfeltey:
I'm closing this and will submit a new pull request, the implementation has changed significantly to work better with the new macro expander.

PR #149, opened 2015-06-08
add newtypes

AlexKnauth:
inspired by haskell newtypes
http://www.mail-archive.com/racket-users@googlegroups.com/msg27118.html
https://www.haskell.org/tutorial/moretypes.html

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
@@ -465,6 +465,25 @@ the type:
     (define-type Bar (U Bar False))]
 }
 
+@section{Defining New Subtypes}
+
+@defform[(define-new-subtype name (constructor t))]{
+Defines a new type @racket[name] that is a subtype of @racket[t].
+The @racket[constructor] is defined as a function that takes a value of type
+@racket[t] and produces a value of the new type @racket[name].
+A @racket[define-new-subtype] definition is only allowed at the top level of a
+file or module.
+@ex[(module m typed/racket
+      (provide Radians radians f)
+      (define-new-subtype Radians (radians Real))
+      (: f : [Radians -> Real])
+      (define (f a)
+        (sin a)))
+    (require 'm)
+    (radians 0)
+    (f (radians 0))]
+}
+
 @section{Generating Predicates Automatically}
 
 @defform[(make-predicate t)]{
--- a/typed-racket-lib/typed-racket/rep/type-rep.rkt
+++ b/typed-racket-lib/typed-racket/rep/type-rep.rkt
@@ -579,6 +579,17 @@
   [#:frees (λ (f) (make-invariant (f value)))]
   [#:key 'continuation-mark-key])
 
+;; Distinction
+;; comes from define-new-subtype
+;; nm: a symbol representing the name of the type
+;; id: a symbol created with gensym
+;; ty: a type for the representation, where this will be a subtype of ty
+(def-type Distinction ([nm symbol?] [id symbol?] [ty Type/c])
+  [#:frees (λ (f) (f ty))]
+  [#:intern (list (Rep-seq ty) nm id)]
+  [#:fold-rhs (*Distinction nm id (type-rec-id ty))]
+  [#:key (Type-key ty)])
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 ;; remove-dups: List[Type] -> List[Type]
--- a/typed-racket-lib/typed-racket/base-env/prims-struct.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-struct.rkt
@@ -29,7 +29,8 @@
 (begin-for-syntax
   (lazy-require [syntax/struct (build-struct-names)]))
 
-(provide define-typed-struct -struct define-typed-struct/exec define-type-alias dtsi* dtsi/exec*)
+(provide define-typed-struct -struct define-typed-struct/exec dtsi* dtsi/exec*
+         define-type-alias define-new-subtype)
 
 (define-for-syntax (with-type* expr ty)
   (with-type #`(ann #,expr #,ty)))
@@ -209,3 +210,27 @@
                #'(begin))
          #,(internal (syntax/loc stx
                        (define-type-alias-internal tname type poly-vars))))]))
+
+(define-syntax define-new-subtype
+  (lambda (stx)
+    (unless (memq (syntax-local-context) '(module module-begin))
+      (raise-syntax-error 'define-new-subtype
+                          "can only be used at module top-level"))
+    (syntax-parse stx
+      [(define-new-subtype ty:id (constructor:id rep-ty:expr))
+       #:with gen-id (generate-temporary #'ty)
+       #:with stx-err-fun
+       #'(lambda (stx)
+           (raise-syntax-error
+            'type-check
+            "type name used out of context"
+            stx
+            (and (stx-pair? stx) (stx-car stx))))
+       #`(begin
+           #,(ignore
+              #'(begin
+                  (define-syntax ty stx-err-fun)
+                  (define constructor (lambda (x) x))))
+           #,(internal (syntax/loc stx
+                         (define-new-subtype-internal ty (constructor rep-ty) #:gen-id gen-id))))])))
+
--- a/typed-racket-lib/typed-racket/base-env/prims.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims.rkt
@@ -39,6 +39,7 @@ the typed racket language.
          (all-from-out "case-lambda.rkt")
          (all-from-out (submod "prims-contract.rkt" forms))
          define-type-alias
+         define-new-subtype
          define-typed-struct
          define-typed-struct/exec
          ann inst
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -351,6 +351,8 @@
          (listof/sc (t->sc elem-ty))]
         [(Base: sym cnt _ _)
          (flat/sc #`(flat-named-contract '#,sym (flat-contract-predicate #,cnt)) sym)]
+        [(Distinction: _ _ t) ; from define-new-subtype
+         (t->sc t)]
         [(Refinement: par p?)
          (and/sc (t->sc par) (flat/sc p?))]
         [(Union: elems)
--- a/typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
@@ -21,6 +21,7 @@
   internal
 
   type-alias
+  new-subtype-def
   type-refinement
   typed-struct
   typed-struct/exec
@@ -31,6 +32,7 @@
   typecheck-failure
 
   type-alias?
+  new-subtype-def?
   typed-struct?
   typed-struct/exec?)
 
@@ -50,6 +52,7 @@
   (internal-forms internal-literals
                   require/typed-internal
                   define-type-alias-internal
+                  define-new-subtype-internal
                   define-type-internal
                   define-typed-struct-internal
                   define-typed-struct/exec-internal
@@ -127,6 +130,8 @@
 (define-internal-classes
   [type-alias
     (define-type-alias-internal name type args)]
+  [new-subtype-def
+    (define-new-subtype-internal name (constructor rep-type) #:gen-id gen-id)]
   [type-refinement
     (declare-refinement-internal predicate)]
   [typed-struct
--- a/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
@@ -67,6 +67,17 @@
       [_:type-alias
        (list)]
 
+      ;; define-new-subtype
+      [form:new-subtype-def
+       ;; (define-new-subtype-internal name (constructor rep-type) #:gen-id gen-id)
+       (define name (syntax-e (attribute form.name)))
+       (define sym (syntax-e (attribute form.gen-id)))
+       (define rep-ty (parse-type (attribute form.rep-type)))
+       (define new-ty (-Distinction name sym rep-ty))
+       (register-type (attribute form.constructor) (-> rep-ty new-ty))
+       (register-type-alias (attribute form.name) new-ty)
+       (list)]
+
       ;; declare-refinement
       ;; FIXME - this sucks and should die
       [t:type-refinement
--- a/typed-racket-lib/typed-racket/types/abbrev.rkt
+++ b/typed-racket-lib/typed-racket/types/abbrev.rkt
@@ -286,6 +286,10 @@
 (define-syntax-rule (->opt args ... [opt ...] res)
   (opt-fn (list args ...) (list opt ...) res))
 
+;; from define-new-subtype
+(define (-Distinction name sym ty)
+  (make-Distinction name sym ty))
+
 ;; class utilities
 
 (begin-for-syntax
--- a/typed-racket-lib/typed-racket/types/printer.rkt
+++ b/typed-racket-lib/typed-racket/types/printer.rkt
@@ -517,6 +517,8 @@
      `(Sequenceof ,@(map t->s ts))]
     [(Error:) 'Error]
     [(fld: t a m) `(fld ,(type->sexp t))]
+    [(Distinction: name sym ty) ; from define-new-subtype
+     name]
     [else `(Unknown Type: ,(struct->vector type))]
     ))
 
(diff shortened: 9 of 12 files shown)


takikawa:
This is cool! I think we will want tests and docs if we were to merge this.

What kinds of guarantees did you intend to provide when these newtypes are used at typed-untyped boundaries? Right now you seem to just compile to the contract of the parent type, but this won't preserve the semantics you have in the typed modules when you interact with untyped modules. In particular, you can end up producing values of the new type without using the constructor that you define in the typed module.

AlexKnauth:
Where would be the best place to put the tests? 

Yes, if it crosses from untyped code into typed code, it will get the type without going through the constructor, and that's intentional.  Going by the example with radians on the racket-users list, if the `sin` function is provided with the type `Radians -> Real` then untyped code could call it with a normal real number without going through the "constructor," but I think it would be too restrictive not to allow untyped code to call `sin` with a real number, but in typed code it would be helpful to specify that it needs radians.  Then `sin` could be a drop-in replacement for the existing `sin` but with the extra benefit of typed code checking that it's supposed to be in radians.  The bad part of that for me though, is that `(make-predicate Radians)` will just act like `real?` would.

takikawa:
Re: tests, you could put them in `typecheck-tests.rkt` in the unit tests folder if you can formulate your tests as unit tests (I think you could by putting the `define-newtype` in a `let`? Or do you not support that?). Otherwise you can put them in the `succeed`/`fail` folders as integration tests.

lexi-lambda:
I don't think "newtype" is the right name for this, even if Haskell uses it. That nomenclature is confusing enough. It'd be nice to come up with a more specific name.

AlexKnauth:
I tried using `define-newtype` in a let, but I got this:
```racket
#lang typed/racket
(let ()
  (define-newtype Radians (radians Real))
  (: f : [Radians -> Real])
  (define (f x)
    (sin x))
  (f (radians 6.28)))
```
```
. Type Checker: parse error in type;
 type name `Radians' is unbound in: Radians
```
I'm not sure what to do about it though.

AlexKnauth:
Yes, I don't really think `newtype` is the best name. Can you think of anything better though?

AlexKnauth:
Also it probably should be called something different anyway because in Haskell I don't think `Radians` would be a subtype of `Real`.  In the `Natural` example they had to explicitly add `Natural` to the `Num` type-class afterwards.

takikawa:
@AlexKnauth I think you need to modify `tc-let-unit.rkt` to parse the `define-newtype` expansion and install the definitions. Like how `define-type` works.

AlexKnauth:
@takikawa 
I've tried a few things and I don't really know what to do to make that work.

AlexKnauth:
I am now getting this:
```
tc-let-unit.rkt tc/letrec-values syntax-parse form:newtype
tc-let-unit.rkt tc/letrec-values new-alias1: .#<syntax (define-values () (begin (quo...>
tc-let-unit.rkt tc/letrec-values new-alias2: .#<syntax (define-values () (begin (quo...>
. . Documents/DrRacket/stuff/GitHub/repos/typed-racket/typed-racket-lib/typed-racket/utils/tc-utils.rkt:264:0: Internal Typechecker Error: not define-type-alias:
(define-values () (begin (quote (define-type-alias-internal Real2 Real ())) (#%plain-app values)))
while typechecking:
(letrec-values (((radians) (lambda (x) x)) (() (begin (quote (define-newtype-internal Radians (radians Real) #:gen-id Radians1)) (#%plain-app values))) (() (begin (quote (:-internal f (Radians -> Real))) (#%plain-app values))) ((f) (lambda (x) (#%app sin x)))) (#%app f (#%app radians (quote 6.28))))
originally:
(let () (define-newtype Radians (radians Real)) (: f : (Radians -> Real)) (define (f x) (sin x)) (f (radians 6.28)))
```

stamourv:
Re name, a few suggestions.
* Dimension (probably the most common use case for that feature)
* Unit (as above, but bad idea as the name is already taken)
* Distinction
* something having to do with opaqueness, which is similar

AlexKnauth:
I personally don't really like unit or dimension because it's not that, even if it can be used for that. 
Distinction seems better to me, but `define-distinction` just sounds a bit weird to me.
Would `define-new-subtype` or `define-distinct-subtype` make a bit more sense?
I don't think something related to opaqueness would make sense, because I made the new types subtypes of the representation type, so for example `Radians` would be subtype of `Real`. This means that a value with type `Radians` can be used anywhere that expects a `Real`, which is completely different from opaque types. So I guess it's one-way-transparent, or something?

AlexKnauth:
Now I'm getting this again:
```racket
#lang typed/racket
(let ()
  (define-newtype Radians (radians Real))
  (: f : [Radians -> Real])
  (define (f x)
    (sin x))
  (f (radians 6.28)))
```
```
. Type Checker: type mismatch
  expected: Radians
  given: Real in: (let () (define-newtype Radians (radians Real)) (: f : (Radians -> Real)) (define (f x) (sin x)) (f (radians 6.28)))
```

AlexKnauth:
And this also gives me the same error:
```racket
#lang typed/racket
(let ()
  (define-newtype Radians (radians Real))
  "nothing to do with anything")
```
```
. Type Checker: type mismatch
  expected: Radians
  given: Real in: (let () (define-newtype Radians (radians Real)) 5)
```

AlexKnauth:
I need some help with this.  It seems like the problem is that I'm trying to use basically `(register-type #'radians (-> Real Radians))` in `tc-let-unit.rkt` [here](https://github.com/racket/typed-racket/pull/149/files#diff-27e4b9186e9cd4f6c6de29ab0986b4ccR122) and while that seemed to work in `tc-toplevel.rkt` [here](https://github.com/racket/typed-racket/pull/149/files#diff-fc25ff57bf095926d212a0dcd41563c5R77), it is not working for `define-newtype` as an internal definition. How can I do this?

If I have this expression:
```racket
(let ()
  (define-newtype Radians (radians Real))
  "nothing to do with anything")
```
Then starts to expand to:
```racket
(let-values ()
   (begin
     (begin
       (define-syntax Radians (lambda (stx) (raise-syntax-error ...)))
       (define radians (lambda (x) x)))
     (define-values () ; this gets processed by the type-checker later
       (begin
         '(define-newtype-internal Radians (radians Real) #:gen-id Radians1)
         (#%plain-app values))))
   "nothing to do with anything")
```
And then after some more expansion the type checker gets to the `define-newtype-internal` form and does [this](https://github.com/racket/typed-racket/pull/149/files#diff-27e4b9186e9cd4f6c6de29ab0986b4ccR116), which includes [this](https://github.com/racket/typed-racket/pull/149/files#diff-27e4b9186e9cd4f6c6de29ab0986b4ccR122).  

For `define-newtype` at the module level, this worked fine, and the type-checker trusted that declaration and didn't check to see if the `radians` constructor (which is `(lambda (x) x)`) actually returns a value of the type `Radians`.  

I'm not sure if that's intentional or if that could be a bug in the type-checker, but that's how `define-newtype` currently works at the module level.

But within the `let`, the type-checker doesn't trust that declaration, and actually checks it, and when it does, it finds that `(lambda (x) x)` can receive a `Real`, but then returns a `Real` where it expected `Radians`.

So how can I get this to work within the `let`?

AlexKnauth:
Ok, I finally gave up on trying to make `define-newtype` work as a local definition (struct doesn't work as a local definition either), and it now raises an error message unless it's used at the module level.

AlexKnauth:
Now about the name; would `define-new-subtype` be better? Or?

AlexKnauth:
Ok I renamed it `define-new-subtype`.

AlexKnauth:
I just added documentation.

AlexKnauth:
@takikawa @lexi-lambda @stamourv 
What do you think?

johanatan:
Subtype implies inheritance which I think should not be evoked here. I actually thought 'newtype' was fine.

AlexKnauth:
Well, all operations that work on values with the base type also work on values with the new subtype, which inheritance also gives you, and it even inherits the representation of the base type.
That's closer to inheritance than what subtype already means in typed racket:
http://docs.racket-lang.org/ts-guide/types.html#%28part._.Subtyping%29

johanatan:
I would be more inclined to call that an 'equivalent type' i.e., == rather than <=. This is really just a type alias isn't it?

lexi-lambda:
No, it isn't, `define-type` creates type aliases. I think “subtype” is the right nomenclature here.

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -579,6 +579,17 @@
   [#:frees (λ (f) (make-invariant (f value)))]
   [#:key 'continuation-mark-key])
 
+;; Newtype
+;; comes from define-new-subtype
+;; nm: a symbol representing the name of the Newtype
+;; id: a symbol created with gensym
+;; ty: a type for the representation
+(def-type Newtype ([nm symbol?] [id symbol?] [ty Type/c])
```
lexi-lambda: Since we can change this relatively easily now, would it be a good idea to make the internal name match the external one as closely as possible? Just calling it `Subtype` would likely be equally confusing, though, so I'm not sure what the right name is.

johanatan:
Would it be too much trouble to explain how this 'new-subtype' differs from an alias? i.e., does it allow extending the original type with new methods?

typed-racket-lib/typed-racket/rep/type-rep.rkt
AlexKnauth: I'm not sure either. That's why I left it as `Newtype`.

lexi-lambda:
It's probably best demonstrated by example, but I agree that it's important to make sure it has a good explanation in the docs that's phrased well. The quintessential example, and the one that the current documentation and integration tests use, is distinguishing between radians and degrees. They're both the same type, `Real`, but it would be nice if the type system could catch the mistake of passing degrees to `sin`, which expects radians.

You can make this work by creating a subtype of `Real` called `Radians` with a constructor `radians` of type `Real -> Radians`. You can still use a value of type `Radians` in places where a `Real` is expected, but trying to pass a `Real` into `sin` will complain (if `sin` is retyped as `Radians -> Real`).

johanatan:
F# calls these'Units' (and it does also have the notion of 'Unit' representing void) but subtype makes sense now.

Can this mechanism also be used to bring new opaque types into existence at the API boundaries with untyped code; i.e., an external untyped API gives me a token, moniker, handle or the like and expects me to pass it back to it at some point. Would be nice to see a type emerge at that boundary and propagate through the call stack and back to the exit point to 3rd party/ external code.

AlexKnauth:
Re: F# calls these 'Units':
Since multiplying with these doesn't work like it does with units, it wouldn't be a good idea to call them units, and racket already has a different thing called units. It might be possible to use this new feature to define units, but doing that properly would require redesigning the type system a lot more to allow, for instance, `*`, `expt`, etc. to multiply the units properly, and for `+` (and similar functions) to raise a type error when given numbers with different units. 

Re: Can this mechanism also be used ...
What do you mean?

lexi-lambda:
@johanatan Doesn't the concept of opaque types already exist through `require/typed`'s `#:opaque` clause? Or are you referring to something different?

AlexKnauth:
Also `define-new-subtype` can be used for other things not necessarily related to units or even numbers. The first idea that came into my head was to use this for certain vector types, so that ... well, it might be easier to explain with this: 
http://pkg-build.racket-lang.org/doc/vector-struct/index.html
Using this `vecstruct` with `define-new-subtype` would help you make sure that you never confuse a normal vector with one of these.

Another idea that just popped into my head was to use this for keeping track of things that are supposed to be immutable, or come to think of it, a better way of doing a few of the things you could do refinement types (if they worked).

johanatan:
@AlexKnauth-- I explained it precisely the first time. Is there some aspect of that explanation that you could ask specific questions on? It would be easier for me to help you understand as I have no other words to explain it any differently than I did the first time.

johanatan:
@lexi-lambda: Possibly. I will have to read up on those. Do you think they fit my description? i.e., do you understand what I'm asking for?

lexi-lambda:
I think so, and I think it's likely that's the sort of thing you're looking for.

johanatan:
@lexi-lambda: can I be used to bring a new type into existence at the untyped/typed boundary? Reading through your notes on opaque types and there's a paragraph that expressed doubts about how this will work with untyped boundaries.

lexi-lambda:
You'll want to look at [the documentation for `require/typed`](http://docs.racket-lang.org/ts-reference/special-forms.html#%28form._%28%28lib._typed-racket%2Fbase-env%2Fprims..rkt%29._require%2Ftyped%29%29), which explains how `#:opaque` works. It's specifically designed to handle that sort of thing across typed/untyped boundaries, and it works quite well.

johanatan:
@lexi-lambda: Perfect. Thank you.

johanatan:
Yea, that was one of my open questions about this actually. Thx for clarifying.

rfindler:
Did you try consulting the author of newtypes in Haskell to see if they had
a name they'd wished they used?

AlexKnauth:
Um, no, I didn't ... I don't know who did, I don't know how to contact them, and I don't know if it might be bothering them too much ...
Also, in Haskell the new type wouldn't be a subtype of the base type, right? So this isn't really the same thing.

plt:
I think `define-new-subtype` sounds like the right name -- it isn't really `newtype` from Haskell since there's a relationship with the old type. 

Requiring use at the top-level seems fine too.

typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
```diff
@@ -465,6 +465,23 @@ the type:
     (define-type Bar (U Bar False))]
 }
 
+@section{Defining New Subtypes}
```
plt: The docs should mention the restriction to module top level.

typed-racket-lib/typed-racket/rep/type-rep.rkt
plt: I think a different name than `Newtype` is needed here.

samth:
Sorry, didn't mean to comment as @plt. This looks good to me.

typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
AlexKnauth: Ok done.

typed-racket-lib/typed-racket/rep/type-rep.rkt
AlexKnauth: What though? I can't exactly call it `Subtype` or `New-Subtype` to match `define-new-subtype`, so I'm not sure what to do with this.

samth: Why not `New-Subtype`?

AlexKnauth: My thinking is that since "subtype" means something outside of what a `Newtype` (or whatever you call it) is, we shouldn't call it `New-Subtype` because otherwise it would imply that new types have to be `New-Subtype` types to be subtypes of other types.

rfindler: I don't mean to be a pita but why is "new" a useful word here? Isn't ever
definition "new" in this sense?

lexi-lambda: I've considered that point, too, and I'm ambivalent about it. I feel like explicitly including the word "new" in `define-new-subtype` adds a little bit of semantic information wrt intent, but on the other hand, it does feel redundant.

AlexKnauth: Well, new as in it's not just a type alias to the old type, that it's a different type, distinct from the old one (I'm guessing that's why stamourv suggested distinction as another possible name for this).

johanatan: This is exactly my objection to the word 'subtype' in the externally visible name.

samth: @johanatan It is, precisely, a subtype, so that part of the name make sense.

@rfindler It's "new" in the sense of "fresh" -- different definitions produce unrelated types, unlike `define-type`, for example.

rfindler: Should we say "new-cons" and "new-string" and "new-let" (since those are in
new scopes)?

>

AlexKnauth: Only if `(string=? (new-cons a b) (new-cons a b))`, `(string=? (new-string a ...) (new-string a ...))`, and `(equal? (new-let ([x a]) b) (new-let ([x a]) b))` produce false. But if according to a `type=?` predicate  the `a` and `b` types are equal if `a` is a subtype of `b` and `b` is a subtype of `a` (see [this](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/types/subtype.rkt#L697)), then since two types generated by different `define-new-subtype` declarations with the same base type, then those two types won't be the same type, and they won't even be related by sub-typing, so they are "new" (as in "fresh" or "distinct").

AlexKnauth: I've been thinking about this more, and even though it isn't related to the name `define-new-subtype`, would `Distinction` be a good name for the internal `type-rep.rkt` type? (idea from @stamourv https://github.com/racket/typed-racket/pull/149#issuecomment-110394335)

AlexKnauth:
Does this look good to merge?

PR #153, opened 2015-06-27
Add typed/file/md5, typed/web-server/http.rkt, typed/web-server/confi…

wargrey:
…guration/responders.rkt

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/libraries.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/libraries.scrbl
@@ -68,6 +68,8 @@ The following libraries are included with Typed Racket in the
 @defmodule/incl[typed/net/smtp]
 @defmodule/incl[typed/net/uri-codec]
 @defmodule/incl[typed/net/url]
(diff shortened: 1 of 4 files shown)


lexi-lambda:
1. `typed/file/md5` is already provided by `typed-racket-lib`, so adding it to `typed-racket-more` is an error.
2. Using `{`braces`}` is a little unorthodox in traditional Racket style. I don't have any opposition to them, myself, but keeping things normalized in the TR repo might be worthwhile, though it's probably not a big problem in `typed-racket-more`. Do you have an opinion, @samth?

wargrey:
Sorry, I should tell the details.
In my point of view, `{}` indicates that this is a `Data Expression` and will not evaluated as normal `Racket Expression`. So the only purpose of `{}` is making it easier to parse by eyes. Nonetheless, I do not insist on this style.

samth:
I think we should follow the [style guide](http://docs.racket-lang.org/style/), which doesn't recommend using `{}`.

samth:
Thanks for the new wrappers!

PR #154, opened 2015-06-28
make append preserve non-emptyness of first argument

AlexKnauth:
see https://github.com/racket/scribble/pull/8#discussion_r33423152


samth:
Can you add a unit test for this? Otherwise, looks good.

AlexKnauth:
Yes, and I also fixed some other tests because append now returns a more specific type.

samth:
Feel free to merge this now, @AlexKnauth

PR #156, opened 2015-06-30
Fix missing match cases when trying to bind AnyValues typed "results"

btlachance:
I'm not 100% sure this is strictly related to AnyValues but that's what initially triggered this bug. I'm happy to hear of a way to word this better, if it matters.


btlachance:
I'm not 100% sure this is strictly related to AnyValues but that's what initially triggered this bug. I'm happy to hear of a way to word this better, if it matters.

samth:
You probably need an integration test to actually cover the tc-toplevel code.

btlachance:
Thanks for catching that! PR updated

takikawa:
Merged.

PR #158, opened 2015-07-20
Remove dependency on `unstable-list-lib`

stamourv:
Depends on Racket PR #972, which moves most of its contents to core libraries.

Moves a few `unstable/list` functions to TR itself.

### DIFF
--- a/typed-racket-lib/info.rkt
+++ b/typed-racket-lib/info.rkt
@@ -5,7 +5,6 @@
 (define deps '(("base" #:version "6.2.0.2")
                "pconvert-lib"
                "unstable-contract-lib"
-               "unstable-list-lib"
                "source-syntax"
                "compatibility-lib" ;; to assign types
                "string-constants-lib"))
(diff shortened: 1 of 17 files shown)


typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -27,6 +27,10 @@
 
 (define-logger online-check-syntax)
 
+(define (filter-multiple l . fs)
```
samth: I think this belongs in `utils.rkt`.

samth:
Otherwise, looks good.

typed-racket-lib/info.rkt
```diff
@@ -5,7 +5,6 @@
 (define deps '(("base" #:version "6.2.0.2")
```
samth: Needs a dependency on a new base version.

stamourv:
Merged, with comments.

PR #160, opened 2015-07-23
Cancel marks for 'disappeared-use ids in colon macros

btlachance:
Fixes check-syntax arrows for the id and optional second :


AlexKnauth:
This looks good to me, but is there anyone with any objection to merging this?

samth:
@btlachance Can you give an example program that this fixes?

AlexKnauth:
I tried it with this program:
```racket
#lang typed/racket
(: x : Real)
(define x 3)
```

samth:
Ok, looks good to me.

btlachance:
D'oh, that makes sense to have an example.. Thanks for providing one, Alex

PR #161, opened 2015-07-27
Merge `print-syntax?` and `show-input?` parameters into just the `print-syntax?` parameter

dfeltey:
It seems that both of these parameters had the intent of printing expanded syntax before typechecking. This pull request merges the behavior into the `print-syntax?` parameter, allowing expanded syntax to be printed based on a single flag in tc-setup.rkt


samth:
Looks good to me.

PR #162, opened 2015-07-27
Prevent -let-internal name from leaking to users

btlachance:
Because -let-internal wasn't rename-out'd, users would see its name in
syntax error messages. Adding #:context forces the error messages to be
phrased in terms of the given form.


samth:
LGTM.

takikawa:
Merged.

PR #170, opened 2015-08-02
fix issue #169

### LINKED ISSUE #169: bug with define-new-subtype and for*/list (and inference)
In this program, `all-poss` is given the type `(Listof Nothing)` when it should have the type `(Listof Pos*)`:
```racket
#lang typed/racket
(define-type Pos Integer)
(define-new-subtype Pos* (p Pos)) ; using Integer instead of Pos here fixes it
(define all-poss ;: (Listof Pos*) ; this annotation fixes it
  (for*/list ([x (in-range 3)]) : (Listof Pos*) ; using for/list instead of for*/list also fixes it
    (p x)))
(ann all-poss (Listof Nothing)) ; this should fail, but works
```

AlexKnauth:
fixes #169

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-types-extra.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-types-extra.rkt
@@ -17,7 +17,7 @@
 (define-other-types
   -> ->* case-> U Rec All Opaque Vector
   Parameterof List List* Class Object Values Instance Refinement
-  pred Struct Struct-Type Prefab Top Bot)
+  pred Struct Struct-Type Prefab Top Bot Distinction)
 
 (provide (rename-out [All ∀]
                      [U Un]
--- a/typed-racket-lib/typed-racket/types/subtype.rkt
+++ b/typed-racket-lib/typed-racket/types/subtype.rkt
@@ -293,6 +293,9 @@
          [((or (? Struct? s1) (NameStruct: s1)) (Value: (? (negate struct?) _)))
           #f]
          ;; from define-new-subtype
+         [((Distinction: nm1 id1 t1) (app resolve (Distinction: nm2 id2 t2)))
+          #:when (and (equal? nm1 nm2) (equal? id1 id2))
+          (subtype* A0 t1 t2)]
          [((Distinction: _ _ t1) t2)
           (subtype* A0 t1 t2)]
          ;; sequences are covariant
--- a/typed-racket-test/fail/issue-169-2.rkt
+++ b/typed-racket-test/fail/issue-169-2.rkt
@@ -1,0 +1,8 @@
+#lang typed/racket
+(define-type Pos Integer)
+(define-new-subtype Pos* (p Pos))
+(define lst : (Listof Pos*)
+  '())
+(define lst*
+  (reverse lst))
+(ann lst* (Listof Nothing)) ; this should fail
--- a/typed-racket-test/fail/issue-169-2.rkt
+++ b/typed-racket-test/fail/issue-169-2.rkt
@@ -1,3 +1,5 @@
+#;
+(exn-pred #rx"expected: \\(Listof Nothing\\).*given: \\(Listof Pos\\*\\)")
 #lang typed/racket
 (define-type Pos Integer)
 (define-new-subtype Pos* (p Pos))
--- a/typed-racket-lib/typed-racket/base-env/prims-struct.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-struct.rkt
@@ -206,10 +206,9 @@
             stx
             (and (stx-pair? stx) (stx-car stx))))
        #`(begin
+           (define-type-alias ty (Distinction ty gen-id rep-ty))
            #,(ignore
-              #'(begin
-                  (define-syntax ty stx-err-fun)
-                  (define constructor (lambda (x) x))))
+              #'(define constructor (lambda (x) x)))
            #,(internal (syntax/loc stx
                          (define-new-subtype-internal ty (constructor rep-ty) #:gen-id gen-id))))])))
 
--- a/typed-racket-lib/typed-racket/private/parse-type.rkt
+++ b/typed-racket-lib/typed-racket/private/parse-type.rkt
@@ -100,6 +100,7 @@
 (define-literal-syntax-class #:for-label values)
 (define-literal-syntax-class #:for-label Top)
 (define-literal-syntax-class #:for-label Bot)
+(define-literal-syntax-class #:for-label Distinction)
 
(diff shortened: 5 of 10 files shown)


typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -266,7 +266,7 @@
          ;; [(_ (Univ:)) A0]
          ;; error is top and bot
```
andmkent: So, does this mean this comment here is no longer true?

AlexKnauth: I think so. I think it means that it is only top now. I'm not sure why it would be bot in the first place, but changing this seems to break more tests, so I'm not sure how this should work.

samth:
This is not the right change. `Error` is intended to be allowable anywhere, and only generated when there is a type error to prevent additional downstream type errors.  If in #169 there's an `Error` generated, then we should figure out why there's no type error resulting from that.

AlexKnauth:
Ok. I just hadn't expected the subtype check here (https://github.com/AlexKnauth/typed-racket/blob/fix/typed-racket-lib/typed-racket/infer/infer-unit.rkt#L483) to succeed, and that was what caused #169 .  

I'll look for another way to solve this then.

AlexKnauth:
@samth Is the `parse-type` here (https://github.com/racket/typed-racket/blob/0c1d58a07b4276575df283b5d6ac87e99c8832db/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt#L570) the right way to do this? 

Could that be what's causing `Error` to appear here for user-defined types? Is there a better way of doing this that avoids that?

AlexKnauth:
Re: `parse-type`, maybe not. I put a printf in there, and `parse-type` is returning the right thing (`Pos` not `Error`). So I have no idea where the `Error` type is coming from.

AlexKnauth:
Nope. I was wrong. `parse-type` is returning `Error`, but I got thrown off because it was printing as `Pos`, even though it matches the pattern `(Error:)`.  So, is there something else I should be using instead of `parse-type`?

samth:
The printing issue is almost certainly because we end up with `Pos` aliased to `Error`.

Do you know why `parse-type` is failing?

AlexKnauth:
No I don't.

samth:
I think that's the first thing to investigate.

AlexKnauth:
It seems like the Error type (still printed as Pos) is in the mapping table before I even call `parse-type`, and that's where it's coming from. But how did that get there?

AlexKnauth:
Oh. It gets filled with all error types, and then replaced with the correct types later.

AlexKnauth:
Ok I took a completely different approach and it's working now.

typed-racket-test/fail/issue-169-2.rkt
```diff
@@ -0,0 +1,8 @@
+#lang typed/racket
```
samth: Add an error message specification here; see the other files for examples.

typed-racket-test/fail/issue-169-2.rkt
```diff
@@ -0,0 +1,10 @@
+#;
+(exn-pred #rx"expected: (Listof Nothing).*given: (Listof Pos*)")
```
AlexKnauth: Like this?

samth: Yeah.

AlexKnauth: Oh, except the parens and `*` probably aren't good in the regexp there.

AlexKnauth:
How does this look now?

typed-racket-lib/typed-racket/base-env/base-types-extra.rkt
```diff
@@ -17,7 +17,7 @@
 (define-other-types
   -> ->* case-> U Rec All Opaque Vector
   Parameterof List List* Class Object Values Instance Refinement
-  pred Struct Struct-Type Prefab Top Bot)
+  pred Struct Struct-Type Prefab Top Bot Distinction)
```
samth: Why is this now exposed to users? What does it correspond to?

AlexKnauth: It's not; that's what this except-in is for: https://github.com/racket/typed-racket/pull/170/files#diff-be2db6e0db44e2e94aa8b096cd9b3850R21

samth: Ok, now I think I understand. Why not create a new internal form, instead of a new type constructor that isn't exposed?

AlexKnauth: I couldn't think of any other way to allow both new-subtypes to refer to type aliases and for other forms to refer to new-subtypes properly. 

But actually, this still is broken, in this example:
```racket
#lang typed/racket/base
(define-new-subtype Lst (make-Lst (U Null (Pairof Elt Lst))))
(define-new-subtype Elt (make-Elt (U Symbol Lst)))
(: cns : Elt Lst -> (Pairof Elt Lst))
(define cns
  (inst cons Elt Lst))
(: lst : Elt * -> Lst)
(define (lst . args)
  (for/fold ([lst (make-Lst '())]) ([elt (in-list (reverse args))])
    (make-Lst (cns elt lst))))
```
With this error highlighting the `lst` in the `for/fold` accumulator:
```
. Type Checker: type mismatch
  expected: Lst
  given: Lst in: lst
```

samth: I guess I'm confused why the work done by this `define-type-alias` form can't be done manually in `define-new-subtype/pass1`.

samth: Also, I am not sure that we even want to support recursive uses of `new-subtype` like this.

AlexKnauth: Oh it might be because I never allowed Distinction types to be subtypes of themselves.

AlexKnauth: Nope, still broken, even with that.

samth: Probably @takikawa is the right person to help with recursive aliases.

AlexKnauth: Ok I just fixed it. There was a missing call to resolve in there, and that fixed it. I also added a version of that example as another test.

AlexKnauth:
The travis build is passing now, thanks to #175

PR #175, opened 2015-08-13
Reimplement toplevel typechecking with trampoline macros

takikawa:
This pull request reimplements top-level typechecking. This new approach fixes the PR13747 regression due to the new expander and lets us eliminate hacks used to support the top-level. It is an alternative to PR #172.

One slight backwards incompatibility is that I changed `:print-type` and related functions to only work on expressions and not definitions or other top-level only constructs such as `require/typed`. I doubt people ever use `:print-type` on anything but expressions, so this seems ok.

I'll squash most of these commits before pushing.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/top-interaction.rkt
+++ b/typed-racket-lib/typed-racket/base-env/top-interaction.rkt
@@ -46,10 +46,13 @@
     (typecheck tc-app-helper typechecker)
     (rep type-rep)
     (utils tc-utils)
+    (for-syntax racket/base syntax/parse)
     (for-template racket/base))
   (provide
     :type-impl :print-type-impl :query-type/args-impl :query-type/result-impl)
 
+  ;; this one doesn't quite fit the pattern of the next three REPL operations, so
+  ;; this one isn't defined with a macro as below
   (define (:type-impl stx)
     (syntax-parse stx
       [(_ (~optional (~and #:verbose verbose-kw)) ty:expr)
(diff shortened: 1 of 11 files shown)


typed-racket-lib/typed-racket/base-env/top-interaction.rkt
```diff
@@ -77,16 +77,16 @@
...
-                    "This form has no type (it does not produce a value)."
-                    (pretty-format-type
-                     (match type
-                       [(tc-result1: t f o) t]
-                       [(tc-results: t) (-values t)]
-                       [(tc-any-results: f) (-AnyValues f)]))))))]
+       (define type
+         (tc-expr (local-expand #'e 'expression (list #'module*))))
+       #`(displayln
+          #,(if (eq? type 'no-type)
```
samth: Does this happen anymore given your changes?

stamourv:
I'd consider `:print-type` on non-expressions to be a bug, so that's fine by me.

typed-racket-lib/typed-racket/base-env/top-interaction.rkt
```diff
@@ -114,17 +114,17 @@
     (syntax-parse stx
```
samth: I feel like these three forms share a lot of code ...

typed-racket-lib/typed-racket/base-env/top-interaction.rkt
takikawa: Oh good catch, no I don't think it does.

takikawa:
Ok, I've addressed the code feedback so far.

I did abstract the `#%top-interaction` forms, but I'm not sure it actually makes the code easier to read. I pushed the commit anyway, but what do you think?

PR #177, opened 2015-08-15
fix inference with define-new-subtype

AlexKnauth:


AlexKnauth:
On my computer, I ran all the tests the travis build normally runs, and they all passed.

PR #179, opened 2015-08-17
print more Error types as Error types

AlexKnauth:
A change like this probably would have saved me a lot of trouble debugging.

### DIFF
--- a/typed-racket-lib/typed-racket/types/printer.rkt
+++ b/typed-racket-lib/typed-racket/types/printer.rkt
@@ -378,7 +378,10 @@
       [(Value: '()) null]))
   (match type
     ;; if we know how it was written, print that
-    [(? Rep-stx a) (syntax->datum (Rep-stx a))]
+    [(? Rep-stx a)
+     (if (Error? a)
+         `(Error ,(syntax->datum (Rep-stx a)))
+         (syntax->datum (Rep-stx a)))]
     [(Univ:) 'Any]
     ;; struct names are just printed as the original syntax
     [(Name/struct: id) (syntax-e id)]


stamourv:
That sounds like a good idea. Do you have an example whose behavior would change?

typed-racket-lib/typed-racket/types/printer.rkt
```diff
@@ -378,7 +378,10 @@
       [(Value: '()) null]))
   (match type
     ;; if we know how it was written, print that
-    [(? Rep-stx a) (syntax->datum (Rep-stx a))]
+    [(? Rep-stx a)
+     (if (Error? a)
+         `(Error ,(syntax->datum (Rep-stx a)))
+         (syntax->datum (Rep-stx a)))]
```
AlexKnauth: Does this make it clearer?

AlexKnauth:
Re: example: No, I don't have an example that this would change behavior for. This would have changed the printing of the types of constructors for `define-new-subtype` before https://github.com/racket/typed-racket/commit/f8cc9e8dcdf22bb5b8a1ca4dacd774132e094210, but since that's fixed now, I don't have an example that would change.

typed-racket-lib/typed-racket/types/printer.rkt
stamourv: The original code was fine, I'd just have move it a line earlier. Just a nitpick anyway.

stamourv:
I guess what I'm asking is: in which circumstances would such types arise?

AlexKnauth:
One way they appear that I know of: in env/type-alias-helper.rkt the function get-type-alias-info sets the given type-aliases all to Error, to be replaced later. But then if parse-type is called on one of those types, then it will return an Error type with the given identifier as the Rep-stx I think.

stamourv:
Ok, LGTM then.

PR #184, opened 2015-08-20
Fix require/typed and similar macros in some module* or #lang contexts

takikawa:
This closes several bugs related to the typed context flag. In addition to the issues mentioned in the commit message, this should also fix a problem with `with-typed` and requires from other typed modules. (I'll add a test for that later)

This PR relies on a PR that's pending for Racket itself: https://github.com/plt/racket/pull/1020


takikawa:
Merged as 241f04bcdb82d46af73890ccca484df3edbf27fa

PR #185, opened 2015-08-20
fixed unbox-let optimization srcloc propigation

florence:
The let unboxing optimization would drop the location of a syntax object it reassembled, because `this-syntax` was bound to something unexpected.


stamourv:
Merged, thanks!

PR #186, opened 2015-08-20
add measures

AlexKnauth:


samth:
@AlexKnauth, did you think about trying to encode this with the existing type system?

AlexKnauth:
I thought about it, but if I understand correctly (which I probably don't) it would require path-elements for measure-types which would have to point to the units, (I don't know if that's possible, are those supposed to point to objects?), and then after that it seems like a place for dependent types, so I don't know.

AlexKnauth:
Ok, I can see now that putting these unit-arith rules in the types might work for `m+`, `m*`, and similar functions, so I'll try that.

AlexKnauth:
Ok I just realized I approached unit inference the wrong way.  Fixing that ...

AlexKnauth:
Before I could stat to fix that, I ran into another problem. It seems like in code like:
```racket
(: mm/ : (All (u5 u6)
            (-> (Measure Real u5)
                (Measure Real u6)
                (Measure Real (u* u5 (u^ u6 -1))))))
(define (mm/ x t)
  (measure (ann ((ann / [Real Real -> Real]) (ann x Real) (ann t Real)) Real) (u* u5 (u^ u6 -1))))
```
It seems like the `u5` and `u6` in the body of `mm/` point to F tvars with uninterned symbols inside them, and that's messing it up because the units can't be equal when one has interned symbols and the other has uninterned symbols.
Because I'm getting:
```
warning: tvar-env.rkt: uninterned fresh-var symbol: 'u5
warning: tvar-env.rkt: uninterned fresh-var symbol: 'u6
warning: parse-measure-unit.rkt: uninterned symbol: 'u5, from: 
.#<syntax:/Users/Alex/Documents/GitHub/typed-racket/typed-racket-test/succeed/measures.rkt:16:82 u5>
warning: parse-measure-unit.rkt: uninterned symbol: 'u6, from: 
.#<syntax:/Users/Alex/Documents/GitHub/typed-racket/typed-racket-test/succeed/measures.rkt:16:89 u6>
subtype.rkt Distinction: 299
measure-unit-rep.rkt measure-unit=? measure-unit/F:
  a-deps: (set 'u5 'u6)
  b-deps: (set 'u5 'u6)
  a: (u* (u^ u6 -1) (u^ u5 1))
  b: (u* (u^ u6 -1) (u^ u5 1))
  a-deps /= b-deps
  a-deps isn't made of interned symbols!
subtype.rkt Distinction: 301
. Type Checker: type mismatch
  expected: (Measure Real (u* (u^ u6 -1) (u^ u5 1)))
  given: (Measure Real (u* (u^ u6 -1) (u^ u5 1))) in: (ann ((ann / (Real Real -> Real)) (ann x Real) (ann t Real)) Real)
```

AlexKnauth:
Ok I think I figured it out, but I think I'll close this and redo it. I thought that Distinction types would be a good reusable base for measures, and it would be, if not for unit polymorphism. So I'm closing this, and I'll open a new pull request that does this more properly keeping that in mind.

PR #187, opened 2015-08-21
fixed src loc propigation for n-ary*->binary

florence:
Some of the arity changes for the math operations did not properly propagate syntax source locations.


stamourv:
Merged, thanks!

PR #191, opened 2015-08-29
Fix typo in generated contract for Nonpositive-Integer

lexi-lambda:
In the contract for `Nonpositive-Integer`, `nonpositive?` was misspelled as `nonpostive?`, creating an unbound identifier error when a contract needed to be generated.


takikawa:
Thanks!

PR #194, opened 2015-09-03
Allow AnyValues as a return type in user code

btlachance:
Two specific areas I'm unclear on and would appreciate comments about:

1) how filters should interact with AnyValues -- the representation does have a place to store one and I'm just using AnyValues with -top

2) the change in parse-tc-results (at the end of parse-type.rkt), mainly because I'm not clear on how parse-tc-results is (supposed to be) used.


stamourv:
Merged, thanks!

PR #196, opened 2015-09-08
Add types for writeln and println

lexi-lambda:
When I added these to Racket, I never added them to TR's base environment.


samth:
Looks good.

takikawa:
Merged as 4bf3479776aaca8aacaa7382648d7cf510d2020f.

PR #197, opened 2015-09-11
Add docs for typed units

dfeltey:


takikawa:
I noticed some of the lines are fairly long (pretty sure longer than 80 and maybe longer than whatever the style guide says) so can you word wrap these?

stamourv:
Merged, thanks!

PR #198, opened 2015-09-15
Simplified bad syntax error reporting for let (changes c8f02eb by samth)

jsmaniac:
Changed the c8f02eb commit by samth, which fixed srcloc for syntax errors in `let`, to achieve the same thing but with slightly more readable code. Following our discussion on IRC last week:

```
<georges-duperon> samth: thanks for the let srcloc fix.
<georges-duperon> samth: Just curious, wouldn't #:context (datum->syntax stx `(,form) stx stx) have been enough, instead of having a custom raise-syntax-error as well?
<samth> georges-duperon: try it out, if it works submit a PR
```

It seems to work on my copy of racket.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims.rkt
@@ -203,7 +203,7 @@ the typed racket language.
   (let ([mk (lambda (form)
               (lambda (stx)
                 (syntax-parse stx
-                  #:context (datum->syntax stx `(,form) stx stx)
+                  #:context (list (syntax-e form) stx)
                   [(_ (bs:optionally-annotated-binding ...) . body)
                    (quasisyntax/loc stx (#,form (bs.binding ...) . body))])))])
     (values (mk #'let) (mk #'let*) (mk #'letrec))))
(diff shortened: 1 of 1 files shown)


samth:
This looks good, but @rmculpepper changed `syntax-parse` so that now we can do this even more easily: https://github.com/plt/racket/commit/8d9eb0534728888c286e7ae77ce82149bd1ef482

jsmaniac:
@samth I tried changing it to `#:context (list (syntax-e form) stx)`, but interestingly, the error message for `(let)` becomes `let: bad syntax in: (-let-internal)`, whereas when using `#:context (datum->syntax stx `(,form) stx stx)`, I get `let: bad syntax in: (let)`, as expected.

Other cases like `(let ())` or `(let x)` do not have this issue.

samth:
Ok, that's something for @rmculpepper to (maybe) improve about `syntax/parse`, but this is a good change for TR now.

rmculpepper:
I'm not clear on what else `syntax-parse` should be doing in this case.

samth:
I would like to write `#:context <something>` and get the effect of 

```
#:context (datum->syntax stx `(,form) stx stx)
```

samth:
Merged, finally.

PR #201, opened 2015-09-20
add measures

AlexKnauth:


AlexKnauth:
Oh, the travis build is failing because https://github.com/racket/pconvert-lib/pull/4 hasn't been merged.

samth:
A few initial thoughts:

1. I think this is something good to have for TR in general.
2. I'm surprised that this has to be built in as deep as this is. Particularly, the need to have a new kind of base data I think is a bad thing, but in general this seems to touch a lot of parts of the system. I would have expected units of measure to be less invasive.
3. Where would an entirely-library version of unit of measure be weaker than this? I assume there are some places, since this is quite comprehensive. Can we make the change to TR be smaller by pushing more into libraries?

I'll look more in detail at the code soon.

AlexKnauth:
Originally I thought it would be easy to build it on top of `define-new-subtype` and Distinction types, but I was running into all sorts of problems with polymorphic units, so I switched to a new data type that would allow polymorphism and substitution handled by fold-rhs, not by special cases everywhere.

What do you mean by entirely-library version? Do you mean something like
https://github.com/AlexKnauth/measures-with-dimensions
http://pkg-build.racket-lang.org/doc/measures-with-dimensions/index.html
Or do you mean something else?

samth:
Yes, that's exactly what I mean. Can you give a comparison between using that library and this PR?

AlexKnauth:
Well, for one thing, with that library all measures and units are structs handled at runtime, whereas with this pull request they are numbers at run time with the units handled with type-checking. 

Something like `(m+ Time Distance)` type-checks as being `(U Time Distance)` with the library instead of giving a type error. 

With the library it's also kind of impossible to get functions (or macros) like `m*` and `m/` working properly, because as far as I know I just can't express the rules of unit multiplication and division with the existing type system. Something like `(m/ Distance Time)` has to yield an "AnyMeasure" type, because the type-checker can't figure out that a distance divided by a time is a velocity, and I can't add something like that in a library (okay I could with a special case for every pair of units, but there are infinitely many possible units, so ...). 

Also, with this pull request `(Measure Real (u*))` can be a subtype of `Real`, and can be used wherever a Real can be used.

AlexKnauth:
All tests are now passing. What do you think?

andmkent:
I think supporting measures is something that should be done by further expanding the expressive power of paths in TR. By adding (1) some notion of user-definable paths/path-elements and (2) type-level paths, we would allow type-level measures---and many additional desirable features---to be implementable as just a standard library.

AlexKnauth:
Even with that, and type-level functions, it wouldn't allow measure types to be represented by normal numbers at runtime unless it were built in to the language (as far as I know). I don't think I really understand path-elements yet though, so I don't know.

andmkent:
I am in agreement that the approach should allow measure types to be represented by normal runtime numbers. What I am suggesting is that, in the same way TR reasons compositionally about the types of the _run-time fields_ of things, such as the ``` car ``` of pair, TR could reason about the _type-level fields_ of things, such as the ``` measure ``` field of a number (which I believe with some basic type-level functions should support measures). 

(In the implementation the _fields_ I talked about are called _PathElements_)

PR #204, opened 2015-09-26
fix a bug where cover+tests could call `syntax-local-introduce` when not transforming

florence:
This is kind of a bad fix for florence/cover#83. But I still have no clue *how* the error mentioned there actually happens.


samth:
LGTM

samth:
Is this still needed?

florence:
I haven't run cover on TR in a while so IDK that make this necessary still occurs. Let me rerun the tests and figure out (it takes quite a while).

florence:
Yes, this is still needed.

florence:
I'm not sure I understand why the travis build for this failed, and it's not failing locally, so I'm not sure how to proceed on this one.

Issue #205, opened 2015-09-30
Misprediction of predicate on substruct


samth:
```
#lang typed/racket #:no-optimize

(struct (A) foo ([x : A]))

(struct (A) baz foo ())

(define (f [i : Integer]) : (foo Integer)
  (baz i))

(if (baz? (f 1)) 1 2)
```

Behaves differently under optimization. Reported by Antonio Menezes Leitao on the mailing list.

takikawa:
A variation on this program illustrates some unsound behavior (without the TR optimizer on even)

````
#lang typed/racket #:no-optimize

(struct (A) foo ([x : A]) #:mutable)

(struct (A) baz foo ())

(define (f [i : Integer]) : (foo Integer)
  (baz i))

(: x (foo Integer))
(define x (f 1))

(: y Any)
(define y x)

(if (baz? y) (set-foo-x! y "foo") 2)
(foo-x x)
````

Note that the result doesn't change with the optimizer on.

PR #209, opened 2015-10-05
Fix github issue #205

takikawa:
This PR fixes issue #205 and some related bugs I found, but unfortunately there is one test that doesn't pass right now.

I think that the test actually exhibits broken behavior, so I wanted a second opinion on what to do about it.

The problematic test is this one: https://github.com/racket/typed-racket/blob/master/typed-racket-test/succeed/poly-struct-parent.rkt

On my PR branch, line 25 fails to typecheck because the predicate on that line has a filter that overlaps with the `Box2` type. Unfortunately, the original behavior (that the types don't overlap) is also wrong because it makes that whole conditional branch into dead code (try adding `(+ 1 "foo")` in that cond branch). Anyone have suggestions on how to fix this test?

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
@@ -330,7 +330,7 @@
   (define parent-tys (map fld-t (get-flds parent-type)))
 
   (define names (get-struct-names nm nm fld-names #f #f))
-  (define desc (struct-desc parent-tys tys null #t #f))
+  (define desc (struct-desc parent-tys tys null #f #f))
   (define sty (mk/inner-struct-type names desc parent-type))
 
   (register-sty! sty names desc)
--- a/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
@@ -43,7 +43,12 @@
 ;; setters : Listof[Id] or #f
 (struct struct-names (struct-name type-name struct-type constructor extra-constructor predicate getters setters) #:transparent)
 
-;;struct-fields: holds all the relevant information about a struct type's types
+;; struct-desc holds all the relevant information about a struct type's types
+;; parent-fields : (Listof Type)
+;; self-fields : (Listof Type)
+;; tvars : (Listof Symbol)
+;; mutable: Any
+;; proc-ty: (Option Type)
 (struct struct-desc (parent-fields self-fields tvars mutable proc-ty) #:transparent)
 
 (define (struct-desc-all-fields fields)
--- a/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
@@ -48,8 +48,11 @@
 ;; self-fields : (Listof Type)
 ;; tvars : (Listof Symbol)
 ;; mutable: Any
+;; parent-mutable: Any
 ;; proc-ty: (Option Type)
-(struct struct-desc (parent-fields self-fields tvars mutable proc-ty) #:transparent)
+(struct struct-desc (parent-fields self-fields tvars
+                     mutable parent-mutable proc-ty)
+        #:transparent)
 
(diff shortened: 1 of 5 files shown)


typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -282,15 +295,28 @@
...
+             ;; no parent, so trivially true
+             ['() #t]))
+         (define desc
+           (struct-desc parent-fields types tvars mutable parent-mutable #f))
          (parsed-struct (make-Prefab key (append parent-fields types))
                         names desc (struct-info-property nm/par) #f)]
         [else
+         (define parent-mutable
+           ;; Only valid as long as typed structs must be
+           ;; either fully mutable or fully immutable
```
samth: Can we prepare for what happens when we fix this?

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -311,7 +337,8 @@
   (define parent-tys (map fld-t (get-flds parent-type)))
 
   (define names (get-struct-names nm fld-names #f))
-  (define desc (struct-desc parent-tys tys null #t #f))
+  ;; built-in structs are assumed to be immutable with immutable parents
+  (define desc (struct-desc parent-tys tys null #f #f #f))
```
samth: Is this true everywhere?

takikawa: Yes for the current set of built-in structs, no if that ever changes. This is certainly more correct than the status quo though.

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
takikawa: That would involve changing a lot more code that depends on this fact. I think it'd be better put in a different PR (and really I think we should just fix it in the next TR or something).

takikawa:
Any ideas on what to do about that failing test case?

samth:
I don't totally understand what you're saying about the test case.

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
samth: Ok, I'll assume that the built-in structs aren't going to grow some mutable data.

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
samth: Right, I just meant can we avoid adding more code that's obviously unsound when that happens.

And I agree we should fix that soon.

takikawa:
@samth I was saying that the original test case is wrong, in that it only succeeds to typecheck because a branch is declared "dead" but it really isn't. The PR causes the branch to be live again, but then the RHS doesn't typecheck. I wasn't sure what the best way to fix the test is though.

samth:
Right, but why doesn't `b` have the appropriate type there?

takikawa:
Because the filter on the predicate `Strict-Box2` specifies a `(Strict-Box2 Any)` and therefore that's the type of `b`. This isn't compatible with the return type, which is `(Box2 A)`.

samth:
Shouldn't we refine that to `(Strict-Box2 A)` and then have the test succeed?

takikawa:
I don't think we want to build in the logic that `(Foo Any)` intersected with `(Parent-Foo X)` is `(Foo X)` though because that's not valid in general. Or is that not what you're suggesting?

samth:
That is what I'm suggesting, but I think it is valid.

takikawa:
Well, I can't actually think of a counterexample so you may be right. I'll see if I can implement that and see if it works out.

samth:
Status?

takikawa:
Blocked on implementing that intersection. Unlikely to do it any time soon.

samth:
Ok. I'd like to get #205 fixed, since it's a soundness issue, so is there something simpler we can do? Alternatively, I can look at `restrict`.

takikawa:
Merged because the one remaining issue with this PR went away, possibly due to recent intersection type additions.

PR #217, opened 2015-10-16
Have optimizer ignore code more consistently

stamourv:
The optimizer relied implicitly on the following invariant: all expressions either have a type in the type table, or an entry in the ignore table.
With the added missed optimization logging in the second commit, the optimizer now relies on that invariant explicitly, which turns all the instances where it didn't hold into errors.
This PR restores that invariant everywhere.

This also requires a change to the contract system (https://github.com/racket/racket/pull/1093), since TR relies on it to ignore contract-related code, and the contract system was not always marking the relevant code.

I'm not very happy with the solution in the 3rd commit, so any feedback on how to improve it is welcome.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-objects.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-objects.rkt
@@ -7,7 +7,7 @@
          racket/format
          racket/list
          (typecheck signatures)
-         (types base-abbrev resolve subtype union utils)
+         (types base-abbrev resolve subtype type-table union utils)
          (rep type-rep)
          (utils tc-utils)
 
@@ -25,10 +25,21 @@
 (define-tc/app-syntax-class (tc/app-objects expected)
   #:literal-sets (kernel-literals object-literals)
   (pattern (dmo b cl
-            (#%plain-app list . pos-args)
-            (#%plain-app list (#%plain-app cons (quote names) named-args) ...))
+            (~and pos-arg-list (#%plain-app list . pos-args))
+            (~and named-arg-list (#%plain-app list (#%plain-app cons (quote names) named-args) ...)))
      #:declare dmo (id-from 'do-make-object 'racket/private/class-internal)
-     (check-do-make-object #'cl #'pos-args #'(names ...) #'(named-args ...)))
+     (begin0
+       (check-do-make-object #'cl #'pos-args #'(names ...) #'(named-args ...))
+       ;; synthesize a type for #'pos-arg-list, for the optimizer
+       (add-typeof-expr #'pos-arg-list
+                        (ret (for/fold ([res (-val '())])
+                                 ([a (in-list (reverse (syntax->list #'pos-args)))])
+                               (-pair (match (type-of a) [(tc-result1: t) t])
+                                      res))))
+       ;; making the optimizer ignore named args is conservative, but safe
+       ;; if we could give #'named-arg-list a type, then we'd be able to
+       ;; optimize it
+       (register-ignored! #'named-arg-list)))
   (pattern (dmo . args)
      #:declare dmo (id-from 'do-make-object 'racket/private/class-internal)
      (int-err "unexpected arguments to do-make-object"))
--- a/typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
@@ -273,8 +273,35 @@
        (ret (opt-unconvert (tc-expr/t #'fun)
                            (syntax->list #'(formals ...))))]
       ;; let
-      [(let-values ([(name ...) expr] ...) . body)
-       (tc/let-values #'((name ...) ...) #'(expr ...) #'body expected)]
+      [(let-values bindings . body)
+       (define bindings*
+         (syntax-parse #'body
+           #:literal-sets (kernel-literals)
+           ;; special case: let-values that originates from an application of a
+           ;; kw function. we may need to ignore contract-related arguments
+           [((kw-app1 (kw-app2 cpce s-kp fn kpe kws num) ; see tc-app/tc-app-keywords.rkt
+                      kw-list
+                      (kw-app3 list . kw-arg-list)
+                      . *pos-args))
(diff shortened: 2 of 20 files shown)


typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-objects.rkt
```diff
@@ -25,10 +25,21 @@
...
-            (#%plain-app list . pos-args)
-            (#%plain-app list (#%plain-app cons (quote names) named-args) ...))
+            (~and pos-arg-list (#%plain-app list . pos-args))
+            (~and named-arg-list (#%plain-app list (#%plain-app cons (quote names) named-args) ...)))
      #:declare dmo (id-from 'do-make-object 'racket/private/class-internal)
-     (check-do-make-object #'cl #'pos-args #'(names ...) #'(named-args ...)))
+     (begin0
+       (check-do-make-object #'cl #'pos-args #'(names ...) #'(named-args ...))
+       ;; synthesize a type for #'pos-arg-list, for the optimizer
+       (add-typeof-expr #'pos-arg-list
```
samth: Won't this give a (potentially) different type to `pos-arg-list` than what we actually use in `check-do-make-object`?

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
```diff
@@ -273,8 +273,35 @@
...
+                      . *pos-args))
+            #:declare cpce (id-from 'checked-procedure-check-and-extract 'racket/private/kw)
+            #:declare s-kp (id-from 'struct:keyword-procedure 'racket/private/kw)
+            #:declare kpe  (id-from 'keyword-procedure-extract 'racket/private/kw)
+            #:declare kw-app1  (id-from '#%app 'racket/private/kw)
+            #:declare kw-app2  (id-from '#%app 'racket/private/kw)
+            #:declare kw-app3  (id-from '#%app 'racket/private/kw)
+            #:when (contract-neg-party-property #'fn) ; contracted
+            ;; ignore the rhs which refers to a contract-lifted definition
+            (syntax-parse #'bindings
```
samth: This needs a better `else` case.

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
```diff
@@ -273,8 +273,35 @@
...
+              [(c1 [(contact-lhs) contract-rhs] cs ...)
+               ;; give up on optimizing the whole let, part of it is missing type info
+               ;; (not that this expansion could be optimized anyway)
+               (register-ignored! form)
```
samth: This means that `(blah (+ float-expr float-expr))` won't be optimized in some cases, right?

samth:
I still don't understand well enough why the big change to checking `let-values` is needed. What was happening before, and why wasn't that maintaining the invariant you describe?

stamourv:
@samth: The big change is needed because one of the RHSs contains code that can't be typechecked anymore, because it refers to a contract lift. It *used* to be typechecked, because the lift was not ignored, but we need to ignore that lift for the optimizer now.

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
stamourv: Correct.

stamourv: That only affects code inside calls to kw+contract functions, though.
Fixing it would require the optimizer to be able to handle partially typed expressions, which would be a big change.

samth:
Why do we need to ignore that lift? Can we have the contract system (or can we use some other means) distinguish between the things we actually need to ignore and the things we can handle?

stamourv:
I haven't found a criterion that would distinguish between those. That's part of the reason why I originally asked you whether "lift = ignored" was the invariant we wanted.

samth:
Well, the criteria we really want is "type checks = not ignored" :) but that's hard to operationalize.

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-objects.rkt
stamourv: `check-do-make-object` does not give a type to the list itself. It doesn't even see the list. That's why I have to reconstruct one here, otherwise that piece of code has no type.

stamourv:
That, and quite brittle, too. It's not clear whether any of those really can typecheck in the general case, versus just in the examples we've seen.

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
stamourv: What do you mean? Either we ignore one of the bindings, or we don't.

samth: No, the inner `syntax-parse` has only one clause, so if the outer one matches, and the inner one doesn't (for example, by having only one bindings clause), then you'll get a "bad syntax" error.

samth:
Well, the lifted module name expression should always type check -- it's quite simple. What other lifted expressions do we need to worry about?

stamourv:
There's the computation of negative blame parties, I think. And various box allocations from `opt/c`. The former sound like they may be tricky to typecheck.

samth:
I think the negative blame computation is the module name expression I referred to. Can you give an example?

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
stamourv: Oh, that. I don't think it an fail (unless the kw implementation starts emitting malformed lets), but I can add an internal error.

stamourv:
It's sometimes the module name, sometimes something else.
`#'(partially-applied-id lifted-neg-party)` is one of the things that get lifted, and that may be more complicated. Maybe. I'm not sure.

samth:
Ah, ok -- that sounds harder to check. Can you add a comment to the file with an example that generates something hard to typecheck?

stamourv:
Which file? I'll try to think of something. I don't think I've encountered any concrete examples.

samth:
The file with the big change.

stamourv:
Merged.

PR #221, opened 2015-10-20
extend define-new-subtype to polymorphic types

AlexKnauth:
So for example

``` racket
(define-new-subtype (Meters a) (meters a))
```

To allow `(Meters Real)`, `(Meters Number)`, or even `(Meters (Vectorof Real))`.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
@@ -21,7 +21,6 @@
   internal
 
   type-alias
-  new-subtype-def
   type-refinement
   typed-struct
   typed-struct/exec
@@ -29,12 +28,12 @@
   typed-require/struct
   predicate-assertion
   type-declaration
+  unsafe-type-declaration
   typed-define-signature
   typed-define-values/invoke-unit
   typecheck-failure
 
   type-alias?
-  new-subtype-def?
   typed-struct?
   typed-struct/exec?
   typed-define-signature?
@@ -63,6 +62,7 @@
                   assert-predicate-internal
                   declare-refinement-internal
                   :-internal
+                  unsafe-:-internal
                   typecheck-fail-internal
                   define-signature-internal
                   define-values/invoke-unit-internal))
@@ -141,8 +141,6 @@
 (define-internal-classes
   [type-alias
     (define-type-alias-internal name type args)]
-  [new-subtype-def
-    (define-new-subtype-internal name (constructor rep-type) #:gen-id gen-id)]
   [type-refinement
     (declare-refinement-internal predicate)]
   [typed-struct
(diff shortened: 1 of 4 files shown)


typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
```diff
@@ -157,6 +155,8 @@
     (assert-predicate-internal type predicate)]
   [type-declaration
     (:-internal id:identifier type)]
+  [unsafe-type-declaration
+    (unsafe-:-internal id:identifier type)]
   ;; the check field indicates whether this signature is being
```
pnwamk: `unsafe-:-internal` us too cute a name for inside TR's implementation, IMHO. I think it should be less punny on `:`.

Maybe `unsafe-declare-type` or similar?

typed-racket-test/succeed/define-new-subtype-poly.rkt
```diff
@@ -0,0 +1,13 @@
...
+
+(: m+ : (case-> [(Meters Real) (Meters Real) -> (Meters Real)]
+                [(Meters Number) (Meters Number) -> (Meters Number)]))
+(define (m+ a b)
+  (meters (+ a b)))
+
+(define x1 (m+ (meters 1) (meters 2)))
+(ann x1 (Meters Real))
+(define x2 (m+ (meters (ann 1 Number)) (meters 2)))
+(ann x2 (Meters Number))
```
pnwamk: 1. Can any of these tests be unit tests instead of integration tests? (i.e. in `typecheck-tests.rkt` perhaps)
2. There are no tests for this new feature failing when misused in any way -- I worry we can't be confident if this implementation is right without positive and negative tests. Are there any sensible tests of that variety we can add to this PR?

typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
AlexKnauth: I was following the convention in the previous line. If we decide to change that convention, it shouldn't be part of this pull request.

pnwamk: Aha, I missed that -- sorry.

typed-racket-test/succeed/define-new-subtype-poly.rkt
AlexKnauth: You're right, I should add negative tests. However, I'm a bit busy now, and this pull request can wait. Especially if we plan on replacing Distinction with intersection-based types like you've talked about before.

samth:
Coming back to this after a long time, I'm currently inclined to close this. I think this can be done more easily with intersections now.

AlexKnauth:
Or it can be re-done, `define-new-subtype` could be completely re-written to use intersection types instead of its own private type constructor.

samth:
Yes, that would be my feeling about the whole issue. Perhaps it could even be a library?

Since you agree, I'm going to close this PR.

PR #224, opened 2015-10-24
Memoize the Un constructor for unions

takikawa:
This commit seems to help cut down typechecking time. I tried compiling the math library and also `new-metrics.rkt` with/without the commit and saw about a 10% speedup.

Can someone else try to confirm the speedup?

### DIFF
--- a/typed-racket-lib/typed-racket/types/union.rkt
+++ b/typed-racket-lib/typed-racket/types/union.rkt
@@ -49,11 +49,16 @@
 ;; Normalizes representation by sorting types.
 ;; Type * -> Type
 ;; The input types can overlap and be union types
+(define Un-cache (make-weak-hash))
 (define Un
   (case-lambda 
     [() -Bottom]
     [(t) t]
     [args 
-     (define ts (foldr merge '()
-                       (remove-dups (sort (append-map flat args) type<?))))
-     (make-union* ts)]))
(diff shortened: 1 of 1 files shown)


takikawa:
Here's a benchmark script for running `new-metrics.rkt`:

````
#lang racket

(require benchmark
         plot/no-gui)

(define results
  (run-benchmarks
   (list "typed-racket-test/succeed/new-metrics.rkt")
   (list (list 'patch 'no-patch))
   (lambda (file patch?)
     (if (equal? patch? 'patch)
         ;; parameterize strings here to the two racket installs
         (system (format "~a -l ~a" "~/plt/racket-git/racket/bin/racket" file))
         (system (format "~a -l ~a" "~/plt/racket-fresh/racket/bin/racket" file))))
   #:num-trials 30
   #:extract-time 'delta-time))

(parameterize ([plot-x-ticks no-ticks])
  (plot-file
   (render-benchmark-alts
    ; default options
    (list 'patch)
    results)
   "/tmp/benchmark.png"
   'png
   #:title "memo vs. no-memo"
   #:x-label #f
   #:y-label "normalized time"))
````

stamourv:
I can see a speedup, but it's closer to 5%.
That's on a 2015 macbook air.

The patch looks good to me, and is worth it even for 5%.

samth:
Agreed.

PR #225, opened 2015-10-25
add separate read and write types for Box types

AlexKnauth:
The write type is contra-variant, and the read type is covariant. Unions of mutable box types make sense now, and this also adds support for read-only boxes, without requiring a separate type.
See also
https://github.com/racket/racket/pull/1100


AlexKnauth:
Still need to make a constructor for the more general box type, and fix printing to use that. What should it be called?

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -531,8 +531,8 @@
          [((Channel: t) (Evt: t*)) (subtype* A0 t t*)]
          [((Async-Channel: t) (Evt: t*)) (subtype* A0 t t*)]
          ;; Invariant types
-         [((Box: s) (Box: t)) (type-equiv? A0 s t)]
-         [((Box: _) (BoxTop:)) A0]
+         [((Box: s-w s-r) (Box: t-w t-r)) (subtypes* A0 (list t-w s-r) (list s-w t-r))]
```
stamourv: You may want to move those out of the `invariant types` section.

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -354,6 +354,12 @@ corresponding to @racket[trest], where @racket[bound]
 
 @ex[(box "hello world")]
 
+@defform[(Read-Only-Boxof t)]{A read-only @rtech{box} of @racket[t].}
+
```
stamourv: I think you want to elaborate a bit more here, especially explaining the relationship with regular `Box` types and subtyping. Maybe a discussion of the similarities with C's `const` keyword may help?

typed-racket-lib/typed-racket/types/union.rkt
```diff
@@ -37,6 +37,12 @@
     [(_ _) #:when (currently-subtyping?) (cons a b)]
     [((? (λ _ (subtype a b*))) _) b]
     [((? (λ _ (subtype b* a))) _) (list a)]
+    [((Box: a-w a-r) (list-no-order (Box: b-w b-r) bs ...))
```
stamourv: A brief comment explaining the logic would be nice.

stamourv:
Re more general type: You mean the constructor that exposes both type arguments? I don't know how much sense it would make to expose that.

I'm also not 100% clear on the use case. You mentioned better support for unions of box types, and I remember a discussion on IRC last week on that topic, but the particular example in that discussion left me unconvinced. Would you mind elaborating a bit?

FWIW, ISTR @ntoronto wanting something similar for vectors, to have a nicer API for the math library. I think he ended up going with arrays as functions (which have the right variance) to work around that.

Aside from the comments, implementation looks good.

AlexKnauth:
When I said "unions of box types," I was really thinking in my head "unions of vector types, when I do the same thing for vectors." For that I was thinking of [this](http://lists.racket-lang.org/users/archive/2014-July/063551.html) discussion on the mailing list, which started as talking about `(U (Vectorof Index) (Vectorof Integer))`, which came about because of using the `In-Indexes` type from `math/array`.

AlexKnauth:
I was also thinking of read-only boxes as `(U (Boxof a) (Boxof Nothing))`.

typed-racket-lib/typed-racket/types/subtype.rkt
AlexKnauth: Ok done.

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
AlexKnauth: I explained the subtyping and the relationship with other box types, but what do you mean about `const`?

typed-racket-lib/typed-racket/types/union.rkt
AlexKnauth: Ok done.

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
stamourv: C has a `const` keyword, that you can put on variables that essentially means "I will not modify that" (that also does not imply immutability). That sounds similar, and maybe pointing out that similarity will help people familiar with C understand TR's read-only boxes.

AlexKnauth: Ok, I did not know that. I learned something. Still, doesn't it seem a little too confusing to refer to `const`? Reading through some stuff, I got confused about pointer to const versus const pointer, then eventually figured out what you meant. It'll be simpler to just show an example of a mutable box being passed as a read-only one.

AlexKnauth:
Re more general type: 
Mostly because of printing the types.

For example `set-box!` can take a box that allows `a` to be written, but allows `Any` to be read, which allows the `set-box!` of a union to work. 
Right now the type of `set-box!` prints as:
```racket
(All (a)
  (-> (Unknown
       Type:
       #(struct:Box
         1416
         #(struct:combined-frees #hasheq((a . #(struct:Contravariant))) ())
         #(struct:combined-frees #hasheq() ())
         #f
         box
         a
         Any))
      a
      Void))
```
Where it should really be:
```racket
(All (a) (-> (Box/Write-Read a Any) a Void))
```
Is `Box/Write-Read` the best name for this?

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
stamourv: I agree that explaining read-only boxes in terms of `const` would be confusing. I'm suggesting to just mention it, as in "if you know `const` from C, this is similar".

stamourv:
Ah, that makes sense.
What about `Read-Write-Box`? It's consistent with other type names.

AlexKnauth:
I wanted `write` to be before `read` to make the argument order clear, but `Write-Read-Boxof` sounds good. 

(Though I'm also wondering, would changing `Read-Only-Boxof` to `Boxof/Read-Only` instead make sense?)

stamourv:
Usually, `read` comes before `write`. Could you change it in the box type representation?

AlexKnauth:
Write before read is consistent with `Parameterof`, `parameter/c`, and `box/c`, as well as with the co- and contra- variance of function types. Input before output. I could change it, but then it would be inconsistent with those.

stamourv:
Oh, ugh. In regular English, though, "read" usually goes before "write"...
Consistency with other bits of Racket is more important, let's do w-r.

AlexKnauth:
Would a set of types like this make sense?
```racket
(Boxof/Write-Read w r) ; can write w, read r
(Boxof/Write w) ; = (Boxof/Write-Read w Any)
(Boxof/Read r) ; = (Boxof/Write-Read Nothing r)
(Boxof t) ; = (Boxof/Write-Read t t)
```
Or this, depending on what we think is better:
```racket
(Write-Read-Boxof w r) ; can write w, read r
(Write-Boxof w) ; = (Write-Read-Boxof w Any)
(Read-Boxof r) ; = (Write-Read-Boxof Nothing r)
(Boxof t) ; = (Write-Read-Boxof t t)
```
Would that set of types make sense? Does the `Boxof/Write` type make sense? (It's the type that `set-box!` requires as an argument)

stamourv:
I prefer the latter.
If we're going to expose the second in the type for `set-box!`, then we should provide it.

samth:
I think I prefer the former, but why not just have `Boxof` work like `Parameterof` and take either 1 or 2 type arguments?

AlexKnauth:
Right now, `Boxof` is a polymorphic type, not a type constructor syntax (like `->` for example). If it were a syntax type constructor, there would have to be a special case in `parse-type`, right? Also is there anything you can do with polymorphic types that you can't do with type constructor syntax identifiers? I thought it would be better to keep it as a normal type.

AlexKnauth:
Ok, I just renamed `Read-Only-Boxof` to `Boxof/Read` and added `Boxof/Write` as well. Read operations such as `unbox` take `(Boxof/Read a)` as arguments, and write operations such as `set-box!` take `(Boxof/Write a)` as arguments.

AlexKnauth:
What do you think? Is it worth another special case in `parse-type` to make `Boxof` take an optional argument, or should I leave it as is?

samth:
Yes, I think it's worth the special case. What do other people think?

stamourv:
Sure.

samth:
@AlexKnauth what's the current status here?

AlexKnauth:
I need to rebase again, and then I need to implement the new `parse-type` case for `Box` for the optional argument. I think I'll re-visit this that after the current `cast` and `any-wrap/c` stuff is done.

samth:
@AlexKnauth do you want to pick this up again, or should we close this?

AlexKnauth:
I'll pick it up again starting this Thursday.

AlexKnauth:
Who should I ask about optimizing `box/sc` contracts with contravariant and covariant write and read fields?

samth:
What do you mean? Let's add the semantics first, and then do optimization.

AlexKnauth:
That makes sense. However, the reason I asked is that without me adding anything specifically for `box/sc` optimizations, the optimizer still seems to be trying to optimize parts of them away, resulting in failing optimizer tests. These tests have comments like `don't optimize if any contracts are non-flat --- but do optimize under guarded constructors`, and they test that it goes through the optimizer unchanged. The problem is that it _is_ trying to optimize them, and I'm not sure what the right thing to do is.

bennn:
Those are my comments & tests. If `box/sc` is no longer invariant, they need to be updated.

The idea with the tests is:
- if there's an `(or/sc sc_0 sc_1)`
- and `sc_0` is not flat
- then the optimized contract should be something like `(or/sc sc_0+ sc_1+)` and NOT `any/sc`

To change the tests, I think you just need to change what's inside the expected `or/sc`s. You could also change them to use `vector/sc` instead of `box/sc`.

AlexKnauth:
Is the "actual" correct for all 10 of those failures, for both "positive" and "negative" as trusted sides?

bennn:
yes

AlexKnauth:
The only remaining test failures are these, where the "expected" and "actual" look the same visually printed out, but the test thinks they're different:
```
--------------------
tc-expr tests > Typechecker tests > Unit Tests > Typed Racket Tests > 1436 (if (zero? (random 2)) (box hello) (box (quote hello)))
FAILURE
location:   ("<pkgs>/typed-racket-test/unit-tests/typecheck-tests.rkt" 1436 8 60243 109)
actual:     #(struct:tc-results (#(struct:tc-result (Boxof/Read (U String Symbol)) (Top | Bot) -)) #f)
expected:   #(struct:tc-results (#(struct:tc-result (Boxof/Read (U String Symbol)) (Top | Bot) -)) #f)
tc-expr did not return the expected value. The types don't match.
--------------------
--------------------
tc-expr tests > Typechecker tests > Unit Tests > Typed Racket Tests > 1440 (if (zero? (random 2)) ((inst box Natural) 3) ((inst box Integer) -5))
FAILURE
location:   ("<pkgs>/typed-racket-test/unit-tests/typecheck-tests.rkt" 1440 8 60472 108)
actual:     #(struct:tc-results (#(struct:tc-result (Boxof Nonnegative-Integer Integer) (Top | Bot) -)) #f)
expected:   #(struct:tc-results (#(struct:tc-result (Boxof Nonnegative-Integer Integer) (Top | Bot) -)) #f)
tc-expr did not return the expected value. The types don't match.
--------------------
```

AlexKnauth:
When `tc-e` gets the type, it doesn't normalize it, and subtyping on unions of boxes doesn't work without normalizing it first. How should I fix this? Do I fix subtyping to merge unions of boxes before going through them?

samth:
I don't understand why this is needed. What's different than how pairs work?

AlexKnauth:
You're right. The analogous case for `Boxof` is this:

For merging boxes `(Boxof Aw Ar)` and `(Boxof Bw Br)`, if the merged version works
```
(Boxof (∩ Aw Bw) (U Ar Br)) <: (Boxof Cw Cr)
```
Then `Cw <:` both `Aw` and `Bw`, and `Ar` and `Br` are both `<: Cr`. Using those, the non-merged verision
```
(U (Boxof Aw Ar) (Boxof Bw Br)) <: (Boxof Cw Cr)
```
Should also work because `(Boxof Aw Ar) <: (Boxof Cw Cr)` and `(Boxof Bw Br) <: (Boxof Cw Cr)`.

So I think I need to replace each `tc-e` test I currently have with two tests, one `tc-e` with the non-merged type, and one subtype test checking that the non-merged type is a subtype of the merged type.

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -351,10 +351,42 @@ corresponding to @racket[trest], where @racket[bound]
...
 
-@defform[(Boxof t)]{A @rtech{box} of @racket[t]}
+@defform*[[(Boxof t)
+           (Boxof write-t read-t)]]{
+  The first form is a @rtech{box} of @racket[t].
+
+  The second form is a @rtech{box} that supports write operations like
+  @racket[set-box!] taking @racket[write-t], and read operations like
+  @racket[unbox] returning @racket[read-t]. For this type to make sense,
+  @racket[write-t] should be a subtype of @racket[read-t]. This form can
```
bennn: Should "should be" say "must be"?

typed-racket-lib/typed-racket/types/union.rkt
```diff
@@ -30,6 +30,19 @@
...
       [(subtype t* t) (list t)]
       [(subtype t t*) ts]
+      ;; the union of two box types is a box type where the write type has to
+      ;; satisfy both write types, and the read type can satisfy either of the
+      ;; two read types
+      [(and (Box? t) (ormap Box? ts))
+       (match* (t ts)
+         [((Box: a-w a-r) (list-no-order (Box: b-w b-r) bs ...))
+          (define w
+            ;; should this use some sort of intersection, or would that
```
bennn: Can you use `intersect` from `infer/intersect.rkt` here?

I think that would be better than the `cond`

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
AlexKnauth: No, because the type `(Boxof A B) A </: B` can still exist, it's just impossible to write an expression that produces a value of that type. (Unless you use `cast`, in which case it's nearly impossible for that value to keep to that contract for everything you could do to it. "nearly" because weird uses of impersonators might be able to fake it, but it doesn't make sense in normal code.)

typed-racket-lib/typed-racket/types/union.rkt
AlexKnauth: I'm not sure what the implications of using `intersect` might be. @pnwamk what do you think?

AlexKnauth:
@bennn That would mean replacing `-BoxTop` with `(-Boxof/Read Univ)` [here](https://github.com/racket/typed-racket/blob/4a869d97d19059828e8a8f36c1e48cd715c79fc2/typed-racket-lib/typed-racket/base-env/base-env.rkt#L60).

Would that actually change anything? `BoxTop` already allows reading `Any` from it.

bennn:
> @bennn That would mean replacing -BoxTop with (-Boxof/Read Univ) here.

yes

> Would that actually change anything? BoxTop already allows reading Any from it.

I was thinking it would refine `(Boxof String)` to `(Boxof/Read String)`

AlexKnauth:
@bennn No, it wouldn't, because `(Boxof String)` intersected with `(Boxof/Read String)` should still be `(Boxof String)` or the equivalent. Changing it to `Boxof/Read` wouldn't be a refinement, it would be a "loss of information," assigning it to a supertype.

Speaking of intersections, they weren't there when I first started with this pull request. Should I have to change anything about intersections to handle intersections of boxes?

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -351,10 +351,42 @@ corresponding to @racket[trest], where @racket[bound]
...
+  be used to restrict writing to the box, or to safely express uncertainty
+  about the values other code could put in the box.
+}
 
 @ex[(box "hello world")]
 
+@defform[(Boxof/Read t)]{
+  A @rtech{box} of @racket[t] that only supports read-only operations
+  such as @racket[unbox]. A @racket[(Boxof/Read t)] is a subtype of
+  @racket[(Boxof t)], and a @racket[(Boxof/Read a)] is a subtype of
```
bennn: typo: `(Boxof/Read t)` is _not_ a subtype of `(Boxof t)`

AlexKnauth: I’ll change it to say supertype.

AlexKnauth:
We need to think of how this pull request should handle programs like this. Should we make `box-immutable` return a `(Boxof X)` instead of a `(Boxof/Read X)` for backwards-compatibility with programs like this?
```racket
#lang typed/racket

(define b (box-immutable 5))

(: f : (Boxof Integer) -> Integer)
(define (f b)
  (unbox b))

(f b)
```

Typed Racket currently lets this pass, but this pull request makes it an error, because `box-immutable` returns a read-only box, which can't be used as a `(Boxof Integer)`.

samth:
I think the point of this PR should be to change what programs people can write, by allowing them to specify read and write types, rather than to change the meaning of existing programs. We definitely shouldn't break that example.

AlexKnauth:
So `box-immutable` applied to (generalized) `X` should return a `(Boxof X)`, not `(Boxof/Read X)`. I need to change that back.

typed-racket-lib/typed-racket/types/union.rkt
AlexKnauth: @pnwamk In what situations is it appropriate to use `intersect`?

pnwamk: Hey sorry -- I'm a little slow coming up to speed on this.

`intersect` is supposed to compute the logical "and" of two types (performing some simplifications where possible), so if something is both of type `T1` and type `T2`, then saying it is of type `(intersect T1 T2)` should always be fine.

I guess the only "exception" I can think of is in a few places our handling of intersections is not complete (i.e. at the moment we do not fully handle the application of an intersection of function types -- but this isn't a problem with `intersect` or intersection types, it's just a deficiency in our code that handles function application at the moment).

Anyway, did that answer your question?

AlexKnauth: So I'm guessing it is appropriate to use `intersect` in `normalize-type` to determine the "write" type of the merged result of a union of two box types?

pnwamk:
I can’t think of a reason why you should avoid using intersect in
normalize-type. Should be fine!

typed-racket-lib/typed-racket/types/union.rkt
AlexKnauth: (Re: https://github.com/racket/typed-racket/pull/225#issuecomment-354801561)
Okay, I've changed this to use `intersect`.

AlexKnauth:
After seeing discussion #1129, I'm thinking this might need a more generalized way of dealing with the opposite-variance-type-pairs for types such as `MPairof` that have multiple type-arguments. If I did the same thing to `MPairof` that I'm doing with `Boxof` here, it would be "ugly" to just put in 4 type-arguments with no extra syntax so signal that they come in opposite-variance pairs. A type-rep data structure for opposite-variance pairs might also be useful, to re-use the common subtyping logic.

Issue #203, opened 2015-09-23
`any-wrap/c` is very restrictive on opaque structures


samth:
This program currently produces a contract error:

```
#lang racket
(require typed-racket/utils/any-wrap)
(struct s ())
(contract any-wrap/c (s) 'a 'b)
```

However, it shouldn't need to.

This was discovered by @giels and reported on #racket.

samth:
Right now, the error comes because of [this code](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/utils/any-wrap.rkt#L54-L55). Those lines error when the struct type is opaque.

However, I think that this error is unecessary. My reasoning is as follows:
 * Opaque structs are potentially dangerous here because they prevent `any-wrap/c` from chaperoning some structure field.
 * This would be exploitable by the untyped code having an accessor for that field and accessing the relevant data, which would then be unprotected.

But how did the untyped code get that accessor? There are a few possibilities:
1. The untyped side defined the accessor. That means that the untyped code defined the structure, and handed over the constructor to the typed side. But then all of the relevant fields (the ones in the untyped part of the structure) are already protected by whatever contract was generated for the constructor.
2. The typed side defined the accessor, and provided it to the untyped side (via some contracted mechanism, by assumption). Then there would be a contract on the accessor, which would protect the _result_ of the accessor (if that result is higher-order).

So therefore, I this this check isn't needed. But I'd like other people's (such as @rfindler @takikawa @mflatt @endobson @sstrickl) thoughts here.

samth:
IRC discussion leading to this starts here: https://botbot.me/freenode/racket/2015-09-23/?msg=50351483&page=2

bennn:
Is there any "legal" way for untyped code to access struct fields without the accessor functions? (I thought this was possible with `struct->vector`, but I haven't gotten that to work.)

Otherwise, I agree about the accessors.

samth:
You can use struct->vector, but you need a sufficiently powerful inspector (TR has other soundness issues if you have one of those). Similarly, if you can get the struct type, you can make your own accessor function, but that requires the same inspector.

Issue #202, opened 2015-09-23
#:opaque fails to check inputs to predicate, is unsound


samth:
This program has a type error that's not caught:

```
#lang racket/base

(module m1 racket/base
  (define (bad x) (set-box! x 5) #t)
  (provide bad))

(module m2 typed/racket/base
  (require/typed (submod ".." m1) [#:opaque T bad])
  (: b (Boxof String))
  (define b (box "hi"))
  (if (bad b) #t #f)
  (string-append (unbox b) ""))

(require 'm2)
```

The solution is to protect the opaque predicate (here `bad`) with a contract such as `(-> Any Any)`.

samth:
Currently, because of #203, adding this contract would break almost everything that uses opaque types.

samth:
IRC discussion leading to this here: https://botbot.me/freenode/racket/2015-09-23/?msg=50351483&page=2

PR #226, opened 2015-10-27
optimize `(-> Any Bool)` contract

bennn:
Compile `(-> Any Bool)` types to an `(or/c struct-predicate-procedure? (-> any/c boolean?))` contract.
This avoids a great many indirections when using a struct (constructor, predicate, accessor ...) across a typed/untyped boundary.

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -16,7 +16,7 @@
  racket/format
  racket/dict
  syntax/flatten-begin
- (only-in (types abbrev) -Bottom)
+ (only-in (types abbrev) -Bottom -Boolean)
  (static-contracts instantiate optimize structures combinators)
  ;; TODO make this from contract-req
  (prefix-in c: racket/contract)
@@ -361,6 +361,17 @@
          (if numeric-sc
              (apply or/sc numeric-sc (map t->sc non-numeric))
              (apply or/sc (map t->sc elems)))]
+        [(and t (Function: arrs))
+         #:when (any->bool? arrs)
+         ;; Avoid putting (-> any boolean) contracts on struct predicates
+         ;; Optimization: if the value is typed, we can assume it's not wrapped
+         ;;  in a type-unsafe chaperone/impersonator and use the unsafe contract
+         (let* ([unsafe-spp/sc (flat/sc #'struct-predicate-procedure?)]
+                [safe-spp/sc (flat/sc #'struct-predicate-procedure?/c)]
+                [optimized/sc (if (from-typed? typed-side)
+                                  unsafe-spp/sc
+                                  safe-spp/sc)])
+           (or/sc optimized/sc (t->sc/fun t)))]
         [(and t (Function: _)) (t->sc/fun t)]
         [(Set: t) (set/sc (t->sc t))]
         [(Sequence: ts) (apply sequence/sc (map t->sc ts))]
@@ -768,6 +779,15 @@
           [(Name: _ _ #f) (escape #t)]
           [_ type])]))
     #f))
+
+;; True if the arities `arrs` are what we'd expect from a struct predicate
+(define (any->bool? arrs)
+  (match arrs
+    [(list (arr: (list (Univ:))
+                 (Values: (list (Result: (== -Boolean) _ _)))
+                 #f #f '()))
+     #t]
+    [_ #f]))
 
 (module predicates racket/base
   (require racket/extflonum)
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -72,6 +72,7 @@
          racket/struct-info
          syntax/struct
          syntax/location
+         (for-template "../utils/any-wrap.rkt")
          "../utils/tc-utils.rkt"
          "../private/syntax-properties.rkt"
          "../typecheck/internal-forms.rkt"
@@ -467,7 +468,7 @@
                                   si))
 
                          (dtsi* () spec ([fld : ty] ...) #:maker maker-name #:type-only)
-                         #,(ignore #'(require/contract pred hidden (any/c . c-> . boolean?) lib))
+                         #,(ignore #'(require/contract pred hidden (or/c struct-predicate-procedure?/c (c-> any-wrap/c boolean?)) lib))
                          #,(internal #'(require/typed-internal hidden (Any -> Boolean : nm)))
                          (require/typed #:internal (maker-name real-maker) nm lib
                                         #:struct-maker parent
--- a/typed-racket-lib/typed-racket/utils/any-wrap.rkt
+++ b/typed-racket-lib/typed-racket/utils/any-wrap.rkt
@@ -120,4 +120,10 @@
    #:projection (λ (blame) (λ (val) (((val-first-projection blame) val) #f)))
    #:val-first-projection val-first-projection))
 
-(provide any-wrap/c)
+;; Contract for "safe" struct predicate procedures.
+;; We can trust that these obey the type (-> Any Boolean).
+(define (struct-predicate-procedure?/c x)
+  (and (struct-predicate-procedure? x)
+       (not (impersonator? x))))
+
+(provide any-wrap/c struct-predicate-procedure?/c)
--- a/typed-racket-test/succeed/pr226-variation-1.rkt
+++ b/typed-racket-test/succeed/pr226-variation-1.rkt
@@ -1,0 +1,18 @@
+#lang typed/racket
+
+;; Struct predicates should not be wrapped in a contract
+;;  when they cross a typed/untyped boundary.
+;; We know they're safe! Don't suffer the indirection cost.
+
+(module untyped racket
+  (struct s ())
+  (provide (struct-out s)))
+
+(require/typed 'untyped
+  [#:struct s ()])
+
+(require/typed racket/contract/base
+  [has-contract? (-> Any Boolean)])
+
+(when (has-contract? s?)
+  (error 'pr226 "safe struct predicate was wrapped in a contract"))
--- a/typed-racket-test/succeed/pr226-variation-2.rkt
+++ b/typed-racket-test/succeed/pr226-variation-2.rkt
@@ -1,0 +1,20 @@
+#lang typed/racket
+
+;; Chaperoned struct predicates must be wrapped in a contract.
+;; (Even though `struct-predicate-procedure?` will return
+;;  true for these values)
+
+(module untyped racket
+  (struct s ())
+  (define s?? (chaperone-procedure s? (lambda (x) (x) x)))
+  ;; provide enough names to trick #:struct
+  (provide s struct:s (rename-out [s?? s?])))
+
+(require/typed 'untyped
+  [#:struct s ()])
+
+(define (fail-if-called)
+  (error 'pr226 "Untyped code invoked a higher-order value passed as 'Any'"))
+
+(with-handlers ([exn:fail:contract? (lambda (e) 'success)])
+  (s? fail-if-called))
--- a/typed-racket-test/succeed/pr226-variation-3.rkt
+++ b/typed-racket-test/succeed/pr226-variation-3.rkt
@@ -1,0 +1,25 @@
+#lang typed/racket
+
+;; Untyped should not be able to pass arbitrary code in
+;;  in place of a struct predicate.
+
+(module untyped racket
+  (struct s ())
+  (define (s?? x)
+    (when (box? x)
+      (set-box! x (void)))
+    #t)
+  (provide s struct:s (rename-out [s?? s?])))
+
+(require/typed 'untyped
+  [#:struct s ()])
+
+(: suitcase (Boxof '$$$))
+(define suitcase (box '$$$))
+
+(with-handlers ([exn:fail:contract? (lambda (x) (void))])
+  (s? suitcase)
+  (void))
+
+(unless (and (eq? '$$$ (unbox suitcase)))
+  (error 'pr226 "THEY SLIPPED US A RINGER"))
--- a/typed-racket-test/unit-tests/contract-tests.rkt
+++ b/typed-racket-test/unit-tests/contract-tests.rkt
@@ -224,6 +224,44 @@
               (t-sc (-lst Univ) (listof/sc any-wrap/sc))
               (t-sc (Un (-lst Univ) -Number) (or/sc number/sc (listof/sc any-wrap/sc)))
 
+              ;; Github pull request #226
+              (let ([ctc (-> Univ -Boolean)])
+                ;; Ordinary functions should have a contract
+                (t-int ctc
+                       (lambda (f) (f 6))
+                       (lambda (x) #t)
+                       #:untyped)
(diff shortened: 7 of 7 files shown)


samth:
I was talking about this here the other day, and one thing I realized might be a problem is if the function is a chaperone. Does that still answer true to `struct-predicate-procedure?`?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -769,6 +773,19 @@
           [_ type])]))
     #f))
 
+;; True if the arities `arrs` are what we'd expect from a struct predicate
+(define (any->bool? arrs)
+  (and (= 1 (length arrs))
```
samth: This could be written much more elegantly with `match`.

takikawa:
Can you add a test for this? Maybe here: https://github.com/racket/typed-racket/blob/master/typed-racket-test/unit-tests/contract-tests.rkt

bennn:
Originally suggested by @samth in the [gradual-typing-performance](https://github.com/nuprl/gradual-typing-performance/commit/f804ab479d28ac318c03cc726a526209ee5d17d3) repo (sorry, private for now).

I re-ran some results from that repo/paper on Racket 6.3.0.2 with & without this change. Here's one graph. The top row is from our paper, the middle row is 6.3.0.2, the bottom row is 6.3.0.2 + this pull request.

(The blue line gives the number (Y) of programs -- out of all the possible ways of gradually typing this "snake" game with 8 modules -- that run no more than X times slower than the fully-untyped game)

screen shot 2015-10-27 at 2 21 14 pm

takikawa:
@samth:

````
-> (struct foo ())
-> (struct-predicate-procedure? (chaperone-procedure foo? #f))
#t
````

File: typed-racket-lib/typed-racket/private/type-contract.rkt (hunk 1 above)
Comment (takikawa) [inline typed-racket-lib/typed-racket/private/type-contract.rkt]: Also the check using `string=?` doesn't look robust. We should be able to directly match on the type.

Comment (bennn) [inline typed-racket-lib/typed-racket/private/type-contract.rkt]: Yes, this needs to be revised.
(I knew a cleaner way should exist)

**Comment by bennn** [comment]
@takikawa just pointed out that the bottom row shows a worse max overhead. I'm going to re-run to make sure this isn't experimental error (running too many things concurrently), but for now here's another set of results.

screen shot 2015-10-27 at 2 30 00 pm

**Comment by samth** [comment]
Ok, so we need a `(not chaperone?)` check here. Evidence:

```
#lang racket

(struct s ())

(define s* (chaperone-procedure s? (λ (x) (x 0) x)))

(struct-predicate-procedure? s*)

(s* string-append)
```

**Comment by bennn** [comment]
I updated the contract with `(not chaperone?)`, but I still need help matching `(-> any bool)` and writing tests.

- Is there a match expander for `Any`, specifically? I tried `Univ:`, but that seems to match any type.
- Is there a matcher for `Boolean`? (I also realized that `-Boolean` in the tests had a different string representation than the surface-level `Boolean`. Should I be matching against a `Union:`?)

File: typed-racket-lib/typed-racket/private/type-contract.rkt [hunk 2]
```diff
@@ -769,6 +775,17 @@
           [_ type])]))
     #f))
 
+;; True if the arities `arrs` are what we'd expect from a struct predicate
+(define (any->bool? arrs)
+  (match arrs
+   [(list (arr: (list Univ:) x #f #f '()))
```
samth: `Univ:` is just a name, which you're binding. You want `(Univ:)`.

samth: Use `(== -Boolean)` to check for equality to `-Boolean`.

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -361,6 +361,12 @@
          (if numeric-sc
              (apply or/sc numeric-sc (map t->sc non-numeric))
              (apply or/sc (map t->sc elems)))]
+        [(and t (Function: arrs))
+         #:when (any->bool? arrs)
+         (or/sc (and/sc
+                  (flat/sc #'struct-predicate-procedure?)
+                  (flat/sc #'(compose1 not chaperone?)))
```
samth: Don't use `compose` when you care about performance, unless you're sure it's inlined.

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -769,6 +775,18 @@
           [_ type])]))
     #f))
 
+;; True if the arities `arrs` are what we'd expect from a struct predicate
+(define (any->bool? arrs)
+  (match arrs
+   [(list (arr: (list (Univ:))
+                (Values: (list (Result: (Union:
+                  (or (list (Value: #t) (Value: #f))
```
samth: Use `(== -Boolean)`, like I said.

Also, something seems to be wrong with the indentation in your DrRacket. :)

typed-racket-lib/typed-racket/private/type-contract.rkt
bennn: `-Boolean` wasn't bound in the module, but `Union:` works.

samth: Then figure out where to `require` it from. Repeating the definition of booleans isn't the right solution.

bennn:
Now featuring unit tests. Does this look good?

(I'm currently re-running 2 benchmarks, will give an update tomorrow.)

samth:
This looks good to me. I _think_ we only need the `chaperone?` check when we're importing the function from untyped, but optimizing that might not be useful.

bennn:
Fineeee

rfindler:
@bennn @takikawa has the "TR generating contracts that defeat the contract optimization" problem been fixed?

takikawa:
@rfindler It has not. I haven't had time to investigate further and it's tricky. My last thought on this was that doing an inlining pass after everything might work.

rfindler:
I know you said that before, but since it requires only that top constructor to be "inlined" suggests a simple rewrite right at the end should be easy.

I know it won't matter much with the magnitude of the results from Ben's graphs, but it still bugs me.

bennn:
Re-ran snake: there's no difference between the new (optimized) results and the current typed racket. I think `require/typed` might have killed the optimization.

samth:
@bennn are you saying you re-ran the experiment shown in your first figure, and the results are radically different than you posted before?

bennn:
yes
screen shot 2015-10-28 at 1 40 12 pm

rfindler:
@bennn can you explain the issues around `(-> Any Bool)`? The contract system has an optimization for the corresponding contract `(-> any/c boolean?)` that avoids wrapping struct predicates so I'm wondering if that's somehow related to what you're talking about here.

bennn:
@samth it looks like I mixed up the results, and the data from `10-27` was optimized (with `or/c struct-predicate-procedure? (-> any bool)`)

@rfindler that's exactly what I'm hoping to implement for the `(-> Any Bool)` type -- I want to avoid wrapping struct predicates because we know they obey their type.

I didn't realize the contract system already did this (with `predicate/c`, right?). I'd like to just use that contract, but I can't because of the issue Sam pointed out with chaperones.

(We'd either need to change the contract library, or do all the checking this pull request already does.)

```
#lang racket/base

(module s racket/base
  (require racket/contract/base)
  (struct s ())
  (define s?-anyb s?)
  (define s?-pred s?)
  (define s?-chap (chaperone-procedure s? (lambda (x) (x 0) x)))
  (provide
   (contract-out
     (s?-anyb (-> any/c boolean?))
     (s?-pred predicate/c)
     (s?-chap predicate/c))))

(require racket/contract/base 's)

(for ([s? (list s?-anyb s?-pred s?-chap)]
      [name (list "s?-anyb" "s?-pred" "s?-chap")])
  (printf "~a : has-contract? ~a, chaperone? ~a\n" name (has-contract? s?) (chaperone? s?)))

(s?-anyb string-append)
(s?-pred string-append)
(s?-chap string-append) ;; string-append, contract violation
```
We would want to catch the chaperone in a type error.

samth:
Additionally, Typed Racket doesn't generate `any/c` for `Any`, since we might be handing out higher-order values. When we're providing typed values with the type `(Any -> Boolean)`, neither that issue nor the chaperone issue comes up, so we could rely on the existing contract system support there.

@bennn which direction did the issues come up when you measured this before?

rfindler:
@benn the contract `(-> any/c boolean?)` used to be the same as `predicate/c`. I guess I broke that when I added a different optimization to the contract system. 

I'm not sure what you're getting at with this example. I think I understand this code and I don't see why the results it produces is wrong.

samth:
@rfindler If we used `predicate/c` and didn't have the chaperone check, then if the call to `(s?-chap string-append)` could be in a typed module that had imported `s?-chap` with the type `(-> Any Boolean)` and would then misuse the typed `string-append`.

bennn:
@samth it was mostly untyped data used in typed code

@rfindler you're write, I should have said "imagine `require/typed`".

Benn:
I think you put the wrong Benn in the loop.
Cheers
Le 28 oct. 2015 19:00, "Benjamin Greenman" <notifications@github.com> a
écrit :

samth:
@bennn You could avoid the `chaperone?` check when we're providing the value to untyped code. The `typed-side?` boolean tells you that information.

rfindler:
Sorry, Benoît!

rfindler:
@samth @bennn I am not getting how you are concluding from this example that the contract system should include a `chaperone?` check in its optimization. I can see how you need to use TR's `Any` and not `any/c` and thus the contract system's optimization isn't helpful to TR and so it has to be duplicated there somehow. I don't see how the require/provide in TR is relevant unless you are assuming that you will never allow programs like this to typecheck:

```
#lang typed/racket
(: f (-> Any Any))
(define (f x)
  (if (and (procedure? x)
           (procedure-arity-includes? x 1))
      (x 0)
      0))
```

Am I missing something I should know to be able to help TR with a better contract check?

samth:
@rfindler First, I'm not suggesting that the contract system should add a `chaperone?` check. The `chaperone?` check, like the difference in `Any`, is something only TR has to deal with.

Second, no, that program can't ever type check. Here's a counter-example:

```
(: g : String -> String)
(define (g s) (string-append s s))
```

If your program type checks, then `(f g)` type checks, which leads immediately to an error that TR was supposed to prevent.

Third, here's the example that shows that the `chaperone?` check is needed;

```
(module s racket/base
  (require racket/contract/base)
  (struct s ())
  (define s?-chap (chaperone-procedure s? (lambda (x) (x 0) x)))
  (provide s?-chap))

(module t typed/racket/base
  (require/typed s [s?-chap (-> Any Boolean)])
  (: g : String -> String)
  (define (g s) (string-append s s))
  (s?-chap g))
```

This program works in TR today, because currently the contract generated for `(-> Any Boolean)` is `(-> any-wrap/c boolean?)`. Instead, if we changed it to `(or/c struct-predicate-procedure? (-> any-wrap/c boolean?))`, then we'd pass 0 to `g` and get an error from `string-append`. So we switch to the contract that @bennn causes us to generate, and this procedure gets treated just like it does today.

Hopefully that makes things clearer.

rfindler:
I see what you mean about the `g` function. Given that maybe you could use `(-> any/c boolean?)` in a TR -> R position, if you wanted. In general, could you use `any/c` for `Any` when the value is flowing back into TR? (Or probably you already do that?)

samth:
Yes, if you look at @bennn's latest commit, it generates `(or/c struct-predicate-procedure? (-> any/c boolean?))` in that case, and only includes the `chaperone?` check in the other direction. And yes, we already use `any/c` when you write `Any` for a value flowing into TR.

rfindler:
Okay, well the contract system has restored that optimization.

bennn:
@samth, the restored optimization doesn't seem to apply in typed racket. That's good -- was that intentional? I thought `(type->contract (-> Any Boolean))` was indistinguishable from `(-> any/c boolean?)`.

```
#lang typed/racket/base

(module data racket/base
  (struct s ())
  (provide s?))

(require/typed 'data
  [s? (-> Any Boolean)])
(require/typed racket/contract/base
  [has-contract? (-> Any Boolean)])

(has-contract? s?) ;; #t. This is #f in #lang racket
```

samth:
@bennn The issue is related to the `any-wrap/c` contract that I mentioned earlier. If that's just with HEAD racket and typed racket, without this pull request, then that's the behavior I'd expect.

samth:
If you try it in the other direction (providing `Any -> Boolean` to untyped racket) then you should get the behavior you expect.

Benn:
Any chance i could be removed from this list?
Cheers
Benoit
Le 30 oct. 2015 15:56, "Sam Tobin-Hochstadt" <notifications@github.com> a
écrit :

rfindler:
Benoît: I think you need to do it yourself on github. I don't think we can
do it for you.

Sorry again.

bennn:
Oh wow: the updated contract system fixes my particular use case. Most of the slowdowns I'd seen were from structs imported with a `require/typed ... #:struct`. It turns out, `#:struct` explicitly gives the struct predicate an `(any/c . -> . boolean)` contract, without going through `type->contract`. So these are now optimized.

This does means you can sneak a chaperoned struct predicate through, so I'll refactor this PR to fix that, but I don't think there's any longer a strong performance reason to merge this in.

Example of an unsound struct predicate:
```
#lang racket/base

(module bad racket/base
  (struct s ())
  (define s*? (chaperone-procedure s? (lambda (x) (x 0) x)))
  (provide s struct:s (rename-out [s*? s?])))

(module user typed/racket/base
  (require/typed (submod ".." bad)
    (#:struct s ()))
  (s? string-append))

(require 'user) ;; string-append, contract violation
```

takikawa:
@bennn Doesn't that mean you can also sneak higher-order values through the predicate to untyped too?

samth:
@bennn You've discovered a bug, which we need to fix. That should generate the same contract as `(Any -> Boolean)`.

bennn:
That's great :+1:
Here's an Asumu-inspired exploit
```
#lang racket/base

(module bad racket/base
  (struct s ())
  (define (s*? x)
    (when (vector? x) (vector-set! x 0 'pwned))
    #t)
  (provide s struct:s (rename-out [s*? s?])))

(module user typed/racket/base
  (require/typed (submod ".." bad)
    (#:struct s ()))
  (define: v : (Vectorof Natural) (vector 0 1 2))
  (s? v)
  v)

(require 'user)
;; #t
;; '#(pwned 1 2)
```

samth:
See also #202.

samth:
This all looks good to me now. Can you address the last comment, and then we'll be good to merge.

bennn:
Addressed

samth:
Can you squash this into one commit for merging?

bennn:
Squashed, and only using `impersonator?`

```
> (define (f x) x)
> (define g (chaperone-procedure f (lambda (x) #t)))
> (impersonator? g)
#t
```

samth:
Great, I'll merge this imminently. Want to tackle #203 and #202 next?

bennn:
Yep! I'll try to open a pull request tomorrow with examples.

samth:
Merged.

PR #227, opened 2015-10-28
Make TR's contract generation interact better with contract system optimizations

takikawa:
This fixes a performance regression in TR from v6.1.1 to v6.2 that @rfindler discovered.

The regression is due to a combination of two problems. One is some overaggressive definition lifting by the contract generation pass. The other is due to the expansion of contract definition submodule redirection macros.

You can observe the problem by running `typed-racket-test/performance/function-contract.rkt`. In v6.1.1 it should take about 90ms. HEAD is about 270ms. This PR should bring it back down to around 90ms.

There's a problem though, which is that it fails two tests because of something weird with the top-level.

Here's a minimal interaction showing the problem:

````
-> (module a typed/racket
     (: x (Boxof Integer)) (define x (box 3))
     (provide x))
-> (require 'a)
-> (set-box! x "foo")
-> x
; x: broke its own contract
;   promised: Integer
;   produced: "foo"
;   in: (box/c Integer)
;   contract from: a
;   blaming: a
;    (assuming the contract is correct)
;   at: readline-input:3.14
; [,bt for context]
````

For some reason the `x` in the `set-box!` doesn't have a contract, but the `x` later does. This only seems to happen for the top-level, and it also doesn't seem to trigger if the contracted variable is used as the head form of an expression.

Any ideas? Is there something wrong with my use of `syntax-local-lift-require` in commit c26e5dc7199ed142e2ab85804fdee0324923557b?

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -122,16 +122,35 @@
                                  "type" #,(pretty-format-type type #:indent 8)))]
            [else
             (match-define (list defs ctc) result)
+            (define maybe-inline-val
+              (should-inline-contract? ctc cache))
             #`(begin #,@defs
-                     (define-values (ctc-id) #,ctc)
+                     #,@(if maybe-inline-val
+                            null
+                            (list #`(define-values (ctc-id) #,ctc)))
                      (define-module-boundary-contract #,untyped-id
-                       #,orig-id ctc-id
+                       #,orig-id
+                       #,(or maybe-inline-val #'ctc-id)
                        #:pos-source #,blame-id
                        #:srcloc (vector (quote #,(syntax-source orig-id))
                                         #,(syntax-line orig-id)
                                         #,(syntax-column orig-id)
                                         #,(syntax-position orig-id)
                                         #,(syntax-span orig-id))))])]))
+
+;; Syntax (Dict Static-Contract (Cons Id Syntax)) -> (Option Syntax)
+;; A helper for generate-contract-def/provide that helps inline contract
+;; expressions when needed to cooperate with the contract system's optimizations
+(define (should-inline-contract? ctc-stx cache)
+  (and (identifier? ctc-stx)
+       (let ([match? (assoc ctc-stx (hash-values cache) free-identifier=?)])
+         (and match?
+              ;; ->* are handled specially by the contract system
+              (let ([sexp (syntax-e (cdr match?))])
+                (and (pair? sexp)
+                     (or (free-identifier=? (car sexp) #'->)
+                         (free-identifier=? (car sexp) #'->*))))
+              (cdr match?)))))
 
 ;; The below requires are needed since they provide identifiers that
 ;; may appear in the residual program.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by rfindler** [comment]
Does this bad behavior happen if you use just the contract system equivalent of the TR program above?

**Comment by takikawa** [comment]
Here's an attempt to minimize this. I think contracts are actually not necessary to produce the bad behavior, just renaming things.

````racket
#lang racket/load

(module a racket
  (require (for-syntax syntax/private/modcollapse-noctc)
           (for-syntax racket/base)
           (for-template racket))

  (define b (box 3))

  (define-syntax (m stx)
    (syntax-case stx ()
      [(_ x redirect)
       (let ([x2 (syntax-local-introduce #'x)]
             [x3 (syntax-local-introduce #'x)])
         #`(begin
            (module* #%defs #f 
              (provide (rename-out [x #,x2])))
            (define-syntax #,x3 (redirect (quote-syntax #,x2)))
            (provide #,x3)))]))

  (begin-for-syntax
    (require racket/syntax
             racket/trace)
    
    (module #%defs-ref racket/base
      (require racket/runtime-path
               (for-syntax racket/base))
      (define-runtime-module-path-index defs-submod-idx
        '(submod ".." #%defs))
      (provide defs-submod-idx))

    (require (submod "." #%defs-ref))

    (define ((make-make-redirect-to-contract contract-defs-submod-modidx) id)
      (trace-define (redirect stx)
        (cond
         [(identifier? stx)
          (with-syntax ([mp (collapse-module-path-index
                             contract-defs-submod-modidx)]
                        [i (datum->syntax id (syntax-e id) stx stx)])
            (define x (generate-temporary))
            (syntax-local-lift-require
             #`(rename mp #,x #,(datum->syntax #'mp (syntax-e #'i)))
             x))]
         [else
          (datum->syntax stx
                         (cons (redirect (car (syntax-e stx)))
                               (cdr (syntax-e stx)))
                         stx
                         stx)]))
      redirect)

    (define redirect (make-make-redirect-to-contract defs-submod-idx)))

  (m b redirect))

(require 'a)

(unbox b)
````

This program behaves differently when it's in `#lang racket` rather than `#lang racket/load`.

@mflatt any thoughts on why this would behave differently in the top-level? (I can also try to minimize more if that would help)

**Comment by mflatt** [comment]
I've pushed an attempt to repair `syntax-local-lift-require`.

**Comment by takikawa** [comment]
@mflatt Thanks! Looks like that made all the tests pass.

**Comment by stamourv** [comment]
LGTM modulo comment.

File: typed-racket-lib/typed-racket/private/type-contract.rkt [hunk 1]
```diff
@@ -133,6 +138,19 @@
...
+;; A helper for generate-contract-def/provide that helps inline contract
+;; expressions when needed to cooperate with the contract system's optimizations
+(define (should-inline-contract? ctc-stx cache)
+  (and (identifier? ctc-stx)
+       (let ([match? (assoc ctc-stx (hash-values cache) free-identifier=?)])
+         (and match?
+              ;; ->* are handled specially by the contract system
+              (let ([sexp (syntax-e (cdr match?))])
+                (and (pair? sexp)
+                     (free-identifier=? (car sexp) #'->*)))
```
samth: Shouldn't we check for `->` too, in case we start generating those?

samth:
LGTM modulo that one comment.

Issue #223, opened 2015-10-23
(member) has wrong type, exploiting the hole causes segfault


jsmaniac:
`member`'s type allow passing Any as a query, but passing a function of type `(→ a a Any)`. As a result, the function can be called with something else than a.

When running the code below in command-line racket or in DrRacket, I get a segfault using 6.2.900.11 (I haven't tested with a newer version, but `member` still has the wrong type on HEAD` I believe).

```
SIGSEGV MAPERR si_code 1 fault on addr 0xff
Abandon (core dumped)
```

```
#lang typed/racket

(: equal-length (→ String String Boolean))
(define (equal-length s1 s2)
  (= (string-length s1) (string-length s2)))

;; Okay:
(member "x"
        '("bb" "c" "ddd")
        equal-length)

;; Segfault
(member 123
        '("bb" "c" "ddd")
        equal-length)
```

I think the type of member should be:

```
(∀ (a b) (case→ [→ Any
                     (Listof a)
                     (U False (Pairof a (Listof a)))]
                  [→ b
                     (Listof a)
                     (→ b a Any) ;; Unsure about this one
                     (U False (Pairof a (Listof a)))]))
```

However, I am unsure about the `(→ b a Any)`: Does `member` always pass the needle as the first argument, and an element of the haystack as the second? If not we should then put `(→ (U a b) (U a b) Any)`. If the first case is true for now, we should probably add a test to check for a change in that behaviour:

```
;; Random needle to prevent some of optimization (but not all):
(let ([needle (floor (* (random) 200))])
  (member needle
          (range 1000)
          (λ ([x : Integer] [y : Integer])
            ;; Check the needle is always the first argument
            (check-equal? x needle)
            ;; Check y = needle implies x = needle
            (check-true (or (not (= y needle)) (= x needle)))
            (= x y))))
```

jsmaniac:
I wrote a fix on my fork that also adds the equality-predicate argument to `(assoc …)`. I'm in the process of building it to check that it works.

https://github.com/jsmaniac/typed-racket/tree/fix-assoc-and-member-types

In which file should I add the tests for `(member …)` and `(assoc …)`?

samth:
Thanks for finding and fixing this! Tests should go in `typed-racket-test/unit-tests/typecheck-tests`.

jsmaniac:
Hi @samth,

[...]

PR #230, opened 2015-10-30
Fix types for assoc and member + tests for these

jsmaniac:
Fixes github issue #223: “(member) has wrong type, exploiting the hole causes segfault”

* Fixes the type declaration for `member` so that it has the correct type for the third `is-equal?` argument.
* Changes the type declaration for `assoc` so that it accepts a third `is-equal?` argument
* Adds some tests to check the validity of the hypothesis about `member` and `assoc` always giving the needle as the first argument to the provided `is-equal?` function.
* Adds some tests to check that the typechecker correctly rejects `(member 123 '("bb" "c" "ddd") equal-length)`, where equal-length takes two strings, and the same for assoc.

----

I think it would be nice to also test that, the infered type arguments for `member` and `assoc` are correct  in the following code, but I don't know how to test the infered type arguments with `tc-e` and friends.

* For `member`, `a = (U 'a 'b 'c)` and `b = (U Number Symbol String)`, and not some wider type for `b`.
* For `assoc`, `a = (U 'a 'b 'c)` and `b = (Vectorof Symbol)` (I'm unsure about the `b`) and `c = (U Number Symbol String)`, and not some wider type for `c`.

```racket
       [tc-e (member 3
                     '(a b c)
                     (lambda: ([s1 : (U Number Symbol String)] [s2 : String])
                       (= (string-length (format "~a" s1)) (string-length s2))))
             (t:Un (-val #f) (-lst (one-of/c 'a 'b 'c)))]
       [tc-e (assoc 3
                    '((a . #(a)) (b . #(b)) (c . #(c)))
                    (lambda: ([s1 : (U Number Symbol String)] [s2 : String])
                      (= (string-length (format "~a" s1)) (string-length s2))))
             (t:Un (-val #f) (-pair (one-of/c 'a 'b 'c) (-vec -Symbol)))]
```

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -640,7 +640,10 @@
 
 [assq  (-poly (a b) (Univ (-lst (-pair a b)) . -> . (-opt (-pair a b))))]
 [assv  (-poly (a b) (Univ (-lst (-pair a b)) . -> . (-opt (-pair a b))))]
-[assoc (-poly (a b) (Univ (-lst (-pair a b)) . -> . (-opt (-pair a b))))]
+[assoc (-poly (a b c)
+              (cl->* (Univ (-lst (-pair a b)) . -> . (-opt (-pair a b)))
+                     (c (-lst (-pair a b)) (-> c a Univ)
+                        . -> . (-opt (-pair a b)))))]
 [assf  (-poly (a b) ((a . -> . Univ) (-lst (-pair a b))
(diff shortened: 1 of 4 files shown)


samth:
`tc-e` actually tests that the result isn't a wider type. You can see this in the currently failing tests for this PR.

Can you squash this all into one commit?

samth:
@jsmaniac just so you know, you don't have to close this. If you just force-push to the same branch, github will update the pull request.

jsmaniac:
@samth I guessed that would work, but I tend to avoid force-push it in case anybody (including myself) cloned a copy.

PR #232, opened 2015-11-02
Fix midi-readwrite.

samth:
Fix compilation of `midi-readwrite`. Review by @stamourv requested.

### DIFF
--- a/typed-racket-lib/typed-racket/rep/rep-utils.rkt
+++ b/typed-racket-lib/typed-racket/rep/rep-utils.rkt
@@ -362,8 +362,8 @@
                 [PathElem def-pathelem #:PathElem pathelem-case print-pathelem pathelem-name-ht pathelem-rec-id])
 
 ;; NOTE: change these if the definitions above change, or everything will segfault
-(define-syntax-rule (unsafe-Rep-seq v) (unsafe-struct*-ref v 0))
-(define-syntax-rule (unsafe-Type-key v) (unsafe-struct*-ref v 4))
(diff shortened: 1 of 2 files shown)


stamourv:
LGTM.
Are there other unsafe operations we should get rid of?

samth:
These are the ones that seem to get us in trouble. But I'm particularly interested in your feelings on the last commit, which seems like it's papering over the real problem.

stamourv:
The last commit is fine as a stopgap, but we definitely want to investigate how that `#f` got there.
I assume the test case you added triggers it reliably? If so, I'll look into it.

samth:
`raco pkg install midi-readwrite` :)

stamourv:
Oh, and re "those are the ones that get us in trouble": For now. I think we should reconsider others as well.

samth:
I just skimmed the repo, and I think everything else is either:
* handling some function named `unsafe-...`
* generating unsafe code in the optimizer
* guarded by an immediate test (there was only 1 of these)

stamourv:
Ok, then I guess that was the only one we really ought to remove.

stamourv:
For the sake of having two pairs of eyes on it, I've also done an audit of our uses of `unsafe` ops.
I found one that is not guaranteed to be safe, in the expansion of TR's `for/vector`.
I also found two in the implementation of `seen?`, in `types/current-seen.rkt`. I'm not convinced that they are adequately guarded, so I've added an additional check.
I'll push those changes.

Issue #234, opened 2015-11-05
Does the Unit type need to be provided by typed/racket/base?


AlexKnauth:
Should it only be provided by typed/racket?

I'm asking because I'm getting name collisions even when using base, where I only expected to have collisions with typed/racket.

samth:
I agree, it should not be provided by `typed/racket/base`.

PR #236, opened 2015-11-05
Don't provide Unit from typed/racket/base

AlexKnauth:
fixes https://github.com/racket/typed-racket/issues/234

### DIFF
--- a/typed-racket-lib/typed/racket/base.rkt
+++ b/typed-racket-lib/typed/racket/base.rkt
@@ -19,7 +19,7 @@
            require/typed-legacy
            require-typed-signature)
          typed-racket/base-env/base-types
-         (except-in typed-racket/base-env/base-types-extra Distinction))
+         (except-in typed-racket/base-env/base-types-extra Distinction Unit))
 (provide (rename-out [define-type-alias define-type])
          (all-from-out typed-racket/base-env/prims)
          (all-from-out typed-racket/base-env/base-types)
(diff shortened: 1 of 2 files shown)


samth:
Looks good once the tests pass.

AlexKnauth:
Should this also be added to the release branch?

samth:
I'd prefer not to, since we've basically frozen things. What do other people think, @stamourv @takikawa?

AlexKnauth:
But then the name collision would be there in the release, and code could end up relying on it being provided by base, right?

samth:
@AlexKnauth that's a good point, but the name collision is there in `typed/racket` already.

AlexKnauth:
Should I merge now? (to master)

stamourv:
Aren't we due for another candidate / round of TR testing anyway? In which case, we may be able to bring that change in.

samth:
Yes, merging to master is definitely right, regardless of the release issue.

AlexKnauth:
If you're doing another round of testing anyway, it sounds like it could be added and tested properly with that.

rfindler:
We find it difficult to get everyone to run their tests the first time and
since most of them are run now, it is unlikely we'll get them run a second
time.

Can you say more about why this change is important? Is there a problem
that's new to this release or is the problem already present in the last
release?

samth:
This is a problem new to this release. `typed/racket/base` provides a binding it shouldn't, `Unit`, and fixing it later would be backwards incompatible.

rfindler:
Sounds like something that we should consider as a change to the release,
then.

Robby

samth:
I have merged these commits into the `release` branch.

PR #238, opened 2015-11-08
Handle `in-vector` with range arguments.

samth:
* Add `normalise-inputs` to special function env.
* Treat eta-expansion specially. Now
   `(lambda (x ...) (f x ...))`
 will typecheck like `f` but with a type restricted to
 the size of `x ...`.

  Currently, this special case only works for non-polymorphic
 functions.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
@@ -185,7 +185,16 @@
   (for/list ([arg-types (in-list new-arg-types)])
     (tc-lambda-body arg-list arg-types body)))
 
-
+;; restrict-to-arity : arr? nat -> (or/c #f arr?)
+;; either produces a new arity which is a subtype of arr with arity n,
+;; or #f is that is not possible
+(define (restrict-to-arity arr n)
+  (match arr
+    ;; currently does not handle rest arguments
+    [(arr: args ret #f #f '())
+     #:when (= n (length args))
+     arr]
+    [_ #f]))
 
 ;; formals syntax -> listof[arr?]
 (define (tc/lambda-clause formals body)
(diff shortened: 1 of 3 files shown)


typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -185,7 +185,16 @@
   (for/list ([arg-types (in-list new-arg-types)])
     (tc-lambda-body arg-list arg-types body)))
 
-
+;; restrict-to-arity : arr? nat -> (or/c #f arr?)
+;; either produces a new arity which a subtype of the current one with arity n,
```
stamourv: "which *is* a subtype"

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -196,11 +205,21 @@
 
   (define arg-list (formals-positional formals))
   (define rest-id (formals-rest formals))
+
+  (define eta-expand?
```
stamourv: The name makes it sound like we're trying to decide whether we should eta expand, whereas it looks more like trying to identify whether the code is an eta expansion.

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -217,11 +236,31 @@
...
+               (define possibles (for*/list ([arr (in-list arrs)]
+                                             [restricted (in-value (restrict-to-arity arr (length arg-list)))]
+                                             #:when restricted)
+                                   restricted))
+               (if (null? possibles)
+                   #f
+                   possibles)]))
+       =>
+       (lambda (x)
+         (register-ignored! (car (syntax-e body)))
```
stamourv: Why ignore the body? Don't you have all the info required to add types to the type table?

stamourv:
LGTM modulo comments.

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
samth: I accidentally a word.

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
samth: How about `eta-expanded?`?

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
samth: What types should I add to the type table? Note also that the function could have a `case->` type.

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
stamourv: Sure.

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
stamourv: It should be ok to give the restricted type to the function, no? I was initially worried about that case, but couldn't think of a concrete case where that would be wrong.

samth: Right, but what type do I give to the arguments?

samth:
Merged.

PR #240, opened 2015-11-09
Fix types for `member` and `assoc` + tests for these

jsmaniac:
Fixes github issue #223: “(member) has wrong type, exploiting the hole causes segfault”

*Done*

* Fixes the type declaration for `member` so that it has the correct type for the third `is-equal?` argument.
* Changes the type declaration for `assoc` so that it accepts a third `is-equal?` argument
* Adds some tests to check the validity of the hypothesis about `member` and `assoc` always giving the needle as the first argument to the provided `is-equal?` function.
* Adds some tests to check that the typechecker correctly rejects `(member 123 '("bb" "c" "ddd") equal-length)`, where equal-length takes two strings, and the same for assoc.

----

*Still to be done*

I think it would be nice to also test that, the inferred type arguments for `member` and `assoc` are correct  in the following code, but I don't know how to test the inferred type arguments with `tc-e` and friends.

* The type parameter `a` for `member` should be `(U 'x 'y 'z)` and the type parameter `b` should be `(U Number Symbol String)`, and the type parameter `b` should not be of some wider type.
* The type parameter `a` for `assoc` should be `(U 'x 'y 'z)` and the type parameter `b` should be `(Vector Symbol)` and the type parameter `c` should be `(U Number Symbol String)`, and the type parameter `c` should not be of some wider type.

```racket
       [tc-e (member 3
                     '(x y z)
                     (lambda: ([s1 : (U Number Symbol String)] [s2 : String])
                       (= (string-length (format "~a" s1)) (string-length s2))))
             (t:Un (-val #f) (-lst (one-of/c 'x 'y 'z)))]
       [tc-e (assoc 3
                    '((x . #(x)) (y . #(y)) (z . #(z)))
                    (lambda: ([s1 : (U Number Symbol String)] [s2 : String])
                      (= (string-length (format "~a" s1)) (string-length s2))))
             (t:Un (-val #f) (-pair (one-of/c 'x 'y 'z) (-vec* -Symbol)))]
```

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -632,15 +632,18 @@
 [memq (-poly (a) (-> Univ (-lst a) (-opt (-ne-lst a))))]
 [memv (-poly (a) (-> Univ (-lst a) (-opt (-ne-lst a))))]
 [memf (-poly (a) ((a . -> . Univ) (-lst a) . -> . (-opt (-ne-lst a))))]
-[member (-poly (a)
+[member (-poly (a b)
           (cl->* (Univ (-lst a) . -> . (-opt (-ne-lst a)))
-                 (Univ (-lst a) (-> a a Univ)
+                 (b (-lst a) (-> b a Univ)
                        . -> . (-opt (-ne-lst a)))))]
 [findf (-poly (a) ((a . -> . B) (-lst a) . -> . (-opt a)))]
 
 [assq  (-poly (a b) (Univ (-lst (-pair a b)) . -> . (-opt (-pair a b))))]
 [assv  (-poly (a b) (Univ (-lst (-pair a b)) . -> . (-opt (-pair a b))))]
-[assoc (-poly (a b) (Univ (-lst (-pair a b)) . -> . (-opt (-pair a b))))]
+[assoc (-poly (a b c)
+              (cl->* (Univ (-lst (-pair a b)) . -> . (-opt (-pair a b)))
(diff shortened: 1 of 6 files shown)


samth:
Those two tests already test that -- they will fail if the result is a wider type.

jsmaniac:
@samth I'm concerned about the type parameters I added to `member` and `assoc`, respectively `b` and `c` (see [the diff](https://github.com/racket/typed-racket/pull/240/files#diff-ef79b6b89ee78c4b6ada735b8e32a899)), which are not present in the result type of the function, but instead only used for the `needle` and `is-equal?` arguments.

Since they are not used in the result type, I think these tests would still succeed if they are widened?

samth:
@jsmaniac if they aren't in the result, then it won't check, but if that happened, what would be the problem? Is there a bug you can think of that you're trying to prevent?

jsmaniac:
@samth I can only think of the case where the type gets widened to `Any` (for example), and the lambda, with type `(→ (U Number Symbol String) String Any)` gets rejected, because it would expect `(→ Any String Any)`. This hypothetical bug sounds unlikely to occur without breaking lots of other things, though, so it's okay like this.

By the way, the travis-ci build just failed because of a timeout, but the code is okay, it should be identical (I made a diff to check) to the code in this build, which passed about an hour ago https://travis-ci.org/jsmaniac/typed-racket/builds/90144355 .

jsmaniac:
I'll add a couple of negative tests in fail/, amend the commit and force-push after checking that it works, that should re-run the build.

samth:
Looks good. I'll merge it.

samth:
Merged, thanks!

PR #241, opened 2015-11-11
guard #:opaque predicates

### LINKED ISSUE #202: #:opaque fails to check inputs to predicate, is unsound

bennn:
Proposal to fix #202 , but not #203 .
Guards opaque predicates with an `(-> Any Any)` contract, but allows struct predicates as a special case.


samth:
Why do you need to generate that require?

bennn:
I get a "link mismatch" otherwise... I must be doing something wrong because `#:struct` doesn't have the same problem, but I couldn't figure out the difference.

```
link: namespace mismatch;
 reference to a module that is not available
  reference phase: 0
  referenced module: "/home/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-lib/typed-racket/utils/any-wrap.rkt"
  referenced phase level: 0
  reference in module: (submod "/home/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-test/succeed/pr242-variation-2.rkt" typed)
  in: struct-predicate-procedure?/c
  context...:
   (submod /home/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-test/succeed/pr242-variation-2.rkt typed): [running body]
   /home/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-test/succeed/pr242-variation-2.rkt: [traversing imports]
```

samth:
Can you give an example where this happens? I tried taking it out and didn't see an error.

bennn:
Just updated, I still have the error

1. Edited `base-env/prims-contract.rkt`
2. Compiled `raco make core.rkt`
3. Changed to `typed-racket-test/succeed`
4. Tried running `raco make pr241-variation-0.rkt`

Namespace mismatch on the bindings from `any-wrap.rkt`.

samth:
Ah, actually it's variations 1 and 2 that provoke the error. :)

bennn:
See #248

Issue #214, opened 2015-10-15
Error during contract generation for typed racket modules


akuhlens:
Sorry to bring this error up again, but I believe that I had been told that it was fixed.

The following code...
```racket
#lang typed/racket

(struct (A B) Fum ([a : A] [b : B]))
(struct Fi ())
(struct Foo ())

(define-type Tail
  (Rec T
   (U (Fum (Listof Value) T)
       Fi)))

(define-type Value
 (Rec V
  (U (Fum (Listof Value) V)
     Foo)))

(provide fun1)

(: fun1 (-> Tail))
(define (fun1)
  (error 'foo1))
```
gets this error message

```bash
raco make src/data/racket-error.rkt
"src/data/racket-error.rkt":
  making #<path:/Users/akuhlens/work/gradual/schml/src/data/racket-error.rkt>
hash-ref: no value found for key
  key: #<syntax n*13>
  compilation context...:
   /Users/akuhlens/work/gradual/schml/src/data/racket-error.rkt
  context...:
   loop
   /Users/akuhlens/Desktop/Racket v6.2.900.17/share/pkgs/typed-racket-lib/typed-racket/static-contracts/instantiate.rkt:88:10
   /Users/akuhlens/Desktop/Racket v6.2.900.17/share/pkgs/typed-racket-lib/typed-racket/static-contracts/equations.rkt:47:6: for-loop
   /Users/akuhlens/Desktop/Racket v6.2.900.17/share/pkgs/typed-racket-lib/typed-racket/static-contracts/equations.rkt:45:4: loop
   /Users/akuhlens/Desktop/Racket v6.2.900.17/share/pkgs/typed-racket-lib/typed-racket/static-contracts/equations.rkt:42:0: resolve-equations
   /Users/akuhlens/Desktop/Racket v6.2.900.17/share/pkgs/typed-racket-lib/typed-racket/static-contracts/instantiate.rkt:74:0: compute-recursive-kinds
   .../more-scheme.rkt:261:28
   /Users/akuhlens/Desktop/Racket v6.2.900.17/share/pkgs/typed-racket-lib/typed-racket/private/type-contract.rkt:100:0: generate-contract-def/provide
   /Users/akuhlens/Desktop/Racket v6.2.900.17/share/pkgs/typed-racket-lib/typed-racket/private/type-contract.rkt:173:3: for-loop
[...]

PR #242, opened 2015-11-11
Partial fix to SC instantiation

stamourv:
Fixes the example from #214, but breaks other tests.
I don't have a good intuition for the problem (or for anything SC-related, really).
I plan to continue looking into it (though maybe not right now), but if anyone has thoughts.


takikawa:
Just a guess at why it might be failing: calling `sc-traverse` on the static contract may not be sufficient to find all the referenced Name types because when you traverse a name/sc it's a no-op (to avoid dereferencing the name, which might be problematic). You may have to manually lookup the name and recur on the mapped sc, being careful not to recur forever.

stamourv:
Ah, that sounds plausible. I'll look into that. Thanks!

stamourv:
Ding ding ding, we have a winner!

PR #243, opened 2015-11-12
added missing optional checked aurgument to Checkable-Menu-Item% in g…

earldean:
Checkable-Menu-Item% type was missing optional checked argument.


stamourv:
LGTM. Do you have push access?

PR #248, opened 2015-11-16
Revise #:opaque changes

samth:


typed-racket-test/succeed/pr241-variation-0.rkt
```diff
@@ -0,0 +1,17 @@
...
+ (struct s ())
+ (define val (s))
+ (provide val (struct-out s)))
+
+(module typed typed/racket/base
+ (require/typed (submod ".." untyped)
+   [#:opaque S s?]
+   [val S])
+ (s? 4)
+ (s? val))
```
bennn: We should also assert that `s?` has no contract.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -324,20 +325,25 @@
...
     (pattern #:name-exists))
   (syntax-parse stx
+    [_ #:when (eq? 'module-begin (syntax-local-context))
+       ;; it would be inconvenient to find the correct #%module-begin here, so we rely on splicing
+       #`(begin #,stx (begin))]
     [(_ ty:id pred:id lib (~optional ne:name-exists-kw) ...)
      (with-syntax ([hidden (generate-temporary #'pred)])
+       (define pred-cnt
+         (syntax-local-lift-expression
+          (make-contract-def-rhs #'(-> Any Boolean) #f #f)))
```
bennn: Shouldn't this be `(-> Any Any)`, so that `variation-2` below passes without the `(and ... #t)` change?

samth: No, because we say that the type of the predicate is `(Any -> Boolean)`.  The generated contract and the type need to be consistent.

Changing the type to be `(-> Any Any)` would break programs like this:

```
(require/typed m [#:opaque T t?])
(define x : Boolean (t? #t))
```

Also, keeping the more restrictive type can't break existing programs because we previously used `boolean?` in the contract.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -324,20 +325,25 @@
   (define-syntax-class name-exists-kw
     (pattern #:name-exists))
   (syntax-parse stx
+    [_ #:when (eq? 'module-begin (syntax-local-context))
+       ;; it would be inconvenient to find the correct #%module-begin here, so we rely on splicing
+       #`(begin #,stx (begin))]
```
bennn: This must be solving the import problem from #241 .... how does this work?

samth: No, this solves a different problem (which is what happens if `require/opaque-type` is the only thing in the module). The import problem is solved by the contract generation infrastructure already doing the right thing and inserting the same `require` manually into the `contract-defs` submodule.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
bennn: Okay then, line 342 should say `(-> Any Boolean)`, right?

samth: Yes, definitely.

bennn:
I'm not sure what to think about this -- here's a program that compiles on head, but not on this PR (because `foo?` is not a struct predicate procedure).

```
#lang racket/base

(module u racket/base
  (require racket/generic)
  (define-generics foo)
  (struct foo-struct () #:methods gen:foo [])
  (define f1 (foo-struct))
  (provide f1 foo?))

(module t typed/racket/base
  (require/typed (submod ".." u)
    (#:opaque Foo foo?)
    (f1 Foo))
  (foo? 3)
  (foo? f1))

(require 't)
```

samth:
Right, that's #203.

samth:
That program should work with that additional commit.

bennn:
No, it still fails. So does the example in #203.

(I'm on commit `84db203` and I cleared all the compiled folders in my `typed-racket-lib`.)

samth:
Ah, the problem is this line: https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/utils/any-wrap.rkt#L113

The issue is we have no idea what the value is, and thus we can't be sure that passing it to untyped code is safe. So it could be an opaque struct, in which case the argument I gave in #203 applies, but it also could be some other type that Typed Racket supports that's missing from the enumeration in that function. So we have to make sure that function is complete before removing the error case.

samth:
Ok, so next steps are:

- enumerate all the base types in TR and make sure they're handled in that code
- remove the final error case

I worry that this will not be safe in general, though, but I haven't found a counterexample yet.

bennn:
What if we use `~a` to approximate whether something is an opaque struct?
This would fail if the struct had a custom writer, but we could keep the final error case.

```
(struct s ())
(define s1 (s))
(regexp-match? #rx"^#<.*>$" (~a s1)) ;; Is there a ~write ?
```

samth:
No -- printing things is not an ok way to tell things about them.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -324,20 +325,25 @@
...
+          (make-contract-def-rhs #'(-> Any Boolean) #f #f)))
        (quasisyntax/loc stx
          (begin
            ;; register the identifier for the top-level (see require/typed)
            #,@(if (eq? (syntax-local-context) 'top-level)
                   (list #'(define-syntaxes (hidden) (values)))
                   null)
-           #,(ignore #'(define pred-cnt (any/c . c-> . boolean?)))
-           #,(internal #'(require/typed-internal hidden (Any -> Boolean : (Opaque pred))))
+           #,(internal #'(require/typed-internal hidden (Any -> Any : (Opaque pred))))
```
takikawa: Looks like this line should have the type from above?

PR #250, opened 2015-11-19
Improving error messages, by providing more context.

jsmaniac:
This commit improves error messages given by typed/racket, by always including the type name and arguments in the error message.

When a typing error is located inside macro-expanded code, a message such as “wrong number of arguments to polymorphic type: expected 1 and got 2” does not help much in locating which instantiation is wrong.

As far as I can tell the tests didn't need to be updated to match the new error messages. If you think other messages should be improved, please tell me and I'll take care of them.


samth:
What do the new messages look like with this change?

Also, should we change the tests to make sure that the new information is included?

jsmaniac:
I think no tests need to be updated. Should I add tests to check that the error messages contain the desired information? I'm not sure if it's a good idea to make too specific tests though.

I added "arguments were: (A1 A2)" to the error messages where this information wasn't present. Here are the messages I changed, new versions:

* wrong number of arguments to polymorphic type T: expected 1 and got 2, arguments were: (A1 A2)`
* The expected number of arguments for structure type constructor T does not match the given number: expected 1, given 2, arguments were: (A1 A2)`
* The expected number of arguments for T does not match the given number: expected 2, given 1, arguments were: (A1 A2)`
* Recursive type T cannot be applied at a different type in its recursive invocation, new arguments were: (A1 A2)`
* cannot apply a non-polymorphic type: T with arguments: (A1 A2)`

The old versions were:

* `wrong number of arguments to polymorphic type T: expected 1 and got 2`
* `The expected number of arguments for structure type constructor T does not match the given number of arguments: expected 1, given 2`
* `The expected number of arguments for T does not match the given number: expected 2, given 1`
* `Recursive type T cannot be applied at a different type in its recursive invocation`
* `cannot apply a non-polymorphic type: T`

jsmaniac:
Should I also take this as an opportunity to homogenize the error messages, and make them all of the format:
* `wrong number of arguments to …: expected … and got …, arguments were: …`
* instead of `The expected number of arguments for … does not match the given number: expected …, given …, arguments were: …`
I find the former clearer and it would be more consistent, but it's really not worth fighting an argument (pun indended).

samth:
Sorry for taking a while to review this. I think we want to change how you're printing the other arguments, to look more like the way `raise-argument-error` works: http://docs.racket-lang.org/reference/exns.html?q=raise-arg#%28def._%28%28quote._~23~25kernel%29._raise-argument-error%29%29

Otherwise, this looks good.

jsmaniac:
Thanks for the feedback Sam. I'll be away from keyboard for a couple of
days, but I'll take care of these changes soon.

Le dimanche 22 novembre 2015, Sam Tobin-Hochstadt <notifications@github.com>
a écrit :
> Sorry for taking a while to review this. I think we want to change how
you're printing the other arguments, to look more like the way
raise-argument-error works:
http://docs.racket-lang.org/reference/exns.html?q=raise-arg#%28def._%28%28quote._~23~25kernel%29._raise-argument-error%29%29
>
> Otherwise, this looks good.
>
> —
> Reply to this email directly or view it on GitHub.<
https://ci5.googleusercontent.com/proxy/oP9DKW_egWrGkkpcgnKZ0AvHQO6DEbThH3Tka6FbclBJpLcWnBxj93ep2Fb3AevY3GhmcnHCNwRYL-K2Qm-F2fwUrhrH2uhhXcBkiArAgvbDwnjnQr_bsM0H8cCuSqZzcobVr0vbRmHrde3Nc6YpFQeML75TYQ=s0-d-e1-ft#https://github.com/notifications/beacon/AASTYL5njdRBnlRo4PHojZ6XbB8yzLUVks5pIgXWgaJpZM4GleOX.gif
>

-- 
Georges Dupéron
Doctorant en Informatique à l'UM2

PR #256, opened 2015-11-24
Remove many uses of dict functions for performance

takikawa:
This seems to cut about 10% of the runtime of new-metrics.rkt and about 4% of the compile-time of the math library. It doesn't get rid of all uses of dict functions because some are difficult to remove due to actual genericity (and I also didn't bother with most of the uses in class-related code yet because those aren't exercised by most programs).

Can someone else check if they get similar speedups?

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/signature-prims.rkt
+++ b/typed-racket-lib/typed-racket/base-env/signature-prims.rkt
@@ -14,7 +14,6 @@
                      "../private/syntax-properties.rkt"
                      (typecheck internal-forms)
                      syntax/id-table
-                     racket/dict
                      racket/unit-exptime
                      (utils tc-utils))
          (only-in racket/unit
--- a/typed-racket-lib/typed-racket/base-env/unit-prims.rkt
+++ b/typed-racket-lib/typed-racket/base-env/unit-prims.rkt
@@ -30,7 +30,6 @@
                       "../private/syntax-properties.rkt"
                       (typecheck internal-forms)
                       syntax/id-table
-                      racket/dict
                       racket/unit-exptime
                       syntax/strip-context
                       (utils tc-utils)
--- a/typed-racket-lib/typed-racket/env/init-envs.rkt
+++ b/typed-racket-lib/typed-racket/env/init-envs.rkt
@@ -13,7 +13,7 @@
          (rep type-rep object-rep filter-rep rep-utils free-variance)
          (for-syntax syntax/parse racket/base)
          (types abbrev union)
-         racket/list racket/promise
+         racket/dict racket/list racket/promise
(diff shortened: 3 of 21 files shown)


andmkent:
With these changes I got about 3% better performance on the math lib and 4% better performance on new-metrics.rkt

stamourv:
Specializing to hash and id-table operations is fine to me, but I'm less convinced by the alist op changes. I'm not a fan of all those `match-define`s to take pairs apart, and `assoc` is stone age technology which makes the code less clear.

If we can get decent speedups with just the former, that sounds good. Otherwise, I don't know whether the speedups @andmkent reports justify the readability loss.

takikawa:
I bet the `assoc`s don't have much performance value, so I can take them out and re-time. Or in the case of the type variable environment, it might be better to switch to a hashtable.

stamourv:
Would it? How big are those, typically?

andmkent:
Is using ```assoc``` as a lookup any more unclear than using ```cons``` to add key/value pairs to a dictionary-like data structure?? =)

(begin-rant
I confess, I think deep down inside I really _hate_ uses of cons-based data-structures in non-trivial code bases when there are well established alternatives a few keystrokes away)

takikawa:
Reverting the assoc part doesn't seem to affect time of the math build.

stamourv:
@andmkent: That's an even better example, I agree.
Cons-based data structures are not the problem, IMO. It's that the interface exposes them. If we had `alist-add`, `alist-set`, and `alist-ref`, things would be fine.

samth:
What about just changing all the `racket/dict` requires to `racket/private/dict` (which has no contracts)? My microbenchmarks show that it gets rid of a lot of the overhead.

takikawa:
@samth I'd rather avoid relying on more private modules. Maybe racket/dict could provide a documented non-contracted version? (or maybe those contracts can be removed...)

samth:
I mostly agree with @stamourv, but I'm ok either way and we should just merge this.

rfindler:
It may now be that the contract on `dict-ref`'s contract isn't a bottleneck anymore because of this commit: https://github.com/racket/racket/commit/7056cd5f2abfbf9d6ed75d5ba98789a6a0e92cc2

samth:
@takikawa can you look into reviving this?

takikawa:
Yes, though the reason I didn't push it very much is it was unclear that the meager performance gain is worth the refactoring. I'll rebase and see how it looks though.

samth:
@pnwamk has this been obseleted by your changes?

pnwamk:
This PR is not obsolete -- I'll rebase it onto HEAD now

pnwamk:
see https://github.com/racket/typed-racket/pull/484

PR #257, opened 2015-11-28
removed scope structs from types

andmkent:
scope structs are cute, but don't do anything useful.


takikawa:
What was the purpose of these initially? Was it more useful before?

andmkent:
I do not believe so. I tore these out of the refinement types prototype a long time ago, IIRC after talking w/ @samth about them and concluding they didn't actually *do* anything and were just there to help illustrate the idea of what was going on in the locally nameless approach (although IMHO they're more confusing than helpful in that regard).

takikawa:
LGTM then.

PR #259, opened 2015-12-01
Clarify define-new-subtype docs

AlexKnauth:
Clarifies that `define-new-subtype` defines a type-level distinction that is not there at runtime, and that this means `make-predicate` can't distinguish the new type from the base type.


AlexKnauth:
?

samth:
Looks good.

PR #263, opened 2015-12-07
Add support for #:constructor-name and #:extra-constructor-name to struct

### LINKED ISSUE #253: `struct` should support `#:constructor-name`
See https://groups.google.com/forum/#!topic/racket-users/_UhPzbKETTw

takikawa:
Fixes #253.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims-struct.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-struct.rkt
@@ -72,10 +72,12 @@
 
   (define-splicing-syntax-class struct-options
     #:description "typed structure type options"
-    #:attributes (guard mutable? transparent? prefab? [prop 1] [prop-val 1])
+    #:attributes (guard mutable? transparent? prefab? ecname
+                  [prop 1] [prop-val 1])
     (pattern (~seq (~or (~optional (~seq (~and #:mutable mutable?)))
                         (~optional (~seq (~and #:transparent transparent?)))
                         (~optional (~seq (~and #:prefab prefab?)))
+                        (~optional (~seq #:extra-constructor-name ecname))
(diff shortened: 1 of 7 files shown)


typed-racket-lib/typed-racket/base-env/prims-struct.rkt
```diff
@@ -121,38 +126,48 @@
...
-                                               . opts)))]
-                        [dtsi (quasisyntax/loc stx
-                                (dtsi* (vars.vars ...)
-                                       nm.old-spec (fs.form ...)
-                                       #,@mutable?
-                                       #,@prefab?))])
-            #'(begin d-s dtsi)))]))))
+(define-syntax (define-typed-struct stx)
+  (syntax-parse stx
+    [(_ vars:maybe-type-vars nm:struct-name fs opts:struct-options)
```
stamourv: Should you check the well-formedness of the fields? You check everything else.

stamourv:
Could you add a test to check that you can't use both `#:constructor-name` and `#:extra-constructor-name`? Otherwise, LGTM modulo comment.

typed-racket-lib/typed-racket/base-env/prims-struct.rkt
takikawa: The other three are checked only because they are splicing syntax classes, so they have to be there. I could also check the fields, but `-struct` will also do that.

stamourv: Gotcha. Would checking earlier give better error messages, though? Erroring in terms of `-struct` wouldn't be great.

takikawa: I'll try it and if the error message is better I'll go with that.

PR #267, opened 2015-12-16
default implementation for object/c #:stronger

bennn:
Fixes the included microbenchmark. This does not seem to fix the wrappers from `fsmoo` and `forth`; those contracts must not be `eq?`.

### DIFF
--- a/typed-racket-lib/typed-racket/utils/opaque-object.rkt
+++ b/typed-racket-lib/typed-racket/utils/opaque-object.rkt
@@ -81,7 +81,7 @@
 (begin-for-syntax
  (define-syntax-class object/c-clause
    #:attributes (method-names method-ctcs field-names field-ctcs)
-   (pattern (field [name:id ctc:expr] ...)
+   (pattern ((~literal field) [name:id ctc:expr] ...)
             #:with field-names #'(list (quote name) ...)
             #:with field-ctcs #'(list ctc ...)
             #:with method-names #'null
--- a/typed-racket-lib/typed-racket/utils/opaque-object.rkt
+++ b/typed-racket-lib/typed-racket/utils/opaque-object.rkt
@@ -27,6 +27,9 @@
 ;; blame parties (e.g., a particular module).
 
 (require racket/class
+         (only-in racket/private/class-c-old
+                  base-object/c? build-object/c-type-name object/c-width-subtype?
+                  object/c-common-methods-stronger? object/c-common-fields-stronger?)
          racket/match
          racket/contract/base
          racket/contract/combinator
@@ -57,21 +60,55 @@
       (dynamic-object/c (append methods remaining-methods)
                         (append method-ctcs
                                 (for/list ([m remaining-methods])
-                                  (restrict-typed->/c)))
+                                  (restrict-typed->/c m)))
                         (append fields remaining-fields)
                         (append field-ctcs
                                 (for/list ([m remaining-fields])
-                                  (restrict-typed-field/c obj m)))))
+                                  (restrict-typed-field/c m)))))
     ;; FIXME: this is a bit sketchy because we have to construct
     ;;        a contract that depends on the actual object that we got
     ;;        since we don't know its methods beforehand
     (((contract-late-neg-projection guard/c) blame) obj neg-party)))
 
+(define (object/c-opaque-name ctc)
+  (build-object/c-type-name 'object/c-opaque
+                            (base-object/c-opaque-method-names ctc)
+                            (base-object/c-opaque-method-ctcs ctc)
+                            (base-object/c-opaque-field-names ctc)
+                            (base-object/c-opaque-field-ctcs ctc)))
+
+;; Similar to object/c-stronger, but without width subtyping.
+;; (Intuition: unspecified fields are guarded by the strongest possible contract)
+;; An opaque object contract `this` is stronger than `that` when:
+;; - `that` is an opaque contract
+;;   and `this` specifies at most the same members as `that`
+;;   and `this` has stronger contracts on all members
+;; - `that` is an object/c contract
+;;   and `this` has stronger contracts on their common members
+(define (object/c-opaque-stronger this that)
+  (define that-opaque? (base-object/c-opaque? that))
+  (cond
+   [(or that-opaque?
+        (base-object/c? that))
+    (define this-ctc (base-object/c-opaque-obj/c this))
+    (define that-ctc (if that-opaque? (base-object/c-opaque-obj/c that) that))
+    (and
+      (if that-opaque?
+        ;; then members of `this` should be a SUBSET of members of `that`
+        (object/c-width-subtype? that-ctc this-ctc)
+        #t)
+      (object/c-common-fields-stronger? this-ctc that-ctc)
+      (object/c-common-methods-stronger? this-ctc that-ctc)
+      #t)]
+   [else #f]))
+
 (struct base-object/c-opaque
-  (obj/c ; keep a copy of the normal object/c for first-order checks
+  (obj/c ; keep a copy of the normal object/c for first-order and stronger checks
    method-names method-ctcs field-names field-ctcs)
   #:property prop:contract
   (build-contract-property
+   #:stronger object/c-opaque-stronger
+   #:name object/c-opaque-name
    #:first-order (λ (ctc)
                    (define obj/c (base-object/c-opaque-obj/c ctc))
                    (λ (val)
@@ -126,10 +163,16 @@
                                                 "cannot call uncontracted typed method"))
                            (apply values args))))))
 
-(struct restrict-typed->/c ()
+;; Returns original method name
+(define (restrict-typed->-name ctc)
+  (define name (restrict-typed->/c-name ctc))
+  (build-compound-type-name 'restrict-typed->/c name))
+
+(struct restrict-typed->/c (name)
         #:property prop:chaperone-contract
         (build-chaperone-contract-property
-         #:name (λ (ctc) '<hidden-method>) ; FIXME
+         #:name restrict-typed->-name
+         #:stronger equal?
          #:late-neg-projection restrict-typed->-late-neg-projection))
 
 (define (restrict-typed-field-late-neg-proj ctc)
@@ -147,8 +190,13 @@
        blame val #:missing-party neg-party
        "cannot read or write field hidden by Typed Racket"))))
 
-(struct restrict-typed-field/c (obj name)
+(define (restrict-typed-field-name ctc)
+  (define name (restrict-typed-field/c-name ctc))
+  (build-compound-type-name 'restrict-typed-field/c name))
+
+(struct restrict-typed-field/c (name)
         #:property prop:flat-contract
         (build-flat-contract-property
-         #:name (λ (ctc) '<hidden-field>) ; FIXME
+         #:name restrict-typed-field-name
+         #:stronger equal?
          #:late-neg-projection restrict-typed-field-late-neg-proj))
--- a/typed-racket-test/succeed/opaque-object-name.rkt
+++ b/typed-racket-test/succeed/opaque-object-name.rkt
@@ -1,0 +1,52 @@
+#lang racket
+(require typed-racket/utils/opaque-object rackunit
+         (for-syntax (only-in syntax/srcloc build-source-location-list)))
+
+;; --------------------------------------------------------------------------------------------------
+
+;; object/c-opaque names should be lists:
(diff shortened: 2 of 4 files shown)


rfindler:
Why not recur?

rfindler:
Also, I recommend you implement the name for the contract.

samth:
First, there's no reason to use `eq?` here instead of `equal?`. In general, never use `eq?` without either a fallback to something more comprehensive, or a guarantee that equal things are `eq?` (as with symbols).

Second, to repeat @rfindler's point, we should write a real implementation of this, not just a quick check. 

Third, if that real implementation doesn't fix the other benchmarks, what are the relevant arguments to `contract-stronger?` and why doesn't it produce `#t`?

bennn:
@rfindler @samth I have good & bad news about a unit test I've got for this PR.
https://gist.github.com/bennn/8bc1e02172de200a3c83

Good news: a slow-running object-oriented program is much faster on head (after [efcbd1](https://github.com/racket/racket/commit/efcbd12116e0763ca382163c21cb26916abf904c) and [d58d74](https://github.com/racket/typed-racket/commit/d58d7487e86630ee1b71a67c11b943486965e445)?). The old slowness was due to a method returning `this` across a boundary -- changing to return `void` and making the caller functional solved the slowdown.

Bad news: a minimized version of that program that I was planning to use as a unit test now crashes on head. The error is on vector-ref: `chaperone produced a result that is not a chaperone of the original result`. I've been trying to debug this, but haven't figured out what's wrong. Seems like an opaque object contract is getting replaced by an object contract, but I can't find where that's happening. (I'm also not sure if this is a bug in `class-c-old.rkt` or `opaque-object.rkt`.)

samth:
Can you try git bisect to figure out when it was introduced?

samth:
You'll probably need to first figure out what plain racket program it compares to.

rfindler:
Or just send me a (hopefully smallish) example?

Robby

samth:
Here's a much smaller program with the same problem:

```
#lang racket
(module automata racket
  (provide a)
  (define automata% (class object% (super-new)))
  (define a (new automata%))

  )

(module population typed/racket
  (provide pop)

  (define-type Automata% (Class))

  (define-type Population%
    (Class
     (init-field [a* (Vectorof (Instance Automata%))])
     (match-up* (-> (Instance Population%)))
     (regenerate (-> (Instance Population%)))))

  (require/typed (submod ".." automata)
    (a  (Instance Automata%)))


  (: population% Population%)
  (define population%
    (class object%
      (super-new)
      (init-field a*)

      (define/public (match-up*) this)

      (define/public (regenerate)
        (vector-ref (get-field a* this) 0)
        this)
      ))
  (: pop (Instance Population%))
  (define pop
    (new population% [a* (vector a)]))
  )

(require 'population racket/class)
(send (send pop match-up*) regenerate)
```

Next step, remove TR.

samth:
Since racket/racket#1198 is fixed, what's the status here?

rfindler:
Last I heard, you (or someone) were (was) going to remove TR?

samth:
@rfindler I think that your fix to racket/racket#1198 fixed that bug, and so there's no need to do that. But hopefully @bennn can confirm.

bennn:
Yes that fix fixed the bug. I'll get this PR into shape.

bennn:
Currently:
- an `object/c-opaque` A is stronger than another `object/c-opaque` B if A has no more fields/methods than B and is stronger (in the `object/c` sense) on their common fields/methods
- an `object/c-opaque` A is stronger than an `object/c` C if A is stronger on their common fields/methods (If A has extra members, these extras are stronger than C, which does nothing. If A has fewer members, A is stronger on these because they are opaque.)

The implementation is re-using some functions from `object/c-stronger`. Here: https://github.com/racket/racket/pull/1367

But this is making the sample programs I had earlier (variations 1,2,3) run slower. Need to make sure there's still double wrapping in those; if not, find new programs that show this PR is a good thing.

bennn:
@samth re: the provided identifiers [error from DrDr](http://drdr.racket-lang.org/diff/35900/35901/racket/share/pkgs/typed-racket-test/test-docs-complete.rkt)

1. I can change Racket's `class-internal.rkt` to stop providing the `object/c` helpers ([here](https://github.com/bennn/racket/commit/bb95ffd959946b5a06d5a3640d7a661e0938c2f6))
2. and edit this PR to require directly from `class-c-old.rkt` ([here](https://github.com/racket/typed-racket/pull/267/files#diff-761e8fb3389fa03624b9497d942eb3e5R30)). That should fix the TR docs error.

Sound good? If so, should I edit the previous Racket commit (https://github.com/racket/racket/commit/4650a123505e5cc37954c913c944c7e89166909b) or make a new one removing the exports from `class-internal.rkt`?

samth:
Couldn't class-internal provide them, but not racket/class or racket/contract?

Also, you're not allowed to change commits you've already pushed.

bennn:
`class-internal` could provide them [here](https://github.com/racket/racket/blob/master/racket/collects/racket/private/class-internal.rkt#L32), but we'd have to import them from `class-c-old.rkt` into that scope (they're not there already).

samth:
Ok, I now understand what's going on.

You should just revert the change to `provide-public-names` (as a general rule, don't add public exports without documentation ever) and require `class-c-old` directly, unless @rfindler thinks some other file should be the place to get these functions.

bennn:
ok (yeah I didn't realize I was exporting public names until too late)

rfindler:
Sorry I didn't notice that the names were now public. I think it is okay to require class-c-old to get the names.

bennn:
ok, I will have typed racket get the names from class-internal

bennn:
hah, `class-internal.rkt` is required by `class-c-old.rkt`
```
standard-module-name-resolver: cycle in loading
  at path: /Users/ben/code/racket/fork/racket/collects/racket/private/class-internal.rkt
```

typed-racket-lib/typed-racket/utils/opaque-object.rkt
```diff
@@ -57,21 +60,55 @@
       (dynamic-object/c (append methods remaining-methods)
                         (append method-ctcs
                                 (for/list ([m remaining-methods])
-                                  (restrict-typed->/c)))
+                                  (restrict-typed->/c m)))
                         (append fields remaining-fields)
                         (append field-ctcs
                                 (for/list ([m remaining-fields])
-                                  (restrict-typed-field/c obj m)))))
+                                  (restrict-typed-field/c m)))))
```
bennn: The `obj` field was never used, so I've removed it here. @samth is that ok?

typed-racket-test/succeed/pr267-variation-1.rkt
```diff
@@ -0,0 +1,33 @@
...
+(module u racket/base
+  (require (submod ".." t) racket/contract)
+  (define (main N)
+    (for/fold ([c c])
+              ([i (in-range N)])
+      (id c)))
+  (provide main))
+
+(require 'u racket/sandbox)
+(call-with-limits OPAQUE-TIME 1
```
bennn: Is there a better way to check that `stronger?` kicked in?

typed-racket-lib/typed-racket/utils/opaque-object.rkt
samth: This isn't my code originally. @takikawa?

typed-racket-test/succeed/pr267-variation-1.rkt
samth: Use a custodian to check memory use, maybe.

bennn:
The tests pass, but only because I've removed the bigger examples for now. They're still not really helped by this PR (though I think the current `stronger` check is correct [(tests)](https://github.com/racket/typed-racket/pull/267/files#diff-3829b991151ba49081c47820125a4e9dR1)).

Here's a summary of those tests. I guess I will keep trying this weekend to find a real example where this PR removes a big slowdown.

#### Test 1 [gist](https://gist.github.com/bennn/48046d884507ec5d8804a16e2f0cd5a7)


Repeatedly send an untyped object through the identity function at type `(Instance (Class (f (-> C C)))`. So it's `(id (id ... (id obj) ...))`. Here's how the `time` changes as we increase the number of `id`s.

| num iterations | v6.5.0.5 (ms) | this PR (ms) |
| --- | --- | --- |
| 30000 | 9250 | 7193 |
| 60000 | 17723 | 14423 |
| 90000 | 25791 | 21928 |
| 120000 | 33917 |  28102 |

Small benefits, but nothing dramatic that I think we reliably check for on Travis.


#### Test 2 [gist](https://gist.github.com/bennn/ed372d803d330e9319cf6aff0a3ee305)


One of the slow configurations of the [FSM](https://github.com/nuprl/gradual-typing-performance/tree/master/benchmarks/fsm/typed) benchmark.

| iters | v6.5.0.5 | this PR |
| --- | --- | --- |
| 4 | 2866 | 4022 |
| 6 | 9386 | 11388 |
| 8 | 52496 | 61278 |

About 2 seconds slower. `object/c-stronger` is succeeding, but something else must be causing the exponential slowdown.


#### Test 3 [gist](https://gist.github.com/bennn/cbbc24257beff3ea96ae63c411b5d866)


A slow configuration from the [forth](https://github.com/nuprl/gradual-typing-performance/tree/master/benchmarks/forth/typed) benchmark.

`object/c-stronger` doesn't even work on this code, I guess because it has to compare an infinite `object/c` contract with a contract like `(recursive-contract g27 #:impersonator)`

rfindler:
I'm curious: why are you looking for a performance-related test here? (or
do I misunderstand and you're not?)

bennn:
I am looking for a performance test, because I thought this stronger? check should help gradual typing.

rfindler:
That seems like a reason to do some performance measurements, but it isn't clear why that implies you want a reproducible performance test.

bennn:
Ok. I was hoping to keep one to avoid regressions, but if we can move forward on this without, that would be nice.

rfindler:
I don't feel like we have a good way to test for performance regressions, sadly. :( We can turn some performance regressions into correctness tests (like checking that stronger returns the right things to make sure fast paths are taken) but in general it seems like a hard problem.

bennn:
@takikawa could you review this? Especially:
- is the [`object/c-opaque-stronger`](https://github.com/racket/typed-racket/pull/267/files#diff-761e8fb3389fa03624b9497d942eb3e5R88) predicate correct?
- is it okay that I changed the fields of [`restrict-typed->/c`](https://github.com/racket/typed-racket/pull/267/files#diff-761e8fb3389fa03624b9497d942eb3e5R170) and [`restrict-typed-field/c`](https://github.com/racket/typed-racket/pull/267/files#diff-761e8fb3389fa03624b9497d942eb3e5R196)?
- can we do better than `equal?` as a stronger? check for `restrict-typed-*`?
- any suggestions for more tests?

bennn:
status update: I think this PR implements a correct `stronger?` check for opaque objects, but would like a review from @samth or @takikawa .

(I promise to take another look myself before Sunday & write a few more tests)

samth:
This looks like a nice improvement, but I don't know the object-contract or contract-stronger machinery well enough to be sure (but it does have a lot of tests, which is great). @takikawa, it would be great if you could review this, or failing that perhaps @rfindler.

typed-racket-lib/typed-racket/utils/opaque-object.rkt
```diff
@@ -57,21 +60,54 @@
...
+  (define that-opaque? (base-object/c-opaque? that))
+  (cond
+   [(or that-opaque?
+        (base-object/c? that))
+    (define this-ctc (base-object/c-opaque-obj/c this))
+    (define that-ctc (if that-opaque? (base-object/c-opaque-obj/c that) that))
+    (and
+      (object/c-common-methods-stronger? this-ctc that-ctc)
+      (object/c-common-fields-stronger? this-ctc that-ctc)
+      (when that-opaque?
```
rfindler: I'd write `or` not `when` here. More generally, nothing really "smells" wrong here. I'd say that if you have a good set of tests, then this is probably right.

rfindler: (or maybe `and`...)

samth:
Given @rfindler's comments, I'd go ahead and merge this once you're satisfied.

bennn:
Found an interesting bug in `object/c-clause` (https://github.com/racket/typed-racket/pull/267/commits/c79211bd2db46cd63e039e8f50d01e4e85264ebd) but otherwise this seems ok.

Issue #268, opened 2015-12-17
Can't provide promise for polymorphic struct


jsmaniac:
The fix for GH bug #159 changed the contract for promises in `typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt` from `promise/c` to `(λ (x) (and/c (promise/c x) (not/c promise/name?)))`, but this caused a new bug:

    #lang typed/racket
    (struct (A) s ([f : Any]))
    (define p (delay (s "a")))
    (provide p)

The program above fails to generate the contracts for `p`, with the error message:

    ../../../usr/local/racket-6.3.0.8/share/pkgs/typed-racket-lib/typed-racket/private/type-contract.rkt:100:0: free-identifier=?: contract violation
      expected: identifier?
      given: #<syntax:/usr/local/racket-6.3.0.8/share/pkgs/typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt:156:30 (λ (x) (and/c (promise/c x) ...>
      argument position: 1st
      other arguments...:

samth:
Fixed by #269.

PR #269, opened 2015-12-17
Fixes GH issue #268 “Can't provide promise for polymorphic struct”.

jsmaniac:
See also commit 5cd5f77 “Don't allow promises created with `delay/name` as `(Promise T)`.”.

The contracts in `typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt` should be just a single identifier, not a lambda expression, because `typed-racket-lib/typed-racket/private/type-contract.rkt` relies on that, and passes the contract name to `free-identifier=?`, which won't work on a lambda.

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -166,6 +166,7 @@
       typed-racket/utils/opaque-object
       typed-racket/utils/evt-contract
       typed-racket/utils/sealing-contract
+      typed-racket/utils/promise-not-name-contract
       racket/sequence
       racket/contract/parametric))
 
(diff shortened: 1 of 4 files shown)


samth:
Other than that one thing, looks good!

jsmaniac:
@samth I had followed the example of tr:evt/c , but was getting namespace errors. I'll change that and push a new version.

jsmaniac:
@samth I keep getting the error below, but if I move the contents of the `promise-not-name-contract.rkt` file inside the `evt-contract.rkt` in the same folder, it works fine. Do you have any idea about what I could be doing wrong?

    typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt:157:30: require: namespace mismatch;
     reference to a module that is not available
      reference phase: 0
      referenced module: "/home/travis/build/racket/typed-racket/typed-racket-lib/typed-racket/utils/promise-not-name-contract.rkt"
      referenced phase level: 0
      in: tr:promise-not-name/c

jsmaniac:
My bad, it was because of the `tr:` namespace I think, and I hadn't pushed that change properly. It should work fine now.

PR #270, opened 2015-12-17
Convert TR docs to use scribble/examples

takikawa:
Found two broken examples along the way. These are temporarily disabled or just left alone until we fix the underlying bugs.

### DIFF
--- a/typed-racket-doc/info.rkt
+++ b/typed-racket-doc/info.rkt
@@ -10,7 +10,7 @@
                      "r6rs-lib"
                      "sandbox-lib"
                      "at-exp-lib"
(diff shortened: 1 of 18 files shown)


rfindler:
I made a mistake in a similar commit in redex in not adding a version dependency on scribble lib. I believe you want 1.14.

rfindler:
Actually maybe scribble-lib's version should be bumped to 1.15

rfindler:
Okay, I added 1.15

takikawa:
Closed via f08f3d07d412bcb5e80076b49092fc70f5a09be4 and next two commits

Issue #271, opened 2015-12-17
define-predicate regression


takikawa:
This should be a contract gen error, but instead the contract gen succeeds and leads to an error in Racket:

````racket
#lang typed/racket
 
(define-predicate p? (All (A) (Listof A)))
 
(let ()
  (: x (U (Listof Integer) Integer)) (define x '(1 2 3))
  (if (p? x) 0 (add1 x)))
````

(found via the TR Guide)

**Comment by samth** [comment]
That program is a syntax error for me.

**Comment by samth** [comment]
But this program has the error you described, and is probably what the docs are doing:

```
#lang racket/load
(require typed/racket)

(define-predicate p? (All (A) (Listof A)))

(let ()
  (: x (U (Listof Integer) Integer)) (define x '(1 2 3))
  (if (p? x) 0 (add1 x)))
```

Also, the top-level is hopeless.

takikawa:
This turns out to be a problem with how TR's top-level typechecking handles ignored expressions. I'm looking into a fix.

takikawa:
Fixed already.

PR #272, opened 2015-12-18
Fix GH issue #271

takikawa:
Propagate syntax properties when opening up begins at the top-level so that ignore properties will get transferred.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/toplevel-trampoline.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/toplevel-trampoline.rkt
@@ -66,22 +66,23 @@
           (local-expand/capture* #'e 'top-level (kernel-form-identifier-list))))
        (syntax-parse head-expanded
          #:literal-sets (kernel-literals)
-         [(begin (define-values (n) _) ...
-                 (~and (~or _:ignore^ _:ignore-some^)
-                       (~not (~or _:tr:class^
-                                  _:tr:unit^
-                                  _:tr:unit:invoke^
-                                  _:tr:unit:compound^
-                                  _:tr:unit:from-context^))))
-          head-expanded]
-         ;; keep trampolining on begins
-         [(begin (define-values (n) e-rhs) ... (begin e ... e-last))
-          #`(begin (tc-toplevel-trampoline orig-stx (define-values (n) e-rhs))
-                   ...
-                   (tc-toplevel-trampoline orig-stx e) ...
-                   #,(if report?
-                         #'(tc-toplevel-trampoline/report orig-stx e-last)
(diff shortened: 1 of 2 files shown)


typed-racket-lib/typed-racket/typecheck/toplevel-trampoline.rkt
```diff
@@ -66,22 +66,23 @@
...
        (syntax-parse head-expanded
          #:literal-sets (kernel-literals)
-         [(begin (define-values (n) _) ...
-                 (~and (~or _:ignore^ _:ignore-some^)
-                       (~not (~or _:tr:class^
-                                  _:tr:unit^
-                                  _:tr:unit:invoke^
-                                  _:tr:unit:compound^
-                                  _:tr:unit:from-context^))))
-          head-expanded]
```
stamourv: I don't fully understand how the new TR toplevel works. Why is it now ok to drop this code?

samth: I'm not sure what you mean by "drop", but this code is now handled in the `define-values` handling instead of in the `begin` handling -- @takikawa's new code just pushes the syntax properties around and recurs-via-trampoline on the subforms of the `begin`.

samth:
Looks good.

typed-racket-lib/typed-racket/typecheck/toplevel-trampoline.rkt
stamourv: Just looking at the pattern, it looks like that deleted case handled `begin` forms that didn't end with another `begin`. With the new version of the code, those inputs will flow to the catch-all clause of the `syntax-parse`. Is that ok? Am I missing something?

takikawa: It handled all ignored forms including those that start with `begin` and just returned them, but it turns out this is incorrect since all forms (including ignored forms) need to be local expanded so that any lifts that they do are resolved.

So instead this delegates to the other two clauses, which will either open up begins but propagate syntax properties so that ignored things are ignored properly. Alternatively if it's not a `begin` form then the actual typechecker (and optimizer) kick in after local expansion, but since the ignore property is there it should work properly.

There is a caveat that this will transfer all syntax properties, so TR shouldn't put any properties on top-level `begin`s that are invalid when transferred to subforms.

stamourv: Ok, got it. Thanks.

takikawa:
Merged as b18d940f1a81d7faa5b8030a804e4a2d5811bce4

Issue #277, opened 2015-12-21
Missing type annotation for flatten from racket/list


lexi-lambda:
```racket
#lang typed/racket

(flatten '(1 2 3 (4 5 6)))

; Type Checker: missing type for identifier;
;  consider using `require/typed' to import it
;   identifier: flatten
;   from module: typed/racket in: flatten
```

lexi-lambda:
My attempts to create a valid type for this have mostly been thwarted by TR inferring overly general types. I’ve been finding it difficult to create a type that is restrictive enough while still keeping the flexibility of untyped `flatten`.

This is the type I logically came up with for this function:

```racket
(All [a] ((Rec Flattenable (U a (Listof Flattenable))) -> (Listof a)))
```

This works great if manually instantiated:

```racket
> ((inst flatten Integer) '(1 2 3 (4 5 6)))
- : (Listof Integer)
'(1 2 3 4 5 6)
```

Unfortunately, the type is far too general, since `a` can just refer to the entire first argument, producing horribly incorrect result types:

```racket
> (flatten '(1 2 3 (4 5 6)))
- : (Listof
     (List
      One
      Positive-Byte
      Positive-Byte
      (List Positive-Byte Positive-Byte Positive-Byte)))
'(1 2 3 4 5 6)
```

I’m not sure how to fix this. Semantically, the issue seems to be that `a` should never be allowed to be a list, but encoding this in the type system doesn’t seem possible? Perhaps the only way to type this function would be to use a more restrictive type? Something like this:

```racket
(All [a] ((Listof (Listof a)) -> (Listof a)))
```

That type is actually wrong, though, because it will still be incorrect if a list of lists of lists is flattened:

```racket
> (flatten '(((1) (2)) ((4) (5))))
- : (Listof (List Positive-Byte))
'(1 2 4 5)
```

Is this function typeable with Typed Racket?

AlexKnauth:
That more restrictive type is what `append*` does, which is sort of like `flatten` except exactly 2 levels deep. 

But `flatten` can go arbitrarily many levels deep, even with the base case of zero levels deep: `(flatten "something") = '("something")`. As you said, this would require that `a` can't be a any kind of list type, which can't be expressed with the current type system.

samth:
Unfortunately, `flatten` can't be given a particularly useful type, as @AlexKnauth points out. But we should give it _some_ type, such as `((Listof Any) -> (Listof Any))`.

AlexKnauth:
[...]

PR #278, opened 2015-12-24
add type for flatten

AlexKnauth:
fixes https://github.com/racket/typed-racket/issues/277

Can’t make it polymorphic, settle for `Any -> (Listof Any)`


samth:
Looks good once the tests pass. And we should add a progress output so they stop timing out.

AlexKnauth:
I'm getting a lot of integration test failures because of errors like `"build-path: contract violation\n  expected: (or/c path-for-some-system? path-string? 'up 'same)\n  given: #f\n  argument position: 1st\n  other arguments...:\n   \"pkgs\""`.

PR #286, opened 2016-01-06
Add typed/racket/random

takikawa:
Should probably specify a package dependency on a newer "base".

### DIFF
--- a/typed-racket-compatibility/info.rkt
+++ b/typed-racket-compatibility/info.rkt
@@ -11,4 +11,4 @@
 
 (define pkg-authors '(samth stamourv))
 
-(define version "1.3")
+(define version "1.5")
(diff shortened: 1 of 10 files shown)


samth:
Didn't the type of random also change?

stamourv:
@takikawa: Depending on the current version of base should be fine. It's not a released version.

takikawa:
I'm actually not sure what the type of `random` should be. It's currently this:

````
[random
  (cl->* (->opt -PosFixnum [-Pseudo-Random-Generator] -NonNegFixnum)
         (->opt -Int [-Pseudo-Random-Generator] -Nat)
         (->opt [-Pseudo-Random-Generator] -Flonum))]
````

so now that it has min and max arguments, should those be PosFixnum or do we need 4 cases here or something?

samth:
I think you can rely on the error when the min is bigger than the max to reduce the number of cases.

stamourv:
There should be cases for these arguments to be fixnums, but there should also be cases where they're allowed to be arbitrary Integers.

PR #288, opened 2016-01-09
adjust doc `extfvector` example

mflatt:
Make the example typeset in a Racket build that does not support
extflvectors.


samth:
Looks good.

samth:
But actually, won't this give the wrong impression on systems that don't support extflonums?

mflatt:
I'm not sure what you have in mind. On most platforms, the documentation arrives built as part of the installation, so I don't think there's a simple way to report the expected result on the current platform (which would also assume that you're reading the docs on the relevant platform, or something like that).

samth:
I'm thinking about the difference between this PR, and #284. In either one, docs that are built on a system with extflonum support with show the same thing, just like they currently do. With yours, if @wargrey builds on their system, it will look the same as on docs.racket-lang.org. With #284, the build on an openindiana system will show that using extflonums will fail.

mflatt:
I think we want the docs to be independent of the machine where they're built, so that why I went with the solution here (in contrast to #284).

samth:
If we want consistency, then this is definitely the right choice.

PR #295, opened 2016-01-16
Follow TR error message conventions, as suggested by samth in PR #250 .

jsmaniac:
I finally got around to change how arguments and other pieces of information are printed in TR's error messages, to follow the [error message conventions](http://docs.racket-lang.org/reference/exns.html?q=raise-arg#%28part._err-msg-conventions%29), as suggested by @samth [in PR #250](https://github.com/racket/typed-racket/pull/250#issuecomment-158788270).

One thing worth noting: The `(Poly-unsafe: n _)` and `(Name: name-id num-args #f)` now both give the same `wrong number of arguments to polymorphic type` error, as I couldn't find a meaningful difference for when they occur, from a user's perspective (it seems the former occurs in recursive applications, and the latter in non-recursive applications, but I'm not 100% sure). If the messages should make a distinction, please tell me what's the difference between the two cases, and I'll adjust the error messages.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-types-extra.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-types-extra.rkt
@@ -8,7 +8,7 @@
      #'(begin (define-syntax nm
                 (lambda (stx)
                   (raise-syntax-error 'type-check
-                                      (format "type name ~a used out of context in ~a"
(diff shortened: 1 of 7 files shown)


stamourv:
Merged, thanks!

PR #299, opened 2016-01-25
add hash-iterate-pair and hash-iterate-key+value

stchang:
To accompany https://github.com/racket/racket/pull/1224


stchang:
If I merge this, does the version need to be bumped?

PR #300, opened 2016-01-26
Add contract profiling instrumentation to combinators defined by TR.

stamourv:
Aside from `Any`, I'm not sure how to trigger those contracts reliably (without causing them to blow up), which I'd like to do for testing. Any suggestions?

### DIFF
--- a/typed-racket-lib/typed-racket/utils/any-wrap.rkt
+++ b/typed-racket-lib/typed-racket/utils/any-wrap.rkt
@@ -69,6 +69,7 @@
      "Attempted to use a higher-order value passed as `Any` in untyped code: ~v" v))
   
   (define (wrap-struct neg-party s)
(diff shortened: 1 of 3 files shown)


takikawa:
The examples in succeed tests should work. Like https://github.com/racket/typed-racket/blob/master/typed-racket-test/succeed/sealing-contract-1.rkt and https://github.com/racket/typed-racket/blob/master/typed-racket-test/succeed/opaque-object-contract.rkt

takikawa:
BTW, there is also a custom contract for evts in TR. Should that be instrumented too?

stamourv:
Re tests: Ok, will have a look at those.
Re `evt/c`: It basically reuses Racket's `evt/c`'s except for one case that errors right away, so the instrumentation in `evt/c` should do the job.

stamourv:
Merged, thanks!

PR #301, opened 2016-02-02
[wip] add type-out provide-spec

bennn:
(Work-in-progress because there's a bug with `(type-out (struct ...))` -- you can't use the struct inside the defining module.)

Adds initial support for a `type-out` provide spec.

Design goals:
- Behavior is a superset of `contract-out` (currently missing `#:exists` and `#:forall`)
- Look like an ML-style interface, with all type defs. and typed identifiers

Current look:

```
#lang typed/racket/base

(provide
  (type-out
    (type FullName (List String String))
    (struct person ([name : FullName] [id : Natural]))
    (get-first-name (-> FullName String))))

(define (get-first-name fn)
  (car fn))
```

Later on, I'd like to add `#:exists` & `#:forall` to bind variables across multiple type definitions, and options to export opaque types and ADTs. But I think it's better to address those separately.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
@@ -315,11 +315,20 @@
               (~datum prefix-all-defined) (~datum prefix-all-defined-except)
               (~datum expand)))))
 
+;; Move type declarations to the beginning of the list,
+;;  keep other declarations in the same relative order.
+;; (-> (Listof Syntax) (Listof Syntax))
+(define (lift-type-declarations form*)
+  (define (is-type-decl? form)
+    (syntax-parse form [_:type-declaration #t] [_ #f]))
+  (let*-values ([(type-decl* form*) (partition is-type-decl? form*)])
+    (append type-decl* form*)))
+
 ;; actually do the work on a module
 ;; produces prelude and post-lude syntax objects
 ;; syntax-list -> (values syntax syntax)
 (define (type-check forms0)
-  (define forms (syntax->list forms0))
+  (define forms (lift-type-declarations (syntax->list forms0)))
   (do-time "before form splitting")
   (define-values (type-aliases struct-defs stx-defs0 val-defs0 provs signature-defs)
     (filter-multiple
--- a/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
@@ -2,6 +2,7 @@
 
 @begin[(require "../utils.rkt" scribble/example racket/sandbox)
        (require (for-label (only-meta-in 0 [except-in typed/racket])
+                           (only-in racket/contract contract-out)
                            (only-in racket/base)))]
 
 @(define the-eval (make-base-eval))
@@ -514,9 +515,6 @@ for function types.
     (: var4 : String -> Integer)]
 }
 
-@defform[(provide: [v t] ...)]{This declares that the @racket[v]s have
-the types @racket[t], and also provides all of the @racket[v]s.}
-
 @defform/none[#{v : t}]{ This declares that the variable @racket[v] has type
 @racket[t].  This is legal only for binding occurrences of @racket[_v].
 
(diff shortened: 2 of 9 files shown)


typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -315,11 +315,20 @@
...
+    (syntax-parse form [_:type-declaration #t] [_ #f]))
+  (let*-values ([(type-decl* form*) (partition is-type-decl? form*)])
+    (append type-decl* form*)))
+
 ;; actually do the work on a module
 ;; produces prelude and post-lude syntax objects
 ;; syntax-list -> (values syntax syntax)
 (define (type-check forms0)
-  (define forms (syntax->list forms0))
+  (define forms (lift-type-declarations (syntax->list forms0)))
```
bennn: Possible issue: I'm re-ordering the top-level forms so that type declarations come first. Without this, type annotations aren't associated with the right definitions.

Not sure whether this will change to fix the bug with structs.

samth:
A few thoughts:

* Can you describe the intended semantics of this briefly? Is it a replacement for top-level type annotations, or an ML-style signature (ie, you'd still need annotations internally), or something else?
* I really don't think struct _definitions_ should go in the `provide` form. That's not how it works in `contract-out` -- you have to have a definition of the struct in the module, and it's not how it works in ML either. Changing that should make the struct issue easier to solve.

bennn:
- Yes, this is intended as a replacement for top-level type definitions.* The idea is to put types where the user is more likely to see them & [keep abusing](http://docs.racket-lang.org/style/Units_of_Code.html) `provide` as an API. I also want to offer a smooth migration from contract-out -- just change "contract" to "type" and you're rolling.
- Removing struct defs would solve the issue, but then `(type-out (struct ...))` is pretty much a no-op. We can only check that both type annotations (in the provide & at the definition) match.

`*` Non-provided functions would still need an annotation

bennn:
Leif pointed out that we definitely don't want struct properties declared in a provide form, so now I'm in favor of writing the fields twice (in code & in the `type-out`). We could then also use the `type-out` to hide some fields or use different types.

samth:
I think we should start with something that just handles type annotations, and see how that works. I don't think provide should produce definitions (except as an implementation detail, as in contract-out), and I think information hiding a la ML modules is a larger design issue that deserves more thought before we start implementing.

bennn:
Ok. Do you mind if it allows `define-type` aliases?

```
... (type-out T (Listof Integer)) ...

=== becomes ===>

(define-type T (Listof Integer))
(provide T)
```

samth:
That's what I mean about definitions. I don't think provide should introduce new bindings in the module. Nothing else in Racket works that way.

bennn:
Checkpoint:
- No longer makes definitions, just adds annotations & renames.
- TODO: struct types are currently ignored. We should check that:
    - field types match implementation
    - super declarations are correct
    - all fields are annotated (can't leave out field / super field from `type-out`)
    - (anything else?)

Is there an easy way to get struct information from the struct type name? I tried `struct-type-info` but I haven't got the right inspector. (I can probably copy what `contract-out` does, once I understand that.)

```
#lang typed/racket/base
(struct s ())
(call-with-values (lambda () (struct-type-info struct:s)) list)
;; struct-type-info: current inspector cannot extract info for structure type
```

bennn:
Things are pretty good now, but I'm blocked on 1 problem and have a few questions:

__Problem:__ this is currently incompatible with the `#:type-name` option because I assume the struct name is the type name. How can I get a type name from a struct name?

__Questions__
- `type-out` annotations for struct fields are only checked, not enforced. So `(struct s ([x : Integer])) ... (type-out s ([x : Real]))) ...` is valid and users will get integers from `s-x`. Should we fix this, and reject non-exact `type-out` annotations?
- Currently check struct recursively to make sure all supertype fields are given annotations. Should we not do that (I'm thinking we should not do that)
- Order of struct field annotations is not enforced. (Probably should be)

bennn:
status update: this PR is not ready to merge because it doesn't handle structs as well as `contract-out`. Need to fix #322 first.

PR #303, opened 2016-02-03
Add typed/images/icons and typed/images/logos

wargrey:
1. Add type definitions for images/icons and images/logos.
2. Fix up some mis-groupped or outdated modules in documentation.

I am not sure if these files should be pushed here, maybe the images-lib is the proper package to push into.

### DIFF
--- a/typed-racket-more/typed/images/icons.rkt
+++ b/typed-racket-more/typed/images/icons.rkt
@@ -1,0 +1,210 @@
+#lang typed/racket/gui
+
+(require typed/racket/draw)
+
+(require "../private/require-batch.rkt")
+
+;;; The Ray Tracer API has not stable enough yet,
+;;; hence all the commented #:material arguments.
+;;; As an alternative, please consider working
+;;; with images/private/deep-flomap.rkt.
+
+(require/typed/provide
+ images/icons/style
+ [default-icon-height (Parameterof Nonnegative-Real)]
+ [toolbar-icon-height (Parameterof Nonnegative-Real)]
+ [default-icon-backing-scale (Parameterof Positive-Real)]
+ [light-metal-icon-color (U String (Instance Color%))]
+ [metal-icon-color (U String (Instance Color%))]
+ [dark-metal-icon-color (U String (Instance Color%))]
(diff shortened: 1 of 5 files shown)


typed-racket-more/typed/images/logos.rkt
```diff
@@ -0,0 +1,22 @@
+#lang typed/racket
+
+(require typed/racket/draw)
+
+(module digitama typed/racket
+  (provide (all-defined-out))
+  
+  (define-syntax (require/typed/provide/batch stx)
```
samth: I think this should go in `typed/private`, since it's probably useful elsewhere.

typed-racket-more/typed/images/icons.rkt
```diff
@@ -0,0 +1,212 @@
+#lang typed/racket/gui
+
+(require "logos.rkt" (submod "logos.rkt" digitama))
+(require typed/racket/draw)
+
+(require/typed
+ images/private/deep-flomap
+ [#:opaque Flomap-Material deep-flomap-material-value?])
```
samth: `deep-flomap` is already typed, I think.

wargrey: Yes, so `deep-flomap` is.
But here is `deep-flomap-material-value` which is left untyped, and lots of values of its instances are undocumented.

typed-racket-more/typed/images/logos.rkt
wargrey: Oh, thanks.
I just did not know how to handle this before.

typed-racket-more/typed/images/icons.rkt
samth: Sure, but either you should use the typed version in `images/private/deep-flomap` or the untyped version in `deep-flomap-untyped-parameters`. Using `require/typed` here just adds an extra set of contracts.

takikawa:
@samth were your comments resolved on this and should we merge it for 6.5?

samth:
This looks good to me.

takikawa:
Merged as 65441301c23998b08f7ac0c1aa35a3a8605053f7

PR #309, opened 2016-02-08
Fixup wrong type definitions

### LINKED ISSUE #307: Wrong type definition of snip%
(get-snipclass) should return `(Option (Instance Snip-Class%))` rather than `(Option Snip-Class%)`

wargrey:
(set/get-keymap)
(get-snipclass)

fixed #307


stamourv:
Merged, thanks!

Issue #315, opened 2016-03-01
Syntax properties are not forwarded by `define-typed/untyped-identifier`


jsmaniac:
### What version of Racket are you using?
6.4.0.8 (20160226-15f47ef)

### What program did you run?
```racket
#lang typed/racket

(module t/u racket
  (provide typed/untyped
           (rename-out [typed/untyped-typed original-req-stx]))
  
  (require typed/untyped-utils
           racket/require-syntax
           (for-syntax syntax/strip-context))
  
  (define-require-syntax (typed/untyped-typed stx)
    (displayln 'typed)
    (syntax-case stx ()
      [(_ m s) (replace-context stx #'(submod m s typed))]))
  
  (define-require-syntax (typed/untyped-untyped stx)
    (displayln 'untyped)
    (syntax-case stx ()
      [(_ m s) (replace-context stx #'(submod m s untyped))]))
  
  (define-typed/untyped-identifier typed/untyped
    typed/untyped-typed
    typed/untyped-untyped))

(module m racket
  (module typed typed/racket)
  (module untyped typed/racket/no-check))

(require 't/u)

;; Works:
(require (original-req-stx "." m))
;; Does not work: typed/untyped: not a require sub-form
(require (typed/untyped "." m))
;; But this works:
(define-syntax renamer (make-rename-transformer #'original-req-stx))
(require (renamer "." m))
```

### If you got an error message, please include it here.

```
typed/untyped: not a require sub-form in: (typed/untyped "." m)
```
### What should have happened?

`typed/untyped` should be accepted as a require sub-form.

I suspect that [typed/untyped-renamer](https://github.com/racket/typed-racket/blob/ca46d80189058d612d0ecbae650dba59d8dc7546/typed-racket-more/typed/untyped-utils.rkt#L32) does not forward the syntax properties of the original identifiers.

jsmaniac:
I tried fixing it by returning `(make-rename-tranformer i)` from `typed/untyped-renamer`, but it doesn't work ([write: cannot marshal value that is embedded in compiled code: #<rename-transformer>](https://travis-ci.org/jsmaniac/typed-racket/builds/112851743#L291), because that's returning a transformer procedure, instead of returning a piece of syntax I think).

`make-rename-transformer` seems to be defined in racket's core, so I don't really know how to emulate its behaviour.

stamourv:
Closed by #323

PR #323, opened 2016-03-16
Fixes GH issue #315 Syntax properties are not forwarded by `define-typed/untyped-identifier`

jsmaniac:
@samth Thanks for your suggestions earlier. I simply used `make-typed-renaming` from https://github.com/racket/typed-racket/blob/26c4a199fb614cb9244adae1f239e0323b9a8b18/typed-racket-lib/typed-racket/typecheck/renamer.rkt , as directly returning a rename-transformer didn't seem to work.

I suspect that the `(if (unbox typed-context?)` actually gets evaluated where the module is defined, not where it is required. Also, even if it did get evaluated when required, I doubt the identifier would behave properly inside a `(with-typed …)` region since it would be bound to the untyped `rename-transformer` when required, but subsequently used in a typed context. Using `prop:rename-transformer` like `typed-racket-lib/typed-racket/typecheck/renamer.rkt` does seems to be the right way to do it, so I just reused that.

However the typed/untyped identifier generated that way is caught by `un-rename`, so could you please check that this isn't a problem?

### DIFF
--- a/typed-racket-more/typed/untyped-utils.rkt
+++ b/typed-racket-more/typed/untyped-utils.rkt
@@ -4,7 +4,8 @@
                      syntax/parse
                      syntax/stx
                      racket/syntax
-                     typed-racket/utils/tc-utils)
+                     typed-racket/utils/tc-utils
+                     typed-racket/typecheck/renamer)
          typed-racket/utils/tc-utils)
 
 (provide syntax-local-typed-context?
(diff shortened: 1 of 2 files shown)


samth:
This looks like the right thing to me. Once the tests pass, it's good to merge.

stamourv:
Looks like tests are passing. I'll merge.

Issue #325, opened 2016-03-28
Type information shown in mouse over tooltips is overly generic


jackfirth:
When running this program under racket version 6.4:

```
#lang typed/racket/base
(map bytes->string/utf-8 '(#"abc" #"ABC" #"123"))
```

I can mouse over expressions in DrRacket to see their type. However the displayed type of map is the general case variadic polymorphic type. It would be nice if the mouse over type used the type information inferred from the use site. The above example shows this when mousing over `map`:

```
(All (c a b ...)
  (case->
    (-> (-> a c) (Pairof a (Listof a)) (Pairof c (Listof c)))
    (-> (-> a b ... b c) (Listof a) (Listof b) ... b (Listof c))))
```

I'd like it to show this:

```
(-> (-> Bytes String)
    (Pairof Bytes (Listof Bytes))
    (Pairof String (Listof String)))
```

takikawa:
Thanks! Now fixed.

PR #326, opened 2016-03-28
Issue 325

takikawa:
Addresses issue #325.

Improves tooltips for code like

    (map bytes->string/utf-8 '(#"abc" #"ABC" #"123"))

by showing

    (-> (-> Bytes String)
        (Pairof Bytes (Listof Bytes))
        (Pairof String (Listof String)))

instead of a polymorphic type for `map`.

For `case->` functions, it also changes the tooltip. For example the type of `+` in

    (+ 1 2)

is `(-> Postive-Byte Positive-Byte Positive-Index)` instead of `(-> Number * Number)`.


stamourv:
LGTM

jackfirth:
This is fantastic, thank you @takikawa!

PR #328, opened 2016-03-31
Fixup some wrong definitions; provide editor-snip%.

wargrey:
1. Provide editor-snip% with Editor-Snip%.
       The (require/typed) way has a strange behaviour, the caret is not shown in the embedded text%.

2. Fixup the definition of (get-filename) of Editor<%>
      Its argument is optional, this affects all the untyped code that invokes it without argument, including the code inside editor-snip%.

3. Fixup a typo of (make-font).


takikawa:
Thanks! Merged as bf24ebdd654837de18785e57b335f980eea07dda

PR #331, opened 2016-04-07
Fixup wrong type definition: set-caret-owner, it needs (Instance Snip%)

wargrey:
fixup wrong type definition of (set-caret-owner), it needs (Instance Snip%).


takikawa:
Thanks!

PR #335, opened 2016-04-07
Add syntax for poly structs in require/typed

takikawa:
This adds syntax for supporting type variables in `require/typed`. Unfortunately due to contract generation issues, this doesn't actually work in normal Typed Racket.

Using unsafe imports, however, makes it useful. So some design questions: (1) should I make it only work in `unsafe-require/typed` and spit out an error in TR? (e.g., "polymorphic structs in require/typed are not supported") (2) the interaction with inheritance seems to be wonky, maybe I should only allow it for no-parent structs?


typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -152,7 +153,8 @@
    (pattern oc:opaque-clause #:attr spec
      #`(require/opaque-type oc.ty oc.pred #,lib . oc.opt))
    (pattern (~var strc (struct-clause legacy)) #:attr spec
-     #`(require-typed-struct strc.nm (strc.body ...) strc.constructor-parts ...
+     #`(require-typed-struct strc.nm (strc.tvar ...)
```
stamourv: Doesn't this make the list of type variables mandatory?

takikawa: There is a default for this syntax class attribute (`strc.tvar`) that is the empty list.

stamourv: Oh, right. I missed it. Nevermind.

samth:
@takikawa is this done?

takikawa:
@samth It's done in the sense that it's implemented and tested, but the two design questions in the original comment remains. That is, (1) should we still merge it even if it doesn't work in normal TR? And should it be (2) disallowed on structs with parents? (my opinion is yes to both since that's still conservative)

samth:
I agree on yes to both.

takikawa:
Ok I plan to merge this soon. I couldn't reproduce the issues with struct inheritance I saw earlier, so I added a test for it. I also made it raise an error when type variables are provided and it's not in unsafe mode.

takikawa:
Merged.

Issue #336, opened 2016-04-08
Error in TR's internals instead of a propper error message


jsmaniac:
### What version of Racket are you using?
6.4.0.15 (20160407-4bcb657)

### What program did you run?

```
#lang typed/racket
(: foo (∀ (A ... B ...) (→ (List (→ A ... B) ...)
                           Any)))
(define (foo f)
  (apply conjoin f))
```
### If you got an error message, please include it here.

(I formatted the message below for clarity.)

```
../../nix/store/5h7j3m2nkdsdnhmh8i3i1lvxl1dal44s-racket-6.4.0.15/share/racket/pkgs/typed-racket-lib/typed-racket/infer/fail.rkt:17:5:
  match*: no matching clause for
  (list
   (seq '() (dotted-end anonymous-module406473 'anonymous-module406473))
   (seq '() (dotted-end A 'B)))
```

### What should have happened?

I think this program should typecheck, but I'm not sure. Otherwise a propper error message should be given, instead of a failure in TR's internals.

takikawa:
Fixed via af385d6932b6855f199b703f24c569bdc6ef9eb4

PR #339, opened 2016-04-08
Fix issue #336

takikawa:
I think this might be the right fix for #336, but I'm not very familiar with the inference code so I'd appreciate if someone can double check.


takikawa:
Merged as af385d6932b6855f199b703f24c569bdc6ef9eb4

Issue #114, opened 2015-04-10
fx- has totally broken type


endobson:
This typechecks:
```
#lang typed/racket
 
(require racket/fixnum)

(ann (fx- 0 5) Byte)
```

takikawa:
Fixed already.

Issue #113, opened 2015-04-10
fx+ is unsound


endobson:
This typechecks.
```
#lang typed/racket
 
(require racket/fixnum)
  
(ann
  ((ann fx+ (Byte Byte -> Positive-Index))
   0 0)
  Positive-Index)
```

takikawa:
Fixed already.

Issue #111, opened 2015-04-10
fl= is unsound


endobson:
```
endobson@yggdrasil () ~/proj/racket/plt/extra-pkgs/typed-racket % racket
Welcome to Racket v6.2.0.2.
-> (require typed/racket)
-> (require racket/flonum)
-> (: y Positive-Flonum)
-> (define y (let ([x : Nonnegative-Flonum 0.0]) (if (fl= (ann +nan.0 Flonum-Zero) x) 1.0 x)))
-> y
- : Flonum [more precisely: Positive-Flonum]
0.0
->
```

This is because nan.0 is a valid Flonum-Zero

takikawa:
Looks to be fixed already.

PR #340, opened 2016-04-09
regression tests for fl ops

bennn:
For github issues #111 #113 #114 #115

### DIFF
--- a/typed-racket-test/unit-tests/typecheck-tests.rkt
+++ b/typed-racket-test/unit-tests/typecheck-tests.rkt
@@ -425,6 +425,30 @@
         (tc-e (expt -0.0 -1.0) -NonNegFlonum)
         (tc-e (expt 0 (flabs (cos (real->double-flonum 2))))
               -Real)
+        (tc-e/t ;; Github issue #115
+           (let ([x : Flonum 5.0])
+             (if (fl>= x (ann -4.0 Nonpositive-Flonum))
+               x
+               -3.0))
+           -Flonum)
+        (tc-e/t ;; Github issue #115
+          (let ([x : ExtFlonum 5.0t1])
+            (if (extfl>= x (ann -4.0t1 Nonpositive-ExtFlonum))
+              x
+              -3.0t1))
+          -ExtFlonum)
+        (tc-e ;; Github issue #114
+          (fx- 0 5)
+          -NegFixnum)
+        (tc-err ;; Github issue #113
+          ((lambda ([f : (-> Byte Byte Positive-Index)]) 0)
+           fl+))
+        (tc-e/t ;; Github issue #111
+          (let ([x : Nonnegative-Flonum 0.0])
+            (if (fl= (ann +nan.0 Flonum-Zero) x)
+              1.0
+              x))
+          -NonNegFlonum)
         (tc-e (expt
                (sub1 (gcd (exact-round 1)))
                (- (ceiling (real->double-flonum -2.6897657f0))))


typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -425,6 +425,29 @@
         (tc-e (expt -0.0 -1.0) -NonNegFlonum)
         (tc-e (expt 0 (flabs (cos (real->double-flonum 2))))
               -Real)
+        (tc-err ;; Github issue #115
+          ((lambda ([x : Nonpositive-Flonum]) x)
```
bennn: At first I tried using `ann`, but then the tests didn't work. Is there a better way to give an incorrect type that I want to assert fails?

For example if I replace `lambda` with `ann` in this test I get:

```
Typechecker tests > tc-expr tests > 429 (FAIL (ann (let ((x : Flonum 5.0)) (if (fl>= x (ann -4.0 Nonpositive-Flonum)) x -3.0)) Nonpositive-Flonum))
429 (FAIL (ann (let ((x : Flonum 5.0)) (if (fl>= x (ann -4.0 Nonpositive-Flonum)) x -3.0)) Nonpositive-Flonum))
FAILURE
location:   <pkgs>/typed-racket-test/unit-tests/typecheck-tests.rkt:428:8
actual:     (tc-results
 (list
  (tc-result
   (U Float-Nan Float-Positive-Zero Float-Negative-Zero Negative-Float-No-NaN)
   (Top | Bot)
   -))
 #f)
expected:   (tc-results (list (tc-result (U) (Bot | Bot) -)) #f)
tc-expr did not return the expected value. The types don't match.
```

samth:
First, you're right that `(ann e T)` always has type `T`; this is an important invariant in Typed Racket. But I don't see why these need to be error tests in the first place. Couldn't we just check that they have the appropriate type? Or if you want to make sure that certain argument combinations don't work, the use them explicitly.

bennn:
Fair point, I had just copied endobson's examples to make the tests. I've gone back and made 1 of the 4 a positive test.

But for the others, I think it's easier to check errors. For issues 111 113 and 115, the correct type is a supertype of the one we want to rule out. So checking that we get a `Flonum` doesn't rule out that the value would typecheck as a `Nonpositive-Flonum` (right?).

samth:
No, the `tc-e` form checks that the result is exactly as specified (try it out).

bennn:
Oh that's nice. But I'm having trouble getting the types to match exactly. (I guess, on Monday I can talk to Asumu and get these working correctly.)

Here I expect `x` to have type `Flonum` exactly (that's what happens in `#lang typed/racket`)
```
(tc-e
 (let ([x : Flonum 5.0])
  (if (fl>= x (ann -4.0 Nonpositive-Flonum))
   x
   -3.0))
 -Flonum)
```

But it's got the wrong filters, I think.

```
actual:     (tc-results
 (list
  (tc-result
   (U
    Float-Nan
    Float-Positive-Zero
    Float-Negative-Zero
    Positive-Float-No-NaN
    Negative-Float-No-NaN)
   (Top | Bot)
   -))
 #f)
expected:   (tc-results
 (list
  (tc-result
   (U
    Float-Nan
    Float-Positive-Zero
    Float-Negative-Zero
    Positive-Float-No-NaN
    Negative-Float-No-NaN)
   (Top | Top)
   -))
 #f)
```

stamourv:
The `actual` filter looks right to me. You can pass a `#:ret` argument to `tc-e` to specify filters in expected results.

bennn:
Thanks @stamourv. They're all positive tests except for [the one about `fl+`](https://github.com/racket/typed-racket/pull/340/files#diff-14aac72a8cf6cd60384e6a05c1c62423R443). Ok?

stamourv:
Sure. That works.
And, as @samth said, the tests check that the actual types are *exactly* the same as the expected ones, so they already catch errors in both directions.
LGTM.

samth:
Looks good to me too.

bennn:
thanks sam

PR #349, opened 2016-04-22
filter -> prop

andmkent:
This pull request is largely a renaming effort to clean up the TR codebase. There are two primary things I wanted to change:

1. Replace all occurrences of "filter" with "prop" or "proposition"

  * The word "filter" is a meaningless opaque term at this point in the Typed Racket implementation. If anything, it just adds confusion to why things are the way the are. We should use "proposition" instead, since that's what they actually are.

2. Avoid using "Top" and "Bottom" in _both_ the type and proposition realm.

 * Currently the top type is called ```Univ``` and the bottom type is called ```bottom```, while the top proposition is called ```Top``` and the bottom proposition is called ```Bot```. This is just unnecessarily confusing, doesn't really line up w/ the user-space names, and doesn't line up with the names we use in TR formalisms. Worse, all of the top types of primitive types---e.g. the type of all structs, ```StructTop```--- use ```Top```, so it is really easy to get confused about what name to use for these sorts of things. 

With these issues in mind, I made the following changes to names:
```
Top -> TrueProp
Bot -> FalseProp
TypeFilter -> TypeProp
NotTypeFilter -> NotTypeProp
AndFilter -> AndProp
OrFilter -> OrProp
-filter t o -> -is-type o t
-not-filter t o -> -not-type o t
FilterSet -> PropSet
NoFilter -> #f
NoObject -> #f
-FS -> -PS
-top -> -tt
-bot -> -ff
implied-atomic? q p -> implies-atomic? p q
#:Filter -> #:Prop
filter-rec-id -> prop-rec-id
-no-filter -> -no-propset
-top-filter -> -tt-propset
-bot-filter -> -ff-propset
-true-filter -> -true-propset
-false-filter -> -false-propset
#:filters -> #:props
PredicateFilter: -> PredicateProp:
add-unconditional-filter-all-args add-unconditional-prop-all-args
```

There may be some things that need adjusted, but this is at least the initial renaming. Comments welcome.

EDITS: added more details - initially tossed this up last night w/o much info.


samth:
Overall, a good idea. I'm not sure about some of the other renaming, such as getting rid of top/bot.

andmkent:
My main concern is that the whole "Top" and "Bot" notion is used in different places to mean different things and it largely doesn't align with anything outside the implementation (the user-space names, formalisms, etc...) and it is unnecessarily complicated and confusing.

andmkent:
What if we used the user-space names for types (e.g. Any and Nothing instead of Univ and Bottom) and then kept top/bot for propositions? I dunno. There's got to be a clear way to name these things that doesn't have unnecessary overlap.

samth:
What about true-prop?

andmkent:
TrueProp and FalseProp? That seems pretty clear and unambiguous to me.

samth:
Sounds good.

andmkent:
Should they still print to users as Top and Bot as they do in TR today? Or... tt and ff? Or... Truth and Falsehood? hehehe I dunno.

samth:
Let's deal with user-facing issues in a separate change.

lexi-lambda:
This is cool. I think the name “filter” is a little confusing, and I like that you’ve taken the time to try and change it instead of just sticking with it because it’s been around for a while. :)

andmkent:
@lexi-lambda as someone who started reading the papers before looking at TR's internals, I've been dying to make this change =)

andmkent:
These changes cause the following error in the htdp-lib pkg:

```racket
/htdp-lib/typed/test-engine/type-env-ext.rkt:9:26: cannot open module file

  module path: (lib "typed-racket/rep/filter-rep.rkt")
```

This is because htdp-lib relies on a particular file within TR that this PR renames. The file name "filter-rep.rkt" just needs to be changed to "prop-rep.rkt" over in the htdp-lib if/when this pull request is merged.

andmkent:
I added an almost empty "filter-rep.rkt" that just exports the defs from the new "prop-rep.rkt" module--- this commit now doesn't break htdp-lib and passes all the tests just fine.

(we can change htdp-lib later and then cut the stub file)

andmkent:
Oh, and to be clear, none of these changes are user visible (at least that's the intent) -- this is entirely an in-house spring cleaning PR.

samth:
Can you put the contents of this PR message in the commit message?

typed-racket-lib/typed-racket/base-env/base-env-numeric.rkt
```diff
@@ -70,14 +70,14 @@
     (-> t1 t2 B))
   ;; simple case useful with equality predicates.
   ;; if the equality is true, we know that general arg is in fact of specific type.
-  (define (commutative-equality/filter general specific)
-    (list (-> general specific B : (-FS (-filter specific 0) -top))
-          (-> specific general B : (-FS (-filter specific 1) -top))))
+  (define (commutative-equality/prop general specific)
+    (list (-> general specific B : (-PS (-is-type 0 specific) -tt))
```
samth: Why the different order for `-is-type`?

andmkent: It read left-to-right more clearly to me as (-is-type x t), since it then reads in the same order as "x is of type t"

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
```diff
@@ -89,10 +89,10 @@
     (values
       (for/fold ([Γ (replace-props env props)]) ([f (in-list atoms)])
         (match f
```
samth: I think the name `f` here no longer makes sense.

typed-racket-lib/typed-racket/typecheck/tc-metafunctions.rkt
```diff
@@ -74,12 +74,7 @@
         (let* ([p (car worklist)]
                [p (resolve derived-atoms p)])
           (match p
-            [(ImpFilter: a c)
```
samth: Where did implication go?

typed-racket-lib/typed-racket/typecheck/tc-subst.rkt
```diff
@@ -135,80 +135,48 @@
...
-    [(OrFilter: fs) (apply -or (map ap fs))]
-    [(Bot:) -bot]
-    [(Top:) -top]
-    [(TypeFilter: t (Path: p i))
-     (tf-matcher t p i -filter)]
-    [(NotTypeFilter: t (Path: p i))
-     (tf-matcher t p i -not-filter)]))
-
-;; Determine if the object k occurs free in the given type
-(define (index-free-in? k type)
```
samth: Was this function just unused?

typed-racket-lib/typed-racket/types/type-table.rkt
```diff
@@ -78,7 +78,7 @@
     [((tc-result1: old-t) (tc-result1: t-t))
      (ret (Un old-t t-t))]
     [((tc-results: old-ts) (tc-results: t-ts))
-     ;; filters don't matter at this point, since only
+     ;; props don't matter at this point, since only
```
samth: I think this comment is no longer true, so maybe add a note to that effect while you're here.

typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -3126,14 +3127,14 @@
                   (-> Symbol #:b Symbol Symbol)))
            (define z (lambda (a #:b b) a))
            (z "y" #:b "y"))
-         #:ret (ret -String -bot-filter)
-         #:expected (ret -String -no-filter -no-obj)]
+         #:ret (ret -String -ff-propset)
+         #:expected (ret -String #f #f)]
```
samth: Why did these grow more arguments?

samth:

typed-racket-lib/typed-racket/typecheck/tc-metafunctions.rkt
andmkent: It was only ever introduced by TR in ```let``` expressions, and we did away with that last year, so I figured it was okay to scrub them from the code base

typed-racket-lib/typed-racket/typecheck/tc-subst.rkt
andmkent: It was used, but only at that one spot above, and I've never been able to figure out why. Substitution shouldn't be that eager about obliterating terms just because an index is free

typed-racket-test/unit-tests/typecheck-tests.rkt
andmkent: ... huh? I'm either not seeing what your talking about or perhaps you miscounted? I see no differing numbers of arguments here across the diff.

samth: Yeah, I just miscounted.

typed-racket-lib/typed-racket/typecheck/tc-metafunctions.rkt
samth: Ok, seems good.

typed-racket-lib/typed-racket/typecheck/tc-subst.rkt
samth: I'd have to think more about this code, but I worry about taking out that line. When did it go in, and why is it safe to do this if `k` is free in `t`?

andmkent: I had a couple of conversations about that last year and I think we concluded basically "I don't know why we do that" -- It's in the ICFP '10 paper as well, but there's no insight into why (and in fact I think as we examined it it seemed more sensible not to).

I commented it out of the RTR prototype ages ago and never had an issue:
https://github.com/andmkent/typed-racket/blob/pldi2016/typed-racket-lib/typed-racket/typecheck/tc-subst.rkt#L244

samth: Looking at the paper again (relevant figure is Figure 8, clauses where `x \neq z \and z \in fv(\nu)`), I think I understand the issue better now.

The problem is when `\nu` ends up referring to `x` which is no longer in scope. This means that it could end up in the scope of some _other_ `x`, which would be very bad (or even the same `x`, reinstantiated? maybe?). However, the problem if it ending up in the scope of another `x` can't happen here, because we don't just use symbols for variables. So I think this is ok. But maybe a comment along these lines belongs in the relevant code.

andmkent:
closed via f9c5a534d03de43569bd9719edb80531235a4eaf

PR #350, opened 2016-04-26
initial intersection types addition

andmkent:
Adds intersection types as a better to handle the the case when restrict cannot intersect two types, allowing for non-lossy refinements of type info.

### DIFF
--- a/typed-racket-lib/typed-racket/infer/infer-unit.rkt
+++ b/typed-racket-lib/typed-racket/infer/infer-unit.rkt
@@ -19,7 +19,7 @@
          "constraint-structs.rkt"
          "signatures.rkt" "fail.rkt"
          "promote-demote.rkt"
-         racket/match
+         racket/match racket/set
          mzlib/etc
          (contract-req)
          (for-syntax
@@ -525,6 +525,19 @@
           [((? Mu? s) t) (cg (unfold s) t)]
           [(s (? Mu? t)) (cg s (unfold t))]
 
+          ;; find *an* element of elems which can be made a subtype of T
+          [((Intersection: ts) T)
+           (cset-join
+            (for*/list ([t (in-immutable-set ts)]
+                        [v (in-value (cg t T))]
+                        #:when v)
+              v))]
+          
+          ;; constrain S to be below *each* element of elems, and then combine the constraints
+          [(S (Intersection: ts))
+           (define cs (for/list/fail ([ts (in-immutable-set ts)]) (cg S ts)))
+           (and cs (cset-meet* (cons empty cs)))]
+          
           ;; constrain *each* element of es to be below T, and then combine the constraints
           [((Union: es) T)
            (define cs (for/list/fail ([e (in-list es)]) (cg e T)))
--- a/typed-racket-lib/typed-racket/infer/restrict.rkt
+++ b/typed-racket-lib/typed-racket/infer/restrict.rkt
@@ -12,8 +12,7 @@
 
 
 ;; restrict t1 to be a subtype of t2
-;; if `pref' is 'new, use t2 when giving up, otherwise use t1
-(define (restrict t1 t2 [pref 'new])
+(define (restrict t1 t2)
   ;; build-type: build a type while propogating bottom
   (define (build-type constructor . args)
     (if (memf Bottom? args) -Bottom (apply constructor args)))
@@ -21,7 +20,7 @@
   ;; (i.e. pairs of t1 t2) to prevent infinite unfolding.
   ;; subtyping performs a similar check for the same
   ;; reason
-  (define (restrict* t1 t2 pref resolved)
+  (define (restrict* t1 t2 resolved)
     (match* (t1 t2)
       ;; already a subtype
       [(_ _) #:when (subtype t1 t2) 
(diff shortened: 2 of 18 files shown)


samth:
Again, more informative commit message please.

typed-racket-lib/typed-racket/infer/infer-unit.rkt
```diff
@@ -525,6 +525,19 @@
           [((? Mu? s) t) (cg (unfold s) t)]
           [(s (? Mu? t)) (cg s (unfold t))]
 
+          ;; find *an* element of elems which can be made a subtype of T
+          [((Intersection: elems) T)
+           (cset-join
+            (for*/list ([elem (in-set elems)]
```
samth: Let's use `in-hash-set` here (and elsewhere).

typed-racket-lib/typed-racket/infer/restrict.rkt
```diff
@@ -48,6 +48,14 @@
       ;; unions
       [((Union: t1s) _) (apply Un (map (λ (t1*) (restrict* t1* t2 pref resolved)) t1s))]
       [(_ (Union: t2s)) (apply Un (map (λ (t2*) (restrict* t1 t2* pref resolved)) t2s))]
+
+      ;; intersections
+      [((Intersection: t1s) _)
+       (apply -intersect (for/list ([t1 (in-set t1s)])
```
samth: Can we use `for/set` here?

typed-racket-lib/typed-racket/infer/restrict.rkt
```diff
@@ -63,6 +71,6 @@
       [(_ _) #:when (not (overlap t1 t2)) 
              (Un)]
       
-      ;; t2 and t1 have a complex relationship, so we punt
-      [(_ _) (if (eq? pref 'new) t2 t1)]))
```
samth: `pref` is no longer needed, right?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -385,6 +385,8 @@
          (if numeric-sc
              (apply or/sc numeric-sc (map t->sc non-numeric))
              (apply or/sc (map t->sc elems)))]
+        [(Intersection: elems)
+         (apply and/sc (for/list ([elem (in-set elems)]) (t->sc elem)))]
```
samth: Does this really work for complex intersections (for example, polymorphic types and others, or two different function types, etc)?

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -449,6 +451,51 @@
...
+         [(= 1 (set-count t-set)) (set-first t-set)]
+         [else (*Intersection t-set)])]
+      [(cons t ts)
+       (cond
+         [(Bottom? t) t]
+         [(Univ? t) (loop elems ts)]
+         [(Intersection? t)
+          (loop elems (append (set->list (Intersection-elems t)) ts))]
+         [(for/or ([elem (in-list elems)]) (not (overlap elem t)))
+          (*Union (list))]
```
samth: This is just `-bot`, right?

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -449,6 +451,51 @@
...
+(define (-intersect . ts)
+  (let loop ([elems (list)]
+             [ts ts])
+    (match ts
+      [(list)
+       (define t-set (list->set elems))
+       (cond
+         [(set-empty? t-set) (Univ)]
+         [(= 1 (set-count t-set)) (set-first t-set)]
+         [else (*Intersection t-set)])]
```
samth: The set creation should go in the else clause, the other two can be tested on `elems`.

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -449,6 +451,51 @@
...
+      [(cons t ts)
+       (cond
+         [(Bottom? t) t]
+         [(Univ? t) (loop elems ts)]
+         [(Intersection? t)
+          (loop elems (append (set->list (Intersection-elems t)) ts))]
```
samth: Can we maintain the accumulator as a set?

typed-racket-lib/typed-racket/types/remove-intersect.rkt
```diff
@@ -106,6 +109,9 @@
            old]
           [(list (Union: l) rem)
            (apply Un (map (lambda (e) (*remove e rem)) l))]
+          [(list (Intersection: elems) rem)
```
samth: Let's name the contents of `Intersection` consistently.

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -272,6 +273,16 @@
          ;; value types
          [((Value: v1) (Value: v2))
           #:when (equal? v1 v2) A0]
+         [((Intersection: elems) t)
```
samth: Why don't we need to thread through `A` in these cases?

samth:
Needs more tests, especially for the `restrict` use case that this is aimed at.

typed-racket-lib/typed-racket/infer/infer-unit.rkt
andmkent: neat! (and it's ```in-immutable-set```, for any future readers who are curious)

typed-racket-lib/typed-racket/infer/restrict.rkt
andmkent: I don't think so, since -intersect needs to briefly iterate over the arguments... it seems more sensible to just apply it to the list, right?

typed-racket-lib/typed-racket/infer/restrict.rkt
andmkent: correct, nice!

typed-racket-lib/typed-racket/private/type-contract.rkt
andmkent: I'll have to test this and think about it more --- I do not know.

typed-racket-lib/typed-racket/rep/type-rep.rkt
andmkent: there's no abbreviation for the bottom type at this point to my knowledge

andmkent:
Still need to:

1. add more tests

2. ```-intersect``` is probably doing to much work? Or at least ```restrict``` shouldn't have to call it (since both ```restrict``` and ```interesct``` are doing things like checking for overlap, etc.)

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -385,6 +394,24 @@
          (if numeric-sc
              (apply or/sc numeric-sc (map t->sc non-numeric))
              (apply or/sc (map t->sc elems)))]
+        [(Intersection: ts)
```
andmkent: I think this code is correct, but it was built by iteratively throwing things at the wall and seeing what stuck. Would appreciate a sanity check here.

andmkent:
I believed I've addressed all the issues. Let me know if there's anything else we want to change.

typed-racket-lib/typed-racket/infer/restrict.rkt
```diff
@@ -34,35 +33,40 @@
...
       [(_ (? needs-resolving?)) #:when (not (set-member? resolved (cons t1 t2)))
-                                (restrict* t1 (resolve t2) pref (set-add resolved (cons t1 t2)))]
+                                (restrict* t1 (resolve t2) (set-add resolved (cons t1 t2)))]
       
       ;; we don't actually want this - want something that's a part of t1
       [(_ _) #:when (subtype t2 t1)
              t2]
-      
-      ;; there's no overlap, so the restriction is empty
-      [(_ _) #:when (not (overlap t1 t2))
```
samth: Why are we dropping this case?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -385,6 +394,24 @@
...
+         (define-values (chaperons/impersonators others)
```
samth: Spell chaperone right. :)

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -385,6 +394,24 @@
...
+           (for/fold ([cs/is null] [others null])
+                     ([elem (in-immutable-set ts)])
+             (define c (t->sc elem))
+             (match (get-max-contract-kind c)
+               [(== flat-sym) (values cs/is (cons c others))]
+               [(or (== chaperone-sym)
```
samth: Just make this the else case.

typed-racket-lib/typed-racket/infer/restrict.rkt
andmkent: Building an intersection will produce bot if there is no overlap, so this check is redundant.

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -449,6 +451,55 @@
...
+         ;; size > 1, build an intersection
+         [else (*Intersection elems)])]
+      [(cons t ts)
+       (match t
+         [(? Bottom?) t]
+         [(Univ:) (loop elems ts)]
+         [(Intersection: ts*) (loop (set-union elems ts*) ts)]
+         [t (cond
+              [(for/or ([elem (in-immutable-set elems)]) (not (overlap elem t)))
+               (*Union (list))]
```
samth: Why is there a `Union` here?

samth:
Why is the only interface called `-unsafe-intersect`? Is there a safe interface that the rest of TR should use?

typed-racket-lib/typed-racket/rep/type-rep.rkt
andmkent: The intersection of non-overlapping types produces bottom in that line (which is if course an empty union).

andmkent:
@samth restrict is the more thorough way to produce an intersection (it will check lots more possibilities), while -unsafe-intersect only checks overlap and if Top is present.

typed-racket-lib/typed-racket/rep/type-rep.rkt
samth: Ah ok. Don't we have a constant for that?

samth:
Should we rename `restrict` to `intersect` then, now that it's symmetric?

typed-racket-lib/typed-racket/rep/type-rep.rkt
andmkent: Not in this file (because of our somewhat messy dependency tree) -- it's defined in base-abbrevs.rkt

andmkent:
calling restrict intersect w/ these changes makes sense to me

samth:
Ok, with that renaming, everything else looks good to me. Also, feel free to merge this and then rename in a separate commit.

PR #353, opened 2016-05-08
Add typed/syntax/modread, typed/syntax/srcloc, and typed/syntax/readerr

AlexKnauth:
Typed versions of syntax/modread, syntax/srcloc, and syntax/readerr.

### DIFF
--- a/typed-racket-more/typed/syntax/modread.rkt
+++ b/typed-racket-more/typed/syntax/modread.rkt
@@ -1,0 +1,22 @@
+#lang typed/racket/base
+
+(provide with-module-reading-parameterization check-module-form)
+
+(require typed/racket/unsafe)
+
+(unsafe-require/typed syntax/modread
+                      [with-module-reading-parameterization
+                       (All (A) (-> (-> A) A))]
+                      [check-module-form
+                       ;; check-module-form is already protected with the contract:
+                       ;; (-> (or/c syntax? eof-object?)
+                       ;;     (or/c symbol? list?)
+                       ;;     (or/c string? path? false/c)
+                       ;;     any) ; This is any because the contract doesn't need
+                       ;;          ; to enforce that it's (or/c syntax? false/c).
+                       ;;                             ; From the documentation:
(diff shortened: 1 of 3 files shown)


samth:
Looks great. Why the use of `unsafe` in one place but not the others?

AlexKnauth:
Because both the `(Syntaxof Any)` and `Syntax` contracts weren't working for some reason.

samth:
Right, `(Syntaxof Any)` is tricky, as are syntax contracts in general. But there are other files in this PR that use those contracts; I'm wondering why the difference.

AlexKnauth:
I don't know why. Your right, it's very weird.

AlexKnauth:
Although it might have something to do with whether its in a positive or negative position.

AlexKnauth:
It also might have something to do with whether or not it's in a union, although the contracts in that case might be unsound? Yes, it turns out is, but that's a separate issue.

AlexKnauth:
So, once https://github.com/racket/typed-racket/issues/354 is fixed, `typed/syntax/srcloc` will probably have to use `unsafe-require/typed` as well.

AlexKnauth:
If it's handled in the same way that `(-> (Syntaxof Any) Any)` is handled right now, that is.

AlexKnauth:
So, should I translate `typed/syntax/srcloc` to use `unsafe-require/typed`?

samth:
Which programs would work/not work, depending on what you do?

AlexKnauth:
If I leave it as it is, than a fix for #354 would break this. If I change it to use unsafe, then everything would work, but it wouldn't be protected with a contract, but that's fine because typed code won't pass anything unsafe into it, untyped code won't use the unguarded typed values in any unsafe ways, and we can trust the untyped code to return what it promised just like we can for the other trusted racket libraries.

samth:
In that case, let's use unsafe.

AlexKnauth:
Ok, I've updated it to use `unsafe-require/typed` for both `typed/syntax/modread` and `typed/syntax/srcloc`.

samth:
LGTM

PR #360, opened 2016-05-24
Add All-bounded type form for upper-bounded polymorphism

AlexKnauth:
This defines an `All-bounded` form that uses intersection types to simulate upper-bounded polymorphism.

Types like `(All-bounded ([A <: T]) [A -> A])` de-sugar to types like `(All (A) [(∩ A T) -> (∩ A T)])`.

One issue is that this doesn't actually restrict instantiation. When someone instantiates `(All-bounded ([A <: T]) [A -> A])` with a type `X` that is not a subtype of `T`, the instantiation succeeds and still results in `(∩ X T) -> (∩ X T)` instead of raising a type error.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-types-extra.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-types-extra.rkt
@@ -20,7 +20,7 @@
 
 ;; special type names that are not bound to particular types
 (define-other-types
-  -> ->* case-> U Rec All Opaque Vector
+  -> ->* case-> U Rec All All-bounded Opaque Vector
   Parameterof List List* Class Object Unit Values AnyValues Instance Refinement
   pred Struct Struct-Type Prefab Top Bot Distinction Sequenceof
   ∩)
--- a/typed-racket-lib/typed-racket/private/parse-type.rkt
+++ b/typed-racket-lib/typed-racket/private/parse-type.rkt
@@ -5,7 +5,7 @@
 (require (rename-in "../utils/utils.rkt" [infer infer-in])
          (except-in (rep type-rep object-rep) make-arr)
          (rename-in (types abbrev union utils prop-ops resolve
-                           classes prefab signatures)
+                           classes prefab signatures substitute)
                     [make-arr* make-arr])
          (only-in (infer-in infer) intersect)
          (utils tc-utils stxclass-util literal-syntax-class)
@@ -97,6 +97,7 @@
 (define-literal-syntax-class #:for-label Rec)
 (define-literal-syntax-class #:for-label U)
 (define-literal-syntax-class #:for-label All)
+(define-literal-syntax-class #:for-label All-bounded)
 (define-literal-syntax-class #:for-label Opaque)
 (define-literal-syntax-class #:for-label Parameter)
 (define-literal-syntax-class #:for-label Vector)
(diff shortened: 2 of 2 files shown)


samth:
Some high-level thoughts:

1. Restricting instantiation is necessary for this to be reasonable for users.
2. I think the eventual version of this should just be a change to `All`.
3. We should think about how this interacts with poly-dots.
4. What's the contract generation story here? I expect @thinkmoore to have some thoughts.

More thoughts later.

AlexKnauth:
For restricting instantiation, does it need to be enforced for inferred instantiations as well, or just for explicit user-written instantiations? If an instantiation is being inferred, then it would make sense for the type-checker to just infer the intersection type anyway. So would it make sense to only enforce it for explicit `inst` forms?

andmkent:
@AlexKnauth It sounds dangerous to have an invariant/bound that we are not enforcing uniformly, doesn't it? Perhaps w/ some examples that would require inference with these types we could more easily figure out what makes sense. Also, have we looked into what systems with similar features do in these cases (I have not yet).

AlexKnauth:
If we were enforcing it uniformly, then then type-checker would get around it by just putting the intersection type on it. Since this doesn't support lower bounds, that would always work. And that's what it's already doing, except that the intersection types are put there when the type is parsed instead of when it's instantiated.

AlexKnauth:
Examples and invariants:

One example is this program, modified from [this Stack Overflow question](https://stackoverflow.com/questions/29898505/bounded-parameterized-types-with-typed-racket):
```racket
#lang typed/racket

(struct base ([x : Real]))
(struct child base ([y : String]))

(: fun : (All-bounded ([A <: base]) [A A → A]))
(define (fun a b)
  (if (> (base-x a) (base-x b))
      a
      b))
```
The invariant here is that these places in the type:
```racket
[A A -> A]
-^-^----^-
```
Will never be filled with types that are not a subtype of `base`. Or, `A` <: `base`. And that's going to be true whether we restrict instantiation or not, because it will actually be this:
```racket
[(∩ B base) (∩ B base) -> (∩ B base)]
-^^^^^^^^^^-^^^^^^^^^^----^^^^^^^^^^-
```
Where the invariant is still `A` <: `base`. But here that means `(∩ B base)` <: `base`, which is true no matter what `B` is. If there are no explicit instantiations, and all of them are inferred, the type-checker can infer `B` = `(U Integer child)` if it wants, and then `A` will be `(∩ (U Integer child) base)`, equivalent to `child`, which is what the type-checker would have inferred if it had tried to directly infer `A` with the restriction. 

However, for explicit instantiations, if someone writes `(inst f (U Integer child))`, they will think they are setting `A` = `(U Integer child)`, when actually they are setting `B` = `(U Integer child)`, so that `A` = `(∩ (U Integer child) base)`. That would be confusing, so it does make sense to restrict explicit instantiations, so that `(inst f (U Integer child))` will raise a type error just as if they were trying to set `A` equal to that.

jackfirth:
Minor nit: Should it be `All-Bounded` instead of `All-bounded` in keep with multi-word types like `Positive-Integer`?

lexi-lambda:
Substantive, non-bikeshedding discussions aside, I agree with @samth that this sort of thing would be better as an extension to `All` rather than a separate type (at least in the user-facing sense).

thinkmoore:
As @samth mentioned, I've been working on an implementation for bounded-polymorphic contracts. I have a prototype now but I'm currently working on figuring out exactly the best way forward to integrate with the existing contract and chaperone systems. @AlexKnauth let's keep in touch as I try and push that forward.

AlexKnauth:
@thinkmoore 
Ok. Right now this just desugars into intersection types, so the contracts would probably end up having `and/c` contracts as inputs and outputs, and depending on the order of how those are applied and how inputs are wrapped those could fail in weird ways. So for contract generation I think we will need true bounded-polymorphic contracts.

samth:
I think bounded-parametric contracts should be pretty easy to add -- just provide a contract along with each type variable, roughly, and check it first before sealing.

Also, contract generation support in the positive direction is easy -- `(All (A <: T) (A -> A))` should be `(-> T/c T/c)`.

thinkmoore:
I don't think that this example is correct, since it does not constrain the A's in the function signature to be related. To be more concrete, given the following code:

```
(struct base ([x : Real]))
(struct child base ([y : String]))

(define (foo a) (base 0))
```
function `foo` satisfies the contract `(-> base/c base/c)`, but the following code will yield a runtime error that blames the wrong party:
```
(let ([an-a (foo (child 42 "hello"))])
  (string-length (child-y an-a))
```

I have a prototype implementation that deals with these issues, but it uses some hacks to the impersonator system that I don't think we want.

samth:
Your `foo` wouldn't type check with that type -- the translation I gave in the second half of my comment is for typed->untyped boundaries only. For the other direction you need `base?` + sealing.

thinkmoore:
Good catch. Agreed. Unfortunately sealing for the negative part is where all the "fun" is...

samth:
Why doesn't checking `base?` and then sealing work?

PR #362, opened 2016-05-24
fix contract related bugs

andmkent:
The bug in tc-struct.rkt has been there for a while but only occurs when TR contracts are on.

I think we should look into making the TR contracts a part of our automated Travis-CI testing.


samth:
Looks good.

Issue #108, opened 2015-04-09
TR: casting makes type checking unsound for mutable boxes/vectors/structs


philnguyen:
###### Originally submitted on: Fri Jan 30 21:20:01 -0500 2015

I'm not familiar with this problem. Let me guess the right fix: when we cast (Boxof A) to (Boxof B), should we wrap around it a contract ensuring anything going into the box as type A, (in addition to ensuring anything going out of the box has type B, which TR already does)?

##### Steps to Reproduce:

The following 3 expressions all produce "hi" typechecked as Integer:

```racket
(begin
  (define b1 : (Boxof Integer) (box 42))
  (define b2 (cast b1 (Boxof String)))
  (set-box! b2 "hi")
  (unbox b1))

(begin
  (define v1 : (Vectorof Integer) (vector 42))
  (define v2 (cast v1 (Vectorof String)))
  (vector-set! v2 0 "hi")
  (vector-ref v1 0))

(begin
  (struct (X) s ([i : X]) #:mutable)
  (define s1 : (s Integer) (s 42))
  (define s2 (cast s1 (s String)))
  (set-s-i! s2 "hi")
  (s-i s1))
```

##### Release:

```
6.1.1.7--2015-01-17(2b8acb3/a)
```

##### Environment:

```
unix "Linux ubuthink 3.16.0-28-generic #38-Ubuntu SMP Fri Dec 12 17:37:40 UTC 2014 x86_64
 x86_64 x86_64 GNU/Linux" (x86_64-linux/3m) (get-display-depth) = 32
Human Language: english
(current-memory-use) 333255660
raco pkg (show):
Installation-wide:
 Package            Checksum     Source
 main-distribution  b660f028...  catalog main-distribution
 racket-lib         e93899ff...  catalog racket-lib
 [185 auto-installed packages not shown]
User-specific for installation "snapshot":
 Package        Checksum              Source
 rmacs          4bef6487...           catalog...acs/master
 soft-contract                        link...soft-contract
 [1 auto-installed package not shown]



Collections:
("/home/pcn/.racket/snapshot/collects"
 (non-existent-path))
("/home/pcn/bin/Racket/collects"
 (".gitignore" "acks" "compiler" "data" "db" "dynext" "ffi" "file" "info" "info-domain"
 "json" "launcher" "net" "openssl" "pkg" "planet" "racket" "rackunit" "raco" "reader"
[...]

PR #365, opened 2016-05-26
Close cast soundness loophole

AlexKnauth:
Fixes https://github.com/racket/typed-racket/issues/108
Fixes PR 13626
Fixes PR 14955

This makes `cast` generate two contracts, one from typed to untyped using the original type of the expression, and one from untyped to typed, using the type it's casting to. 

So in examples like this:
```racket
#lang typed/racket

(: v Boolean)
(define v #f)
(: f (Boolean -> Boolean))
(define (f x)
  (set! v x)
  x)

v ; #f
((cast f (String -> String)) "hello") ; error!
v ; if that hadn't been an error, this would be "hello" with type Boolean
```
The cast applies a contract for the new type `String -> String` going from untyped to typed, but first it applies a contract for the old type, `Boolean -> Boolean`, going from typed to untyped. This also means that doing `((cast f (Boolean -> Boolean)) #t)` succeeds and sets `v` to `#t`, because it checked that the argument was a boolean first.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -75,6 +75,7 @@
          (for-template "../utils/any-wrap.rkt")
          "../utils/tc-utils.rkt"
          "../private/syntax-properties.rkt"
+         "../private/cast-table.rkt"
          "../typecheck/internal-forms.rkt"
          ;; struct-extraction is actually used at both of these phases
          "../utils/struct-extraction.rkt"
@@ -250,9 +251,27 @@
 ;;  make-predicate
 ;;  cast
 
-;; Helper to construct syntax for contract definitions
+;; Helpers to construct syntax for contract definitions
+;; make-contract-def-rhs : Type-Stx Boolean Boolean -> Syntax
 (define (make-contract-def-rhs type flat? maker?)
-  (contract-def-property #'#f `#s(contract-def ,type ,flat? ,maker? untyped)))
+  (define contract-def `#s(contract-def ,type ,flat? ,maker? untyped))
+  (contract-def-property #'#f (λ () contract-def)))
+
+;; make-contract-def-rhs/from-typed : Id Boolean Boolean -> Syntax
+(define (make-contract-def-rhs/from-typed id flat? maker?)
+  (contract-def-property
+   #'#f
+   ;; This function should only be called after the type-checking pass has finished.
+   ;; By then `tc/#%expression` will have recognized the `casted-expr` property and
+   ;; will have added the casted expression's original type to the cast-table, so
+   ;; that `(cast-table-ref id)` can get that type here.
+   (λ ()
+     (define type-stx
+       (or (cast-table-ref id)
+           (int-err (string-append
+                     "contract-def-property: thunk called too early\n"
+                     "  This should only be called after the type-checking pass has finished."))))
+     `#s(contract-def ,type-stx ,flat? ,maker? typed))))
 
 
 (define (define-predicate stx)
@@ -289,17 +308,17 @@
 (define (cast stx)
   (syntax-parse stx
     [(_ v:expr ty:expr)
-     (define (apply-contract ctc-expr)
+     (define (apply-contract v ctc-expr pos neg)
        #`(#%expression
           #,(ignore-some/expr
-             #`(let-values (((val) #,(with-type* #'v #'Any)))
+             #`(let-values (((val) #,(with-type* v #'Any)))
                  #,(syntax-property
                     (quasisyntax/loc stx
                       (contract
                        #,ctc-expr
                        val
-                       'cast
-                       'typed-world
+                       '#,pos
+                       '#,neg
                        val
                        (quote-srcloc #,stx)))
                     'feature-profile:TR-dynamic-check #t))
@@ -308,8 +327,15 @@
      (cond [(not (unbox typed-context?)) ; no-check, don't check
             #'v]
            [else
-            (define ctc (syntax-local-lift-expression
-                         (make-contract-def-rhs #'ty #f #f)))
+            (define new-ty-ctc (syntax-local-lift-expression
+                                (make-contract-def-rhs #'ty #f #f)))
+            (define existing-ty-id new-ty-ctc)
+            (define existing-ty-ctc (syntax-local-lift-expression
+                                     (make-contract-def-rhs/from-typed existing-ty-id #f #f)))
+            (define (store-existing-type existing-type)
+              (cast-table-set! existing-ty-id existing-type))
+            (when (equal? (syntax-local-context) 'top-level)
+              (store-existing-type #'Any))
             (define (check-valid-type _)
               (define type (parse-type #'ty))
               (define vars (fv type))
@@ -319,7 +345,12 @@
                  "Type ~a could not be converted to a contract because it contains free variables."
                  type)))
             #`(#,(external-check-property #'#%expression check-valid-type)
-               #,(apply-contract ctc))])]))
+               #,(apply-contract
+                  (apply-contract
+                   #`(#,(casted-expr-property #'#%expression store-existing-type)
+                      v)
+                   existing-ty-ctc 'typed-world 'cast)
+                  new-ty-ctc 'cast 'typed-world))])]))
 
 
 
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -19,7 +19,7 @@
 
 (provide require/opaque-type require-typed-struct-legacy require-typed-struct
          require/typed-legacy require/typed require/typed/provide
-         require-typed-struct/provide cast make-predicate define-predicate
+         require-typed-struct/provide core-cast make-predicate define-predicate
          require-typed-signature)
 
 (module forms racket/base
@@ -31,7 +31,7 @@
                     require-typed-struct-legacy
                     require-typed-struct
                     require/typed-legacy require/typed require/typed/provide
-                    require-typed-struct/provide cast make-predicate define-predicate)]))
+                    require-typed-struct/provide core-cast make-predicate define-predicate)]))
   (define-syntax (def stx)
     (syntax-case stx ()
       [(_ id ...)
@@ -43,7 +43,16 @@
         require-typed-struct-legacy
         require-typed-struct
         require/typed-legacy require/typed require/typed/provide
-        require-typed-struct/provide cast make-predicate define-predicate))
+        require-typed-struct/provide make-predicate define-predicate)
+
+  ;; Expand `cast` to a `core-cast` with an extra `#%expression` in order
+  ;; to prevent the contract generation pass from executing too early
+  ;; (i.e., before the `cast` typechecks)
+  (define-syntax (-core-cast stx) (core-cast stx))
+  (define-syntax (cast stx)
+    (syntax-case stx ()
+      [(_ e ty) (quasisyntax/loc stx (#%expression #,(syntax/loc stx (-core-cast e ty))))]))
+  (provide cast))
 
 ;; unsafe operations go in this submodule
 (module* unsafe #f
@@ -304,8 +313,8 @@
      #`(#,(external-check-property #'#%expression check-valid-type)
         #,(ignore-some/expr #`(flat-contract-predicate #,name) #'(Any -> Boolean : ty)))]))
 
-
-(define (cast stx)
+;; wrapped above in the `forms` submodule
+(define (core-cast stx)
   (syntax-parse stx
     [(_ v:expr ty:expr)
      (define (apply-contract v ctc-expr pos neg)
@@ -334,8 +343,6 @@
                                      (make-contract-def-rhs/from-typed existing-ty-id #f #f)))
             (define (store-existing-type existing-type)
               (cast-table-set! existing-ty-id existing-type))
-            (when (equal? (syntax-local-context) 'top-level)
-              (store-existing-type #'Any))
             (define (check-valid-type _)
               (define type (parse-type #'ty))
               (define vars (fv type))
--- a/typed-racket-lib/typed-racket/typecheck/tc-expression.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-expression.rkt
@@ -34,6 +34,14 @@
      (if expected
         (tc-expr/check #'e expected)
         (tc-expr #'e))]
+    [(exp:casted-expr^ e)
+     (define result (tc-expr #'e))
+     (match result
+       [(tc-result1: ty)
+        ((attribute exp.value) ty)
+        result]
+       [_
+        (tc-error/expr "Cannot cast expression that produces multiple values")])]
     [(_ e)
      (if expected
          (tc-expr/check #'e expected)
--- a/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
@@ -541,9 +541,18 @@ returned by @racket[e], protected by a contract ensuring that it has type
 
 @ex[(cast 3 Integer)
 (eval:error (cast 3 String))
-(cast (lambda: ([x : Any]) x) (String -> String))
+(cast (lambda ([x : Any]) x) (String -> String))
+((cast (lambda ([x : Any]) x) (String -> String)) "hello")
 ]
-}
+
+The value is actually protected with two contracts. The second contract checks
+the new type, but the first contract is put there to enforce the old type, to
+protect higher-order uses of the value.
+
+@ex[
+((cast (lambda ([s : String]) s) (Any -> Any)) "hello")
+(eval:error ((cast (lambda ([s : String]) s) (Any -> Any)) 5))
+]}
 
 @defform*[[(inst e t ...)
            (inst e t ... t ooo bound)]]{
--- a/typed-racket-lib/typed-racket/private/cast-table.rkt
+++ b/typed-racket-lib/typed-racket/private/cast-table.rkt
@@ -1,0 +1,25 @@
+#lang racket/base
+
+(provide cast-table-ref
+         cast-table-set!)
+
+(require syntax/id-table)
+
+;; A module that helps store information about the original types of casted
+;; expressions.
+;;
+;; Casts in Typed Racket must generate two contracts. One from typed to untyped,
+;; and another from untyped to typed. The contract from typed to untyped is
+;; generated based on the existing type of the expression, which must be stored
+;; in this table so that it can be looked up later in the contract-generation
+;; pass.
+
+(define cast-table (make-free-id-table))
+
+;; cast-table-set! : Id Type-Stx -> Void
+(define (cast-table-set! id type-stx)
+  (free-id-table-set! cast-table id type-stx))
+
+;; cast-table-ref : Id -> (U False Type-Stx)
+(define (cast-table-ref id)
+  (free-id-table-ref cast-table id #f))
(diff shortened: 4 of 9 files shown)


samth:
Wooo!

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -308,8 +316,15 @@
      (cond [(not (unbox typed-context?)) ; no-check, don't check
             #'v]
            [else
-            (define ctc (syntax-local-lift-expression
-                         (make-contract-def-rhs #'ty #f #f)))
+            (define existing-ty-sym (gensym))
+            (define new-ty-ctc (syntax-local-lift-expression
+                                (make-contract-def-rhs #'ty #f #f)))
+            (define existing-ty-ctc (syntax-local-lift-expression
+                                     ;; TODO: This should be changed to somehow use the
```
takikawa: This TODO comment is now unneeded right?

takikawa:
Thanks! Can you add some test cases, preferably including the ones from those issues/PRs? Otherwise looks good to me.

takikawa:
Oh, one more thing: we should make sure this doesn't break too many programs before shipping it in a release. For example by trying it on TR-written packages on the package catalog. If it does break a bunch of things, we should at least mention the backwards incompatibility on the mailing list.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -250,10 +251,17 @@
...
 ;;  cast
 
-;; Helper to construct syntax for contract definitions
+;; Helpers to construct syntax for contract definitions
+;; make-contract-def-rhs : Type-Stx Boolean Boolean -> Syntax
 (define (make-contract-def-rhs type flat? maker?)
   (contract-def-property #'#f `#s(contract-def ,type ,flat? ,maker? untyped)))
 
+;; make-contract-def-rhs/from-typed : Symbol Boolean Boolean -> Syntax
+(define (make-contract-def-rhs/from-typed sym flat? maker?)
```
samth: Why not make this a version of `make-contract-def-rhs`?

samth: Also, can we just delay all `contract-def-property` constructions so that we have a uniform interface?

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
AlexKnauth: Thanks, fixed now.

AlexKnauth:
I've added the examples from PR 13626 and issue #108 as test cases.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -319,7 +332,12 @@
                  "Type ~a could not be converted to a contract because it contains free variables."
                  type)))
             #`(#,(external-check-property #'#%expression check-valid-type)
-               #,(apply-contract ctc))])]))
+               #,(apply-contract
+                  (apply-contract
+                   #`(#,(casted-expr-property #'#%expression store-existing-type)
```
samth: Why put this on the `#%expression` identifier, instead of on the whole form?

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -308,8 +316,13 @@
      (cond [(not (unbox typed-context?)) ; no-check, don't check
             #'v]
            [else
-            (define ctc (syntax-local-lift-expression
-                         (make-contract-def-rhs #'ty #f #f)))
+            (define existing-ty-sym (gensym))
+            (define new-ty-ctc (syntax-local-lift-expression
```
samth: Could we use an id table based on `new-ty-ctc` (or something else) instead of using a gensym?

typed-racket-lib/typed-racket/typecheck/tc-expression.rkt
```diff
@@ -34,6 +34,14 @@
...
         (tc-expr/check #'e expected)
         (tc-expr #'e))]
+    [(exp:casted-expr^ e)
+     (define result (tc-expr #'e))
+     (match result
+       [(tc-result1: ty)
+        ((attribute exp.value) ty)
+        result]
+       [_
+        (tc-error/expr "Cannot cast expression that produces multiple values")])]
```
samth: Is there a reason we can't support this?

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
AlexKnauth: I could make it an id-table that used `(generate-temporary)`, would that be better?

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -254,11 +254,12 @@
 ;; Helpers to construct syntax for contract definitions
 ;; make-contract-def-rhs : Type-Stx Boolean Boolean -> Syntax
 (define (make-contract-def-rhs type flat? maker?)
-  (contract-def-property #'#f `#s(contract-def ,type ,flat? ,maker? untyped)))
+  (define contract-def `#s(contract-def ,type ,flat? ,maker? untyped))
+  (contract-def-property #'#f (λ () contract-def)))
```
samth: Let's use promises instead of thunks.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
AlexKnauth: Because that's where the `tc/#%expression` function looks for it, and that's what it already does for the `external-check-property` above it.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
samth: Won't the identifier that you get from `syntax-local-lift-expression` be unique?

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
AlexKnauth: I would be little bit uncomfortable with that, because even though promises delay computation, it's usually for a different reason than this. Usually when you see a promise, a forcing it works in the same way that evaluating it in the original context did, and mutation isn't supposed to change that. In this case, we're delaying the computation specifically so that it can rely on a mutation that will happen later, so promises feel a bit wrong to me.

typed-racket-lib/typed-racket/typecheck/tc-expression.rkt
AlexKnauth: The [`contract`](http://docs.racket-lang.org/reference/attaching-contracts-to-values.html#%28form._%28%28lib._racket%2Fcontract%2Fprivate%2Fbase..rkt%29._contract%29%29) form only supports single values, right?

samth: We could implement something with `call-with-values`.

AlexKnauth: Yes. But that's complicated enough that it should be a separate pull request, right?

samth: Sure.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -250,9 +251,17 @@
 ;;  make-predicate
 ;;  cast
 
-;; Helper to construct syntax for contract definitions
+;; Helpers to construct syntax for contract definitions
+;; make-contract-def-rhs : Type-Stx Boolean Boolean -> Syntax
 (define (make-contract-def-rhs type flat? maker?)
-  (contract-def-property #'#f `#s(contract-def ,type ,flat? ,maker? untyped)))
+  (define contract-def `#s(contract-def ,type ,flat? ,maker? untyped))
+  (contract-def-property #'#f (λ () contract-def)))
```
AlexKnauth: From https://github.com/racket/typed-racket/pull/365#discussion_r64797153 which is an "outdated diff" for some reason:

@samth :
Let's use promises instead of thunks.

@AlexKnauth :
I would be little bit uncomfortable with that, because even though promises delay computation, it's usually for a different reason than this. Usually when you see a promise, a forcing it works in the same way that evaluating it in the original context would, and mutation isn't "supposed to" change that. In this case, we're delaying the computation specifically so that it can rely on a mutation that will happen later, so promises feel a bit wrong to me.

Does this make sense?

AlexKnauth: I just thought of something else. If it were a promise, than an "accidental" too-early force would cache the wrong result, and the contract-generation pass would generate a wrapping `Any` contract even though by that time the `tc/#%expression` has put the correct type in the table. Using lambda means it might recompute a few more things, but that gets rid of the danger that a too-early application will ruin the result for future function applications of the lambda.

samth: Can we make a too-early force or call an error?

AlexKnauth: No, because if type-checking somehow stops in a way that it never sees the `casted-expr` property, then it will seem like a too-early force, and in that case it makes sense to generate the wrapping `Any` contract as a fallback.

samth: Why would typechecking stop, but this still get forced?

AlexKnauth: I don't know, it might be impossible. I don't know very much about how type errors get reported here, how the contract generation phase works, or how the type-checker checks typed terms within `ignore-some` expressions, so I'm kind of paranoid. But still, lambda seems safer, even if promises are also safe in this case, the way the rest of the code is designed right now.

samth: Just a quick overview of how this works:

* When you report a type error and we're reporting multiple type errors, it gets stored in a list in a mutable variable.
* Then, after `pass2` in the type checker, we call `report-all-errors`, which will raise a syntax error if needed. That code is [here](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt#L409). 
* The contract generation happens after this, when we (generate the provides)[https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt#L445]. 

I'm ok with sticking with the `lambda`, but I'd like to make this an error if we call the thunk to early.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -250,9 +251,17 @@
...
+
+;; make-contract-def-rhs/from-typed : Id Boolean Boolean -> Syntax
+(define (make-contract-def-rhs/from-typed id flat? maker?)
+  (contract-def-property
+   #'#f
+   (λ () `#s(contract-def ,(or (cast-table-ref id) #'Any) ,flat? ,maker? typed))))
```
AlexKnauth: Re: https://github.com/racket/typed-racket/pull/365/files/357e4a12419fbc6af056a77710823bfc53b1a75e#r64914939

> I'd like to make this an error if we call the thunk too early

Does that mean making this `(or (cast-table-ref id) (error ...))`?

samth: Yes.

AlexKnauth: Hm. I tried that, and it passes the cast tests, both new and old, in a module, in the repl, and at the top level, but I'm getting that error when `special-forms.scrbl` is run, specifically because of these examples here: https://github.com/AlexKnauth/typed-racket/blob/close-cast-loophole/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl#L542

samth: Do those particular examples work at the repl? What about when just calling `eval` on them (in an appropriate namespace)?

AlexKnauth: Yes, they work at the repl, using `eval`, and using `(make-base-eval)` from `scribble/example` (both with the error and with the `Any` fallback).

AlexKnauth: Actually no, (maybe because I hadn't compiled it properly yet), now this program gives me the "thunk called too early" error:
```racket
#lang racket
(require scribble/example)
(define ev (make-base-eval))
(ev '(require typed/racket))
(ev '(cast 3 Integer))
```
Although it still works both at the repl and using `eval`.

AlexKnauth: It might be because of the sandboxed evaluators, because this also fails in the "thunk called too early" version:
```racket
#lang racket
(require racket/sandbox)
(define ev (call-with-trusted-sandbox-configuration
             (λ () (make-module-evaluator "#lang typed/racket"))))
(ev '(cast 3 Integer))
```

AlexKnauth: And in the `Any` fallback version, it's still generating the correct contract, so the cast-table mutation is happening correctly, and the contract-generation pass gets the right contract. So it must be something else that's calling the thunk. So in the `Any` fallback version, this:
```racket
#lang racket
(require racket/sandbox)
(define ev (call-with-trusted-sandbox-configuration
             (λ () (make-module-evaluator "#lang typed/racket"))))
(ev '((cast (lambda ([s : String]) s) (Any -> Any)) 5))
```
Generates the correct error:
```
#<procedure:val>: contract violation
  expected: String
  given: 5
  in: the 1st argument of
      (-> String any)
  contract from: typed-world
  blaming: cast
  at: eval:6.0
```
Where if it were getting the `Any`, it would raise a different error saying `Attempted to use a higher-order value passed as `Any` in untyped code: #<procedure:val>`. Since it's not that, we know that has gotten the correct type by the time the contract-generation pass happens.

samth: Ok, we should figure out where that error is coming from. What's the stack trace when you get the error?

AlexKnauth:
```
Internal Typechecker Error: contract-def-property: thunk called too early
  This should only be called after the type-checking pass has finished.
while typechecking:
here
originally:
here
  context...:
   /Users/Alex/Documents/GitHub/typed-racket/typed-racket-lib/typed-racket/utils/tc-utils.rkt:270:0: int-err
   /Users/Alex/Documents/GitHub/typed-racket/typed-racket-lib/typed-racket/base-env/prims-contract.rkt:268:3
   /Users/Alex/Documents/GitHub/typed-racket/typed-racket-lib/typed-racket/private/type-contract.rkt:82:0: generate-contract-def
   /Users/Alex/Documents/GitHub/typed-racket/typed-racket-lib/typed-racket/private/type-contract.rkt:195:0: change-contract-fixups
   /Users/Alex/Documents/GitHub/typed-racket/typed-racket-lib/typed-racket/typecheck/toplevel-trampoline.rkt:108:24: for-loop
   dots-loop
   /applications/racket/2016-05-23/tr/Racket v6.5.0.5/collects/racket/private/more-scheme.rkt:148:2: call-with-break-parameterization
   .../more-scheme.rkt:261:28
   /applications/racket/2016-05-23/tr/Racket v6.5.0.5/share/pkgs/sandbox-lib/racket/sandbox.rkt:843:5: loop
```
It seems like it's going though `generate-contract-def` and `change-contract-fixups`, which are supposed to be for the contract-generation phase, I think. Is there any chance that the contract-generation phase is happening before type-checking is finished?

samth: I think the trampoline might be doing the wrong thing. Maybe @takikawa has some idea here?

AlexKnauth: Should that be a different issue? To me it seems better to just have the `Any` fallback.

samth: I don't think the `Any` fallback makes sense. This seems to be just a bug, which we shouldn't just paper over.

AlexKnauth: It would be if it generated the wrong contract, yes. But it *is* generating the correct contract in the end.

samth: No, it's a bug regardless of what's happening in the end. The contract shouldn't be called there.

AlexKnauth: Ok. But it's still a much smaller bug than the one this fixes.

samth: Definitely. Hopefully @takikawa will have a suggestion soon.

AlexKnauth: I was adding some `printf`s and looking at it with @takikawa, we discovered that it work if we do `(identity (cast 3 Integer))`, `(define x (cast 3 Integer))` or any other nested expression context. Also for some reason the repl is broken now. It might be because the lifts create begins, and the trampoline does the contract-generation pass on the first expression of the begin (the lifted contract) before it does anything on the next expressions.

samth: Is the REPL broken currently, or with this PR?

AlexKnauth: It's broken with the versions where the thunk errors when it's called to early.

AlexKnauth: I've patched it up so that the repl works, sandboxed evaluation works, and the docs build now. What I did was change `cast` so that in a top-level context it puts `Any` in the cast-table to start with. But for casts that aren't in a top-level-context, it will be an internal type-checker error for the thunk to be called too early, but that doesn't cause any failures because for non-top-level casts the contract-generation happens at the correct time.

AlexKnauth: This was just fixed by @takikawa so that the `Any` isn't necessary after all, even at the top level. What he did was have `cast` just expand to `core-cast` within an `#%expression` form, so that `core-cast` will never actually be at the top level, even if `cast` is.

samth: So now we can just have it error when called too early?

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -250,9 +260,27 @@
...
+   ;; By then `tc/#%expression` will have recognized the `casted-expr` property and
+   ;; will have added the casted expression's original type to the cast-table, so
+   ;; that `(cast-table-ref id)` can get that type here.
+   (λ ()
+     (define type-stx
+       (or (cast-table-ref id)
+           (int-err (string-append
+                     "contract-def-property: thunk called too early\n"
+                     "  This should only be called after the type-checking pass has finished."))))
+     `#s(contract-def ,type-stx ,flat? ,maker? typed))))
```
AlexKnauth: Re: https://github.com/racket/typed-racket/pull/365/files/357e4a12419fbc6af056a77710823bfc53b1a75e#r65783232

> So now we can just have it error when called too early?

Yes, and that's what it does now.

AlexKnauth: It errors when it's called too early, and it never tries an `Any` fallback in any situation.

AlexKnauth:
I've just squashed all of my commits into one, which got rid of the reverts of reverts.

samth:
This looks great. Ready to merge!

AlexKnauth:
Should I merge it then?

PR #367, opened 2016-05-30
Give correct type to `in-port` when used with custom reader

cky:
Currently, `in-port` returns `(Sequenceof Any)` unconditionally, which is correct if the given read function is `read` (default value). However, `(in-port read-line)`, `(in-port read-char)`, etc. should have more specific types.


samth:
Looks good to me.

cky:
I just added a commit that strips out EOF from the sequence type, since the sequence ends rather than return EOF. Tested with: `read`, `read-line`, `read-char`, `read-byte`, and `read-json`.

PR #372, opened 2016-06-10
add type definition for (make-input-port/read-to-peek)

wargrey:


typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -1951,9 +1951,23 @@
 ;; Section 13.1.10.2
 [input-port-append  (->* (list Univ) -Input-Port -Input-Port)]
 
-;TODO write the type for this
-;It is fairly complicated and require events
-;make-input-port/read-to-peek
+[make-input-port/read-to-peek
+ (->opt Univ
+        (-> -Bytes (Un -Nat (-val eof) top-func (-evt -Zero)))
```
samth: Why is `top-func` correct here?

wargrey: I have that question too.
I just copied it from the definition of (make-input-port), and it works.
Maybe because userland Typed Racket will never touch that procedure directly (if the contract system also works for it), so far the `top-func` reduces the complex to read the type annotation for human readers.

So do we need to make both of them more clear?

BTW, why the `input-port` does not have something like _reconsume_?

samth: Sorry for the delay here. I think we should change the types to make it more clear, rather than using `top-func`.

samth: Also, I don't know the answer about _reconsume_.

samth:
This is good to merge once the issue with `top-func` is dealt with.

samth:
Looks good!

PR #374, opened 2016-06-15
Reimplement init-envs.rkt

takikawa:
This PR reimplements the type serialization in `init-envs.rkt` to avoid using pconvert (to avoid the library dependency and also because I think it's easier to understand without it). Also implements a few optimizations that cut a few percent off of zo sizes, up to maybe 10% on plot.

Also changes the debug printer to use the type serialization (easier to read?).


samth:
Wooo!

samth:

takikawa:
Merged with some commit squashing.

Issue #375, opened 2016-06-25
cast table sometimes has non-syntax values


samth:
The cast table (in `private/cast-table.rkt`) can sometimes contain values which are not `syntax?`. This currently breaks the `wrap` and `measures-with-dimensions` packages. The relevant call is line 351 in `prims-contract.rkt`, and the offending `existing-type` value seems to always be a `Type?`.

AlexKnauth:
Is that this call on [line 346](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/base-env/prims-contract.rkt#L346)? When I wrote `Type-Stx` in the signatures in cast-table.rkt, I meant it in the sense of `Path-String`, where it's either a type or a syntax object representing a type. It might make sense to make it never a syntax object, always a proper type. 

The `contract-def` struct here on [line 248](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/base-env/prims-contract.rkt#L284), is it supposed to only take syntax objects? I see it's passed to `parse-type` [here](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/private/type-contract.rkt#L85). Should that be changed so that it only calls `parse-type` if it's not already a type?

samth:
The contract -def struct indeed expects only syntax objects. This error comes from that syntax object being used in an error message constructed in type-contract.rkt.

AlexKnauth:
Would wrapping it in 3d syntax ``#`#,`` fix this?

samth:
No, that doesn't seem right to me. We need syntax that points to the
correct place in the program.

AlexKnauth:
But this is a type that can be computed by the typechecker. Would it make sense to make the source location for the type syntax object point to the location of the expression that produced that type?

samth:
That seems like a good idea to me. Have you managed to reproduce this on your machine?

AlexKnauth:
Yes. I'll try wrapping it in a syntax object with the source location of ``#'v``.

AlexKnauth:
I tried wrapping it in a syntax object but the file in `measures-with-dimensions` was still failing, although with a different (better) error message. This program is a simplified version:
```racket
#lang typed/racket
(define v (ann (vector 1) VectorTop))
(cast v Any)
```
This succeeded before my `cast` fix, but fails now, with the error
```
[...]

PR #377, opened 2016-06-26
add (and/c vector? any-wrap/c) -like contracts

AlexKnauth:
For `VectorTop`, `HashTableTop`, etc. going from typed to untyped.

This fixes examples like
```racket
#lang typed/racket
(define v (ann (vector 1) VectorTop))
(cast v Any)
```
To generate `(and/c vector? any-wrap/c)` contracts instead of raising a compile-time contract-generation error. This reduces some of the backwards-incompatibility from the `cast` soundness fix.

This fixes some of the problems in https://github.com/racket/typed-racket/issues/375, although parts of ~~both measures-with-dimensions and~~ `wrap` are still failing for other reasons.

### DIFF
--- a/typed-racket-test/succeed/cast-mod.rkt
+++ b/typed-racket-test/succeed/cast-mod.rkt
@@ -44,31 +44,49 @@
   (: b1 : (Boxof Integer))
   (define b1 (box 42))
   (define b2 (cast b1 (Boxof String)))
+  (define b3 (ann b1 BoxTop))
+  (define b4 (cast b3 (Boxof (U Integer String))))
   (check-equal? (unbox b1) 42)
+  (check-equal? (unbox b4) 42)
   (check-exn #rx"expected: Integer\n *given: \"hi\""
              (λ () (set-box! b2 "hi")))
   (check-equal? (unbox b1) 42
-                "if the previous test hadn't errored, this would be \"hi\" with type Integer"))
+                "if the previous test hadn't errored, this would be \"hi\" with type Integer")
+  (check-exn #rx"Attempted to use a higher-order value passed as `Any`"
+             (λ () (set-box! b4 "hello")))
+  (check-equal? (unbox b1) 42))
 
 (test-case "cast on mutable vectors"
   (: v1 : (Vectorof Integer))
   (define v1 (vector 42))
   (define v2 (cast v1 (Vectorof String)))
+  (define v3 (ann v1 VectorTop))
+  (define v4 (cast v3 (Vectorof (U Integer String))))
   (check-equal? (vector-ref v1 0) 42)
+  (check-equal? (vector-ref v4 0) 42)
   (check-exn #rx"expected: Integer\n *given: \"hi\""
              (λ () (vector-set! v2 0 "hi")))
   (check-equal? (vector-ref v1 0) 42
-                "if the previous test hadn't errored, this would be \"hi\" with type Integer"))
+                "if the previous test hadn't errored, this would be \"hi\" with type Integer")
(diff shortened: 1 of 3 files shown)


samth:
Can we add a test based on the code from #375? Otherwise, this looks good to me.

samth:
Looks good once the tests pass.

AlexKnauth:
How does `any-wrap/c` deal with mutations to opaque structs? I'm getting a failing test for that here:
https://travis-ci.org/racket/typed-racket/builds/141172580#L346

samth:
I don't understand what the issue is.

`(contract any-wrap/c val a b)` should disallow writes to `val`, even in `val` is a mutable struct for which you don't currently have a controlling inspector.

What behavior are you seeing, and what should happen? Also, can you express that in terms of `contract` and `any-wrap/c` directly?

AlexKnauth:
I'm getting `"No exception raised"` from `check-exn` where the thunk tries to mutate an opaque mutable struct that should be protected with an `any-wrap/c` contract from a `cast` expression.

typed-racket-test/succeed/cast-mod.rkt
```diff
@@ -44,31 +44,49 @@
...
+  (define s4 (cast s3 (s (U Integer String))))
   (check-equal? (s-i s1) 42)
+  (check-equal? (s-i s4) 42)
   (check-exn #rx"expected: Integer\n *given: \"hi\""
              (λ () (set-s-i! s2 "hi")))
   (check-equal? (s-i s1) 42
-                "if the previous test hadn't errored, this would be \"hi\" with type Integer"))
+                "if the previous test hadn't errored, this would be \"hi\" with type Integer")
+  (check-exn #rx"Attempted to use a higher-order value passed as `Any`"
+             (λ () (set-s-i! s4 "hello")))
```
AlexKnauth: It's this test that's failing with `"No exception raised"`.

AlexKnauth: Could the fix for https://github.com/racket/typed-racket/issues/203 have caused this problem?

AlexKnauth:
Should I make it a transparent struct for now so that it can wrap it?

typed-racket-test/succeed/cast-mod.rkt
samth: Ah, yes, that does seem likely. Is there a way to break soundness because of this?

AlexKnauth: Yes:
```racket
#lang typed/racket
(struct (X) s ([i : X]) #:mutable)
(: s1 : (s Integer))
(define s1 (s 42))
(define s2 (ann s1 Any))
(define s3 (cast s2 (s (U Integer String))))
(set-s-i! s3 "unsound!")
(s-i s1) ; the string "unsound!" with the type Integer
```

AlexKnauth: When I make the struct transparent, I get the ``Attempted to use a higher-order value passed as `Any` `` error that I wanted, so that's what I'm doing for the `cast` tests now.

samth: Can you make the unsoundness happen with a typed and an untyped module, though?

AlexKnauth: Yes, see https://github.com/racket/typed-racket/issues/379#issuecomment-229485930

AlexKnauth:
With the transparent struct, the tests for this pass. Should I merge now?

samth:
Yes.

PR #381, opened 2016-07-03
implement sc->constraints for recursive-sc

AlexKnauth:
This pull requests implements the `sc->constraints` generic method for the `recursive-sc` struct.

This changes the error in the `wrap` package mentioned in issue #375 here (https://github.com/racket/typed-racket/issues/375#issuecomment-228580196),

From this:
```
. . ../../../../../../../../../../Applications/Racket/2016-06-16/Racket v6.5.0.5/collects/racket/private/generic.rkt:488:0: 
sc->constraints: not implemented for
(recursive-sc
 (list #<syntax scan1681546> #<syntax scan1681547> #<syntax scan1681548>)
 (list (or/sc2 (list (simple-contract #<syntax:/Applications/Racket/2016-06-16/Racket v6.5.0.5/share/pkgs/typed-racket-lib/typed-racket/private/type-contract.rkt...
```
To this:
```
. Type Checker: Type (U Null (Pairof JsObject (∩ JsList (Listof HashTableTop)))) could not be converted to a contract: 
Intersection type contract contains more than 1 non-flat contract: (∩ JsList (Listof HashTableTop)) in: items-js
```

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/structures.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/structures.rkt
@@ -105,7 +105,7 @@
   ;; Takes a static contract and returns the corresponding contract.
   ;; The function argument should be used for sub parts of the static contract.
   [sc->contract sc f]
-  ;; sc->constraints: static-contract? (static-contract? -> constraint-set?) -> constraint-set?
+  ;; sc->constraints: static-contract? (static-contract? -> contract-restrict?) -> contract-restrict?
   ;; Takes a static contract and computes the constraint set for a static contract.
   ;; The function argument should be used for sub parts of the static contract.
   [sc->constraints sc f]
@@ -139,7 +139,9 @@
                [(recursive-sc names values body)
                 (for-each (λ (v) (f v 'covariant)) values)
                 (f body 'covariant)
-                (void)]))]
+                (void)]))
+           (define (sc->constraints v f)
+             (simple-contract-restrict 'impersonator))]
         #:methods gen:custom-write [(define write-proc recursive-sc-write-proc)])
 
 ;; A use of a contract bound by recursive-sc


samth:
That error message is presumably because we're returning `'impersonator` here, and thus `JsList` appears to be a higher-order contract.

samth:
Also, cc @andmkent for the intersection-related error.

samth:
This is fine to commit now, since it's a strict improvement. But I think we can just recur, since `sc->constraints` is implemented properly on variable references.

But cc @endobson and @takikawa who may know more here.

samth:
I think we should be able to do it using the `f` argument.

AlexKnauth:
I was imagining recurring with `(f body)`, but what about the `names` and `values` fields? How is the contract is actually generated? I noticed the `sc->contract` method isn't implemented either?

AlexKnauth:
Oh, I just looked at the `README` file here:
https://github.com/racket/typed-racket/blob/583a752954b68e29e07a09b542203e2f9e00cbde/typed-racket-lib/typed-racket/static-contracts/README#L39
It says, about the `sc->contract` and `sc->constraints` methods:
> These are not applicable to recursive contract; instantaite uses them in its implementation, and
directly deals with the recursive casses.

But it's used directly here in `type-contract.rkt`, in the `get-max-contract-kind` function:
https://github.com/racket/typed-racket/blob/583a752954b68e29e07a09b542203e2f9e00cbde/typed-racket-lib/typed-racket/private/type-contract.rkt#L225

Should this be changed to use `compute-constraints`, or whatever function is supposed to be used instead of the direct `sc->constraints` method? Or could the `sc->constraints` method be changed to handle recursive cases?

samth:
I think the problem is that contract generation for intersections calls `sc->constraints` (indirectly) on a recursive type, even though that's not supposed to happen. So I think we should just change the method to handle recursive types correctly.

AlexKnauth:
Oh, here:
https://github.com/racket/typed-racket/blob/583a752954b68e29e07a09b542203e2f9e00cbde/typed-racket-lib/typed-racket/private/type-contract.rkt#L407
It uses `get-max-contract-kind` to generate contracts for intersections and that uses `sc->constraints`.

It also appears to be the only thing that uses `get-max-contract-kind`, and `get-max-contract-kind` is the only thing that uses `sc->constraints` directly without handling the recursive case. So what should it use instead of `(get-max-contract-kind c)`?

samth:
I think `get-max-contract-kind` should handle the `recursive-sc` case itself.

Alternatively, we could have a `check-max-contract-kind` that took an expected max kind as input, which might be easier to write, and would be fine here.

AlexKnauth:
The `compute-constraints` function here
https://github.com/racket/typed-racket/blob/583a752954b68e29e07a09b542203e2f9e00cbde/typed-racket-lib/typed-racket/static-contracts/instantiate.rkt#L84
looks like it might be the thing that's supposed to handle the `recursive-sc` case, should it use that? It's provided by the `internals` submodule of `instantiate.rkt`. It requires a max kind as an input, but in this case it's okay if one contract is non-flat, so should the max kind argument be `'impersonator`?

AlexKnauth:
@andmkent It looks like you were the one who added `get-max-contract-kind` for checking intersection type contracts. Do you have any ideas?

andmkent:
I'll sit down and take a gander tomorrow =)

AlexKnauth:
I tried passing in `'impersonator` here:
https://github.com/AlexKnauth/typed-racket/commit/fccfa1bced91fabfa25106c2f952df84910f4e5c
But the `wrap` example still fails and I haven't found anything yet that this fixes. It might doing something better for a few examples, or it might just be returning `'impersonator`, I don't know.

samth:
Just passing in `'impersonator` should always generate `'impersonator` for an actually-recursive contract, because the variable will be assigned `'impersonator`. You could try passing in multiple things and see what works.

samth:
Also, note that if `from-typed?` is `#t` (as it is here), we can just drop some of the conjuncts in the intersection and still be safe (since we'd be placing more restrictions on the untyped code).

AlexKnauth:
Yes, but only if they're flat, right?

AlexKnauth:
Oh! I found an example that that commit helps:
```racket
#lang typed/racket
(define-type T
  (Rec T (U String (Listof T))))
(: f : (Listof T) -> Any)
(define (f x)
  (if (andmap list? x)
      (cast x Any)
      #f))
```
Without that commit, this gives the error:
```
. Type Checker: Type (U Null (Pairof (U Null (Pairof T (Listof T))) (∩ (Listof (Listof Any)) (Listof T))))
could not be converted to a contract: Intersection type contract contains more than 1 non-flat contract: 
(∩ (Listof (Listof Any)) (Listof T)) in: x
```
With this commit, it succeeds.

That commit: https://github.com/AlexKnauth/typed-racket/commit/fccfa1bced91fabfa25106c2f952df84910f4e5c

PR #382, opened 2016-07-05
call compute-constraints instead of sc->constraints in get-max-contract-kind

AlexKnauth:
This fixes programs like:
```
#lang typed/racket
(define-type T
  (Rec T (U String (Listof T))))
(: f : (Listof T) -> Any)
(define (f x)
  (if (andmap list? x)
      (cast x Any)
      #f))
```
Where there is a `cast` from a value with an intersection type that involves recursive types that can be guarded with flat contracts.

This fixes another part of issue #375 discussed in https://github.com/racket/typed-racket/pull/381#issuecomment-230509639, which pull request #381 fell short of solving.

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -18,6 +18,7 @@
  syntax/flatten-begin
  (only-in (types abbrev) -Bottom -Boolean)
  (static-contracts instantiate optimize structures combinators constraints)
+ (only-in (submod typed-racket/static-contracts/instantiate internals) compute-constraints)
  ;; TODO make this from contract-req
  (prefix-in c: racket/contract)
  (contract-req)
@@ -221,9 +222,7 @@
 ;; recurse into a contract finding the max
 ;; kind (e.g. flat < chaperone < impersonator)
 (define (get-max-contract-kind sc)
-    (define (get-restriction sc)
-      (sc->constraints sc get-restriction))
-    (kind-max-max (contract-restrict-value (get-restriction sc))))
+  (kind-max-max (contract-restrict-value (compute-constraints sc 'impersonator))))
 
 ;; To avoid misspellings
 (define impersonator-sym 'impersonator)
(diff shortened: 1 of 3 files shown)


AlexKnauth:
The `wrap` example still fails, but looking at that more closely, neither of the elements in the intersection type could be checked with flat contracts. This pull request only helps programs where the recursive types can be checked with flat contracts.

AlexKnauth:
A should-be-unrelated test is failing with an out-of-memory error:
https://travis-ci.org/racket/typed-racket/builds/142495172#L318
What should I do? Should I increase the memory limit?
In `typed-racket-test/fail/sandboxed-unsafe-ops.rkt`:
```racket
#;
(exn-pred #rx"access disallowed by")
#lang racket/base

;; This test checks that TR's unsafe libraries are not accessible
;; from a sandboxed context

(require racket/sandbox)

(parameterize ([sandbox-memory-limit 5000])
  (define eval (make-evaluator 'typed/racket))
  (eval '(require typed/racket/unsafe))

  ;; should fail
  (eval '(unsafe-require/typed racket/base [values 3])))
```

samth:
Inconveniently, that test failure seems to be something I can't fix -- I've increased the limit before. I'll rerun the tests.

AlexKnauth:
I had re-started the Travis build twice and it hadn't helped.

AlexKnauth:
I just set the memory limit to false, so hopefully that gets rid of these errors. I talked to @takikawa, who added this particular test, and he said there was no real reason for a memory limit on this test.

samth:
Can you open an issue for that test (on this repository) since I think something is going wrong somewhere?

AlexKnauth:
With the memory-limit set to false it's working fine. Should this be merged?

samth:
I'd like @andmkent's opinion, and maybe @takikawa's as well.

andmkent:
LGTM

AlexKnauth:
So should this be merged?

samth:
Yes, this is good to merge.

Issue #380, opened 2016-07-01
Cast doesn't work on types with type variables


AlexKnauth:
In racket version 6.5.0.5, if I run this program:
```racket
#lang typed/racket
(: f : (All (X) (-> X X)))
(define (f x)
  (cast x Any)
  x)
```
I expect it to succeed and create an identity function, but instead I get this compile-time error
```
. . ../../Applications/Racket/2016-06-16/Racket v6.5.0.5/share/pkgs/typed-racket-lib/typed-racket/private/type-contract.rkt:337:4: 
type->static-contract: Recursive value lookup failed. #hash() X
```
This error comes from line 446 of type-contract.rkt, where there's a TODO:
https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/private/type-contract.rkt#L446

Instead of raising this error on types with type variables, should this generate an `any-wrap/c` contract? That would be much better for backwards compatibility.

AlexKnauth:
An easy workaround for these types of examples would be using `(cast (ann x Any) ...)`. However, that's too restrictive to do in general because it disallows higher order uses of casted values. The contract does need to take advantage of the original type, but it should be able to fall back to `any-wrap/c` for things like type variables.

Is that the right thing to do?

PR #384, opened 2016-07-06
Raise a proper contract-generation error for #380

AlexKnauth:
On  `cast`s from types with free-variables, this raises a proper contract-generation type error instead of throwing an internal error. For example this program from issue #380 
```racket
#lang typed/racket
(: f : (All (X) (-> X X)))
(define (f x)
  (cast x Any)
  x)
```
Used to raise the internal error
```
. . ../../Applications/Racket/2016-06-16/Racket v6.5.0.5/share/pkgs/typed-racket-lib/typed-racket/private/type-contract.rkt:337:4: 
type->static-contract: Recursive value lookup failed. #hash() X
```
It now raises the error
```
. Type Checker: Type X could not be converted to a contract because it contains free variables.
in: x
```
With the `x` in `(cast x Any)` highlighted.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -343,13 +343,17 @@
             (define existing-ty-ctc (syntax-local-lift-expression
                                      (make-contract-def-rhs/from-typed existing-ty-id #f #f)))
             (define (store-existing-type existing-type)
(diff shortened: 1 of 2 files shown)


samth:
Can you add a test case? Other than that, LGTM.

AlexKnauth:
I've added a test case in the `fail` directory.

samth:
LGTM

PR #385, opened 2016-07-07
fix any-wrap/c unsoundness on opaque structures

### LINKED ISSUE #379: any-wrap/c is unsound on opaque mutable structures
In racket version 6.5.0.5, if I define an opaque mutable struct in typed code, and pass it as `Any` into either untyped code or a `cast` expression, the `any-wrap/c` contract allows it to be mutated with no way to check that the value coming in matches the type. This can lead to unsoundness, for example
```racket
#lang typed/racket
(struct (X) s ([i : X]) #:mutable)
(: s1 : (s Integer))
(define s1 (s 42))
(define s2 (ann s1 Any))
(define s3 (cast s2 (s (U Integer String))))
(set-s-i! s3 "unsound!")
(s-i s1) ; the string "unsound!" with the type Integer
```

I expected the mutation to raise an error like
```
#<s>: contract violation
  Attempted to use a higher-order value passed as `Any` in untyped code: #<s>
  in: Any
  contract from: typed-world
  blaming: cast
  at: unsaved-editor:6.11
```
Instead, this program succeeds and produces the string `"unsound!"` with the type `Integer`.

AlexKnauth:
Fixes #379 

This modifies `any-wrap/c` to fail on opaque things it doesn't have an inspector for, but it defines a new inspector so that normal uses of `struct` (without a special `#:inspector` argument) create structures that `any-wrap/c` can inspect.

For example on this program:
```racket
#lang typed/racket
(struct (X) s ([i : X]) #:mutable)
(: s1 : (s Integer))
(define s1 (s 42))
(define s2 (ann s1 Any))
(define s3 (cast s2 (s (U Integer String))))
(set-s-i! s3 "unsound!")
;#<s>: contract violation
;  Attempted to use a higher-order value passed as `Any` in untyped code: #<s>
;  in: Any
;  contract from: typed-world
;  blaming: cast
;   (assuming the contract is correct)
;  at: /Users/Alex/Documents/DrRacket/any-wrap-opaque-example.rkt:6.11
(s-i s1) ; if that hadn't errored, this would produce "unsound!" with the type Integer
```
Now avoids unsoundness by producing the error shown in the comments, preventing the `"unsound!"` mutation. Before, this succeeded and allowed `(s-i s1)` to produce `"unsound!"` with the type `Integer`.

### DIFF
--- a/typed-racket-lib/typed-racket/typed-racket.rkt
+++ b/typed-racket-lib/typed-racket/typed-racket.rkt
@@ -5,6 +5,8 @@
              "standard-inits.rkt")
  ;; these need to be available to the generated code
  "typecheck/renamer.rkt" syntax/location
+ ;; this defines the inspector that structs will use
+ "utils/inspector.rkt"
  (for-syntax (submod "base-env/prims-contract.rkt" self-ctor))
  (for-syntax "utils/struct-extraction.rkt")
  (for-syntax "typecheck/renamer.rkt")
@@ -16,6 +18,13 @@
          with-type
          (for-syntax do-standard-inits))
 
+;; This sets the inspector that typed racket structs will use, so that
+;; any-wrap/c can inspect them. This allows any-wrap/c to wrap structs
+;; that are defined in typed racket, and to fail on structs that it
+;; can't wrap safely.
+;; https://github.com/racket/typed-racket/issues/379
+;; https://github.com/racket/typed-racket/pull/385
+(current-inspector new-inspector)
 
 (define-syntax-rule (drivers [name sym] ...)
   (begin
--- a/typed-racket-lib/typed-racket/utils/any-wrap.rkt
+++ b/typed-racket-lib/typed-racket/utils/any-wrap.rkt
@@ -11,7 +11,8 @@
          (only-in racket/udp udp?)
          (only-in (combine-in racket/private/promise)
                   promise?
-                  prop:force promise-forcer))
+                  prop:force promise-forcer)
+         (only-in "inspector.rkt" old-inspector))
 
 (define (base-val? e)
   (or (number? e) (string? e) (char? e) (symbol? e)
@@ -68,11 +69,12 @@
      v 
      "Attempted to use a higher-order value passed as `Any` in untyped code: ~v" v))
   
-  (define (wrap-struct neg-party s)
+  (define (wrap-struct neg-party s inspector)
     (define blame+neg-party (cons b neg-party))
     (define (extract-functions struct-type)
       (define-values (sym init auto ref set! imms par skip?)
-        (struct-type-info struct-type))
+        (parameterize ([current-inspector inspector])
+          (struct-type-info struct-type)))
       (define-values (fun/chap-list _)
         (for/fold ([res null]
                    [imms imms])
@@ -99,7 +101,9 @@
       (cond
         [par (append fun/chap-list (extract-functions par))]
         [else fun/chap-list]))
-    (define-values (type skipped?) (struct-info s))
+    (define-values (type skipped?)
+      (parameterize ([current-inspector inspector])
+        (struct-info s)))
     ;; It's ok to just ignore skipped? -- see https://github.com/racket/typed-racket/issues/203
     (apply chaperone-struct s (extract-functions type)))
  
@@ -163,7 +167,8 @@
        (for/set ([i (in-set v)]) (any-wrap/traverse i neg-party))]
       ;; could do something with generic sets here if they had
       ;; chaperones, or if i could tell if they were immutable.
-      [(? struct?) (wrap-struct neg-party v)]
+      [(? (struct?/inspector old-inspector))
+       (wrap-struct neg-party v old-inspector)]
       [(? procedure?)
        (chaperone-procedure v (lambda args (fail neg-party v)))]
       [(? promise?)
@@ -185,7 +190,10 @@
                                        blame+neg-party
                                        (values v (any-wrap/traverse v neg-party))))
                           (lambda (e) (fail neg-party v)))]
-      [_ (chaperone-struct v)]))
+      [_
+       ;; this would be unsound, see https://github.com/racket/typed-racket/issues/379
+       ;; (chaperone-struct v)
+       (fail neg-party v)]))
   any-wrap/traverse)
 
 (define any-wrap/c
@@ -193,6 +201,10 @@
    #:name 'Any
    #:first-order (lambda (x) #t)
    #:late-neg-projection late-neg-projection))
+
+(define ((struct?/inspector inspector) v)
+  (parameterize ([current-inspector inspector])
+    (struct? v)))
 
 ;; Contract for "safe" struct predicate procedures.
 ;; We can trust that these obey the type (-> Any Boolean).
--- a/typed-racket-lib/typed-racket/utils/inspector.rkt
+++ b/typed-racket-lib/typed-racket/utils/inspector.rkt
@@ -1,0 +1,12 @@
+#lang racket/base
+
+(provide old-inspector new-inspector)
+
+;; Defines a new inspector that typed racket structs will use,
+;; and make this new inspector a sub-inspector of the old one.
+;; The old one is more powerfull than the new one, so any-wrap/c
+;; can use the old one to inspect opaque structs created by
+;; typed racket.
+(define old-inspector (current-inspector))
+(define new-inspector (make-inspector old-inspector))
+
--- a/typed-racket-test/succeed/exn-any-mutation.rkt
+++ b/typed-racket-test/succeed/exn-any-mutation.rkt
@@ -1,0 +1,21 @@
+#lang racket/base
+(module untyped racket/base
+  (provide f)
+  (define (f x) x))
+(module typed typed/racket
+  (require typed/rackunit)
+  (struct (X) s ([i : X]) #:mutable)
+  (require/typed (submod ".." untyped)
+                 [f (-> Any (s (U Integer String)))])
+  (: s1 : (s Integer))
+  (define s1 (s 42))
+  (define s2 (ann s1 Any))
+  (define s3 (f s2))
+  (check-equal? (s-i s1) 42)
+  (check-equal? (s-i s3) 42)
+  (check-exn #rx"Attempted to use a higher-order value passed as `Any`"
+             (λ () (set-s-i! s3 "hi")))
+  (check-equal? (s-i s1) 42
+                "if the previous test hadn't errored, this would be \"hi\" with type Integer")
+  )
+(require 'typed)
--- a/typed-racket-test/succeed/pr241-variation-4.rkt
+++ b/typed-racket-test/succeed/pr241-variation-4.rkt
@@ -5,6 +5,8 @@
 ;; From Issue #203
 ;;   https://github.com/racket/typed-racket/issues/203
 
-(require typed-racket/utils/any-wrap)
+(require typed-racket/utils/any-wrap
+         typed-racket/utils/inspector)
+(current-inspector new-inspector)
 (struct s ())
 (contract any-wrap/c (s) 'a 'b)


typed-racket-lib/typed-racket/utils/any-wrap.rkt
```diff
@@ -105,87 +106,91 @@
...
-                                               (with-contract-continuation-mark
-                                                blame+neg-party
-                                                (any-wrap/traverse v neg-party))))) ;; ref
-                                 (lambda (h k n)
-                                   (if (immutable? v) 
-                                       (values k n)
-                                       (fail neg-party v))) ;; set
-                                 (lambda (h v) v) ;; remove
-                                 (lambda (h k)
+    (parameterize ([current-inspector old-inspector])
```
samth: Can you do this change directly at the call to `struct?`, instead of using `parameterize` around everything?

typed-racket-lib/typed-racket/utils/inspector.rkt
```diff
@@ -0,0 +1,10 @@
+#lang racket/base
+
+(provide old-inspector new-inspector)
+
+;; set up a new inspector that typed racket structs will use,
+;; and make the new inspector a sub-inspector of the old one.
+(define old-inspector (current-inspector))
+(define new-inspector (make-inspector old-inspector))
+(current-inspector new-inspector)
```
samth: I think we should set this somewhere much higher up in the TR implementation, such as `core.rkt`.

typed-racket-lib/typed-racket/utils/any-wrap.rkt
AlexKnauth: I guess, but then there would have to be two `parameterize` forms, one around the call to the `struct?` predicate, and one around the call to `wrap-struct` in that clause. I thought it would be better to have one `parameterize` over the whole thing instead of that.

typed-racket-lib/typed-racket/utils/inspector.rkt
AlexKnauth: Ok, I'll try that.

AlexKnauth: Is `core.rkt` required at compile time? Should it be somewhere else then, like `typed-racket.rkt`?

samth: Yes, relative phase 0 for `core.rkt` is phase 1 for the file using `#lang typed/racket`.

AlexKnauth: So then where should I set the current inspector? Is `typed-racket.rkt` the right place?

typed-racket-lib/typed-racket/utils/any-wrap.rkt
samth: Right, I think that would be clearer.

samth:
This needs documentation, and more tests, but I'd really like to get it in 6.6, before more programs are written that depend on this unsoundness.

cc @stamourv @rmculpepper so that they know about this

stamourv:
I haven't been following this discussion closely.
Does this change break existing programs? If so, is it a regression from 6.5, or only from HEAD?

samth:
Unfortunately, it's possible that this change will break existing programs that work in 6.5, although those programs would not have worked in 6.4 or before. Hopefully the changes that @AlexKnauth has proposed here will avoid that breakage in practice, but it's possible to write programs that will break.

If it turns out that programs we have (such as on the pkg server) break with this change, we'll look into how to fix that.

However, since creating programs with this problem has only been possible recently, I want to release the fix as soon as practical so as to avoid more serious incompatibility down the road.

AlexKnauth:
And this shouldn't break any programs that only rely on opaque structs that were created by typed racket, since typed racket's `struct` form doesn't support the `#:inspector` argument.

typed-racket-lib/typed-racket/utils/any-wrap.rkt
AlexKnauth: So a clause like
```racket
        [(? (λ (v) (parameterize ([current-inspector old-inspector])
                     (struct? v))))
         (parameterize ([current-inspector old-inspector])
           (wrap-struct neg-party v))]
```
Would be better?

samth:
Right, the likely case where this would fail is importing an struct for untyped racket (or a constructor for it), and then passing it back to untyped code as type `Any`. 

This is (I think) a safe thing to do, however, I don't know how to check for it, and I'm not 100% confident that it's safe at this point. If it breaks some existing code, I'll work on something to fix that.

typed-racket-lib/typed-racket/utils/any-wrap.rkt
samth: I think you should write a helper function for `struct?` that takes an inspector, and add an argument to `wrap-struct`.

AlexKnauth:
If a struct is defined in untyped code that can result in unsoundness too, which this pull request also fixes:
```racket
#lang racket/base
(module untyped racket/base
  (provide make-opaque unwrap-opaque)
  (struct opaque (v))
  (define (make-opaque v) (opaque v))
  (define (unwrap-opaque v) (opaque-v v)))
(module typed typed/racket
  (provide wrapped-f)
  (require/typed (submod ".." untyped)
                 [make-opaque (-> Any Any)])
  (define wrapped-f
    (make-opaque (λ ([x : Positive-Fixnum]) (- x 1)))))
(require 'untyped 'typed)
(define f (unwrap-opaque wrapped-f)) ; error
(f "hello") ; if this succeeded this "hello" would be passed as a Positive-Fixnum
```

samth:
No, that wouldn't be a problem, because the function itself is wrapped in `any-wrap/c` because of the type of `make-opaque`.

AlexKnauth:
Oh, right.

typed-racket-lib/typed-racket/utils/any-wrap.rkt
AlexKnauth: Ok, it's now
```racket
      [(? (struct?/inspector old-inspector))
       (wrap-struct neg-party v old-inspector)]
```
With `struct?/inspector` and `wrap-struct` using `parameterize` around their calls to `struct?`, `struct-info`, and `struct-type-info`.

typed-racket-lib/typed-racket/typed-racket.rkt
```diff
@@ -16,6 +18,7 @@
          with-type
          (for-syntax do-standard-inits))
 
+(current-inspector new-inspector)
```
samth: Can you add a comment on this line for what this is doing?

samth:
This looks good to me, with that comment added (and the tests passing).

typed-racket-lib/typed-racket/typed-racket.rkt
AlexKnauth: That comment is up here on line 8:
https://github.com/racket/typed-racket/pull/385/files#diff-bfd28907573d114fba65652f93ae3c05R8

samth: I'd like something that explains what's going on here a bit more.

samth: Including (but not limited to) a link to the relevant threads on github.

AlexKnauth: Is this better:
https://github.com/racket/typed-racket/pull/385/files#diff-bfd28907573d114fba65652f93ae3c05R21
?

samth: That's great.

AlexKnauth:
Should this be put on the release branch as well?

samth:
I would like to, but I want to discuss this with the release managers as well. @stamourv @rmculpepper

samth:
Unfortunately, this breaks the math library: http://drdr.racket-lang.org/35892/racket/share/pkgs/math-lib/math/private/bigfloat/bigfloat-hurwitz-zeta.rkt

We will certainly not be able to put it on the release branch given that, and we need to investigate this failure. I'm looking now.

AlexKnauth:
So I guess it can't know that bigfloats are safe, because they can be represented with ffi pointers?
https://github.com/racket/math/blob/master/math-lib/math/private/bigfloat/mpfr.rkt#L211
And it's defined in untyped code. It looks like it's a predicate generated by `define-cstruct` and I have no idea.

AlexKnauth:
I guess we could keep the inspector stuff but change the last case of `any-wrap/c` to the unsound version. This would still be better than no change at all, because sormal opaque structs defined by typed racket programs will still be wrapped...

samth:
Right, that would be a useful middle ground. 

Another possibility is to warn programmers in the unsound case, so that we don't surprise people as much in 6.7.

stamourv:
As far as the release is concerned, I agree, let's leave this out.

AlexKnauth:
Would it make sense to use a warning instead, as in #388 ?

stamourv:
If that can't cause regressions, I'm fine with that. I'm fine with waiting, too.

samth:
@stamourv I think it's not as simple as just "leave it out", since we want to avoid potential bigger backwards incompatibility down the road.

Here are the options as I see them:

###### Do nothing further (other than fixing the `math` library and the `envy` package, in progress, for 6.7). 
  * Pros: Can't break anything further in 6.6. 
  * Cons: People might write code depending on this unsoundness between now and the 6.7 release, which will then be broken. Also, 6.6 may cause errors related to the fixes to `cast`, which would mean people have to fix twice. Typed Racket is unsound in this way for a release.

###### Add a warning in 6.6
  * Pros: Can't break anything further in 6.6.  Reduces chance that people will write new broken code, encourages fixing old broken code.
  * Cons: Warnings can be hard to see in Racket, and we don't have much history with them. People might write code depending on this unsoundness between now and the 6.7 release, which will then be broken, despite the warning. Also, 6.6 may cause errors related to the fixes to `cast`, which would mean people have to fix twice. Typed Racket is unsound in this way for a release.
  * Choice required: what level to log at.

##### Merge this in 6.6
 * Pros: Fixes unsoundness in Typed Racket soonest. Avoids new code depending on broken behavior.
 * Cons: Breaks code without a warning. 

I'm currently leaning towards the "warning" option, but I'm interested in other perspectives.

@rfindler may also be interested in this.

stamourv:
How long has that unsoundness been in? AIUI, since it broke the math library (which hasn't changed in a while), it must have been around for a long time. So one more release probably won't change much.

samth:
There are two related issues, which is why the math library has worked all the time:

1. What contract do we generate for values of type `Any`, and how does that interact with opaque structures? This is well-represented by this program:

```racket
#lang racket
(module m3 racket
  (define (id x) x)
  (provide id))

(module m4 typed/racket
  (struct (a) b ([c : a]))
  (require/typed (submod ".." m3)
                 [id (-> Any (b (-> String String)))])
  ((b-c (id (b (λ ([x : Integer]) (add1 x))))) "hi"))

(require 'm4)
```

This program behaves unsoundly in 6.5 and 6.4, but not in earlier versions, and is fixed by this pull request.

2. What contract do we generate for the domain of an opaque predicate? This is the relevant issue for the math library. It used to be that the answer was `any/c`, but this is wrong and unsound. We then changed it to be the same as for `Any`, combining this with issue 1. Then we made the mistake in 1, which allowed the math library to work, but only because of that unsoundness. Now we've fixed both problems, leading to the math library issue.

This is represented by this program:

```racket
#lang racket

(module m1 racket
  (define cell #f)
  (define (get) cell)
  (provide q? get)
  (define q? (λ (x) (set! cell x) #t)))

(module m2 typed/racket
  (struct (a) b ([c : a]))
  (require/typed (submod ".." m1)
                 [#:opaque Q q?]
                 [get (-> (b (-> String String)))])
  (q? (b (λ ([i : Integer]) (add1 i))))
  ((b-c (get)) ""))

(require 'm1)
(require 'm2)
```

This program raises a contract error on HEAD, thanks to this pull request, but behaves unsoundly on all prior released versions of Racket that I can test.

So, for direct use of `Any`, the unsoundness first appeared in 6.4. For `#:opaque` specifications, it's been there for much longer (and I worry more there about the backwards compatibility risk for existing code).

samth:
@stamourv suggests (and I agree) that we do the following:

1. Change the behavior for the use of the `Any` type to be sound in 6.6
2. Change the behavior for `#:opaque` to generate a warning (as in #388) in 6.6.
3. Change `#:opaque` to be sound in a future release.

To implement this, I suggest that we parameterize `any-wrap/c` with an argument saying what to do on opaque values, and then call it appropriately in the relevant places (`prims-contract.rkt` for `#:opaque`). 

@AlexKnauth, do you want to modify #388 to do that? Otherwise, I'll do it tomorrow.

AlexKnauth:
I updated https://github.com/racket/typed-racket/pull/388 to add `any-wrap-warning/c`. So what are all the places where I should use it?

samth:
Just the one file I mentioned, I think.

AlexKnauth:
@bennn and I were talking, and we thought, it's safe for `any-wrap/c` to not wrap it if it's an untyped opaque struct, because any typed value in that struct would have had to go through the typed-untyped boundary before then, so that typed value would already be wrapped. So it would still be unsafe it were a truly opaque typed struct, but that could only happen if it defined a new inspector. Even that should be safe if the new inspector is a sub-inspector of `old-inspector` from `typed-racket/utils/inspector.rkt`, and that's something we can check within the code generated by the typed `struct` macro. If we do the inspector check then, we don't have to wrap opaque structs, I think. Will that be sound?

samth:
I've had that thought myself, but I now want to see an argument at greater length than a github comment for this before we go off in a potentially-unsafe direction. We've gotten this wrong too many times already.

bennn:
Agreed about having a stronger argument, but I think the thoughts in #203 still haven't been proven wrong.

AlexKnauth:
The thoughts in #203 cover the case where fully untyped code uses the accessor, but not the case where typed code uses an accessor (or mutator). When typed code uses an accessor that was defined on the typed side, there's no contract boundary to protect the result of the accessor. That was the flaw in the reasoning.

So the cases are the product of these: (I think?)
 - Was the struct type defined in typed or untyped code? 
 - Was the constructor or mutator called in typed or untyped code?
 - Did the struct value flow from typed-to-untyped or untyped-to-typed? Or typed-untyped-typed, or untyped-typed-untyped? What type did it cross as, in each direction?
 - Was the accessor called in typed or untyped code?

There are a few constraints like if the constructor or mutator was called on one side, then it has to start on that side, and end up on the side where the accessor is called. Also if an accessor is used in typed code, the struct value needs to have a type that the accessor accepts.

So arguing those cases:
- If the struct was defined in typed code:
    - If the constructor or mutator was called in typed code: we know that the value in the struct is a typed value (of type `A`), and it's not guarded by a contract
        - If the struct value flowed as only from typed-to-untyped, as `Any`:
            - The accessor was called in untyped code, and it's safe to pass the struct unwrapped because the accessor has its own contract that will protect the result.
        - If the struct value flowed from typed-to-untyped as `Any` and back to typed as `(s B)`: we know that the value inside will be protected as a `B`
            - The accessor was called in typed code, and it's unsafe to pass the struct unwrapped, because there's no contract on the accessor protecting the type, and there's nothing protecting the original type of the value, `A`. This was a case that #203 missed.
    - If the constructor or mutator was called in untyped code: we know that the value in the struct is an untyped value, and it is guarded by a contract because of contract on the constructor/mutator. The contract it's guarded with depends on the type the constructor/mutator was provided with. So let's say it was a constructor with the type `(-> A Any)`, so the untyped value is protected as an `A` as it goes through the typed code of the constructor. The constructor then returns a struct, which flows to untyped code as `Any`.
        - If the struct value then flowed from untyped-to-typed as `(s B)`: we know that the untyped value inside will be protected as both an `A` and a `B`
            - The accessor was called in typed code, and its result will be an untyped value protected with both `A` and `B` contracts. This is safe.
        - If the struct value flowed from untyped-to-typed as `(s B)` and back to untyped as `Any`: we know that the untyped value will be protected as both an `A` and a `B`.
            - The accessor was called in untyped code, and let's say it was provided with the type `(-> Any C)`. The untyped value will now be protected as an `A`, a `B`, and a `C`. This is safe.
        - If the struct value flowed from untyped-to-typed as whatever, and back to untyped as `Any`: the untyped value will still be protected as an `A`.
            - The accessor was called in untyped code, and the untyped value will be protected as an `A` and a `C`.
- If the struct was defined in untyped code:
    - If the constructor/mutator was called in typed code: we know that the value in the struct is a typed value (of type `A`), guarded as a `B` if the constructor/mutator is required as `(-> B Any)` or something. But for this to have typechecked, `A` must be a subtype of `B`.
        - If the struct value flowed from typed-to-untyped as `Any`:
            - The accessor was called in untyped code, and the value it returns is guarded as a `B`. This is safe.
        - If the struct value flowed from typed-to-untyped as `Any`, and back to typed as `(s C)`:
            - The accessor was called in typed code, and the value it returns is has the type `C`. It is guarded both as a `B` and as a `C`, so this is safe (since the original type was a subtype of `B`).
    - If the constructor/mutator was called in untyped code: we know that the value inside is an untyped value, and it's guarded by a contract
        - If the struct value flowed from untyped-to-typed as `(s A)`: the value inside will be guarded as an `A`
            - The accessor was called in typed code. The accessor was required as `(-> Any B)`, and the return value will be guarded as both an `A` and a `B`. This is safe.
        - If the struct value flowed from untyped-to-typed as whatever and back to untyped as `Any`:
            - The accessor was called in untyped code, and it will get an untyped value, so this is safe.

The only unsafe case of these is the one where there is a struct defined in typed code, the constructor/mutator was called in typed code, and the struct value flowed from typed-to-untyped and back to typed, so that the accessor was also called in typed code. Since the struct, the constructor, mutator, and accessor are all in typed code. The only contracts are the `Any` contract going from typed-to-untyped, and the `(s B)` contract going from untyped-to-typed. This means the original `A` value isn't protected an as `A`, unless `Any` can protect it by raising an error.

rfindler:
Sounds like you need a model and a theorem stated in terms of that model.

PR #386, opened 2016-07-07
don't throw internal errors for unreachable cast exprs

### LINKED ISSUE #378: Internal error when cast is used in dead code
In racket version 6.5.0.5, if I compile the program

``` racket
#lang typed/racket
(cond [#false (cast 1 Integer)]
      [else 2])
```

I get the error

```
. . ../../Applications/Racket/2016-06-16/Racket v6.5.0.5/share/pkgs/typed-racket-lib/typed-racket/utils/tc-utils.rkt:270:0: 
Internal Typechecker Error: contract-def-property: thunk called too early
  This should only be called after the type-checking pass has finished.
while typechecking:
here
originally:
here
```

I expected this to succeed and produce `2` when run.

This happens because the type-checker doesn't look for the `cast` expression in dead code, and the contract-generation pass was expecting the type-checker to supply a type for it.

Should we generate an `any-wrap/c` contract in these cases? Or should the type-checker traverse dead code specifically looking for the casted-expr property?

From John Clements' report on the mailing list:
https://groups.google.com/forum/#!topic/racket-users/LGnWkOvvWlY

AlexKnauth:
Fixes #378 

This fills the corresponding entries in the cast table with a dead code type so that when the contract-generation pass calls the `contract-def` thunk, it finds that in the table.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-if.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-if.rkt
@@ -1,11 +1,13 @@
 #lang racket/unit
 (require "../utils/utils.rkt"
          (rep prop-rep)
-         (types utils prop-ops)
+         (types abbrev utils prop-ops)
          (utils tc-utils)
          (typecheck signatures tc-envops tc-metafunctions)
          (types type-table)
-         racket/match)
+         (private syntax-properties)
+         racket/match
+         syntax/parse)
 
 ;; if typechecking
 (import tc-expr^)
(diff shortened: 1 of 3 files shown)


samth:
This needs some tests.

AlexKnauth:
Ok, I've added tests in `succeed/cast-mod.rkt`.

samth:
Could the contract generation pass just fill this in itself, instead of doing extra traversals?

AlexKnauth:
But then how could it tell between dead code and an actual TR bug?

samth:
This is a balance between cases where we'd catch a bug earlier in TR because it didn't check something it should, and doing extra work that's not needed. Since it's probably not much extra work, I'm ok with going this way.

Issue #32, opened 2015-02-24
Typed Racket pretends mutable sets don't exist, and yet it assumes all hashtables could be mutable


lexi-lambda:
Typed Racket only provides types for immutable set operations. It would be painless to add types for mutable operations as well, but TR also considers `Setof` types to be covariant, which obviously isn't true with mutable sets.

This is inconsistent with how TR handles hashtables. In an ideal world, I'd like `HashTable` to be immutable and have a separate `MutableHashTable` for mutable hash tables and mutable operations. This would be nice for a number of reasons:

- This would let hashtables be covariant.
- This would make contracts on immutable hashtables flat contracts, so predicate generation and occurrence typing would work with them.
- Mutability would need to be explicitly noted in the type system.

I'd unfortunately guess that this is not possible to change at this point. Still, the behavior with sets and hashtables are inconsistent. If we can't change `HashTable`, can we make `MutableHashTable` and `ImmutableHashTable` subtypes? Then could we make the equivalent variations for sets?

shekari:
That is because TR's set code was written before mutable sets existed.

Splitting out `HashTable` might be useful, but I'm not sure if you would want to split out `ImmutableHashTable` or `ReadableHashTable`. The first gets you the flat contracts/predicates, the second gets you covariance. Nothing gets you required explicit mutability because you can just revert to `HashTable` and not specify it.

Sets get more complicated because there are also generic sets which TR still doesn't support.

lexi-lambda:
@shekari Could you explain the typing difference between immutable hash tables and readable hash tables? Why wouldn't they be equivalent?

samth:
Immutable hash tables would support `hash-set`, and you can't pass a
mutable hash table where one is expected.

Read-only tables would admit mutable ones, but you could only read them.

Sam

lexi-lambda:
@samth Oh, I see, that makes sense. `ImmutableHashTable` would still get you covariance, though, yes?

Wouldn't having explicit mutability still be helpful, though? Contracts support it, and then `hash-set!` and friends would be checked statically rather than at runtime. Though I suppose it wouldn't be backwards-compatible.

samth:
Yes, all of these things would be good ideas, and I think we could make everything compatible.

lexi-lambda:
[...]

PR #392, opened 2016-07-09
WIP: types for immutable data

bennn:
Work in progress: add types for immutable hashes, vectors, boxes, strings, and bytes.

At the moment this is just types for hashes. The plan is to add types for `Immutable-HashTable` and `Mutable-HashTable`. These are both subtypes for `HashTable` (and `HashTableTop`), which doesn't know if its value is mutable or immutable.

Current status:
- [x] fix (subtyping?) issue: cannot do `(ann (make-hash) (Mutable-HashTable Integer Integer))` without an `inst`. 
- ~~should be able to refine types of `hash-update` etc to type error if called with `Mutable-Hash`~~
- [ ] missing vector, boxes, strings, bytes
- [ ] also add `Mutable-Set` as sibling to `Set` (https://github.com/racket/typed-racket/issues/32) 
- [ ] check that immutable data can be passed as `Any` to untyped code
- [x] cannot do `(ann (make-hash null) (HashTable Integer Integer))`


bennn:
(No point running travis yet. Unit tests fail because I haven't got subtyping right.)

samth:
Great, thanks for working on this.

My number one worry is about compatibility here. For example, I think your item 2 is not a good idea, since people might have programs that start breaking. We will have to think though various scenarios to make sure we get this right.

typed-racket-lib/typed-racket/infer/infer-unit.rkt
```diff
@@ -720,8 +723,15 @@
           [((CustodianBox: t) (Evt: t*)) (cg S t*)]
           [((Channel: t) (Evt: t*)) (cg t t*)]
           [((Async-Channel: t) (Evt: t*)) (cg t t*)]
-          ;; we assume all HTs are mutable at the moment
-          [((Hashtable: s1 s2) (Hashtable: t1 t2))
+          [((Immutable-Hashtable: s1 s2)
+            (or (Hashtable: t1 t2)
+                (Immutable-Hashtable: t1 t2)))
+           ;; for immutable hash tables, covariant
+           (% cset-meet (cg s1 t1) (cg s2 t2))]
```
samth: The case where (IH A B) < (H A' B') doesn't seem right to me. I think that should be invariant.

typed-racket-test/fail/bad-hash-ref.rkt
```diff
@@ -1,8 +1,8 @@
 #lang typed/scheme
 
-(: table (HashTable Integer (-> Integer)))
+(: table (Mutable-HashTable Integer (-> Integer)))
 (define table
-  (make-immutable-hash null))
+  (make-hash null))
```
samth: Add new tests (or new code to this file) rather than changing existing ones.

typed-racket-test/succeed/for-hash.rkt
```diff
@@ -10,73 +11,79 @@
 
 (check-pred
  hash?
- (for/hash: : (HashTable Integer Symbol)
+ (for/hash: : (Immutable-HashTable Integer Symbol)
```
samth: Similarly in this file, don't remove existing tests.

andmkent:
Maybe this is a dumb question, is there a reason why we couldn't do (HashTable A B) ::= (U (MutableHashTable A B) (ImmutableHashTable A B))?

bennn:
I can't think of one. Will try soon.

typed-racket-lib/typed-racket/infer/infer-unit.rkt
bennn: I think it's fine because the value is invariant. You're safe to treat an (IH A B) as an (H A' B') because any write to the (H ...) is a runtime error.

If we remove this case then `for/hash` needs to return a `HashTable` instead of an `Immutable-HashTable`, otherwise unit tests fail. E.g.
```
#lang typed/racket/base

(for/hash : (HashTable Integer Integer)
          ([x (in-list '(1 2 3))])
  (values x x))
```

bennn:
Two notable updates:
- `in-mutable-hash` [etc.](https://github.com/racket/typed-racket/pull/392/files#diff-f2dc58f5e3442d2ff98fabd8e505158eR162) now work in this PR, but I had to add a `Weak-Hash` type.
- The type signatures for `get-map-function-table`& [ht](https://github.com/racket/typed-racket/pull/392/files#diff-21ea6506c1f86d47c270cb30759b7ce3R1004) were wrong, at least for the [`(keymap:get-global)`](https://github.com/racket/typed-racket/pull/392/files#diff-b364a5aa6468e7e864196ebc798022ccR9) map. Those are now changed.

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -905,51 +905,65 @@
...
+[make-hasheqv (-poly (a b) (->opt [(-lst (-pair a b))] (-Mutable-HT a b)))]
+[make-weak-hash (-poly (a b) (->opt [(-lst (-pair a b))] (-Weak-HT a b)))]
+[make-weak-hasheq (-poly (a b) (->opt [(-lst (-pair a b))] (-Weak-HT a b)))]
+[make-weak-hasheqv (-poly (a b) (->opt [(-lst (-pair a b))] (-Weak-HT a b)))]
+[make-immutable-hash (-poly (a b) (->opt [(-lst (-pair a b))] (-Immutable-HT a b)))]
+[make-immutable-hasheq (-poly (a b) (->opt [(-lst (-pair a b))] (-Immutable-HT a b)))]
+[make-immutable-hasheqv (-poly (a b) (->opt [(-lst (-pair a b))] (-Immutable-HT a b)))]
+
+[hash-set (-poly (a b) (cl-> [((-HT a b) a b) (-Immutable-HT a b)]
+                             [((-Immutable-HT a b) a b) (-Immutable-HT a b)]))]
```
samth: This case should come first.

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -905,51 +905,65 @@
...
+[hash-set! (-poly (a b) (cl-> [((-HT a b) a b) -Void]
+                              [((-Mutable-HT a b) a b) -Void]
```
samth: Are these additional cases needed?

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -905,51 +905,65 @@
...
+[hash-ref! (-poly (a b) (cl-> [((-HT a b) a (-> b)) b]
+                              [((-Weak-HT a b) a (-> b)) b]
+                              [((-Mutable-HT a b) a (-> b)) b]))]
 [hash-has-key? (-HashTop Univ . -> . B)]
 [hash-update! (-poly (a b)
                      (cl-> [((-HT a b) a (-> b b)) -Void]
-                           [((-HT a b) a (-> b b) (-> b)) -Void]))]
+                           [((-HT a b) a (-> b b) (-> b)) -Void]
+                           [((-Mutable-HT a b) a (-> b b)) -Void]
+                           [((-Mutable-HT a b) a (-> b b) (-> b)) -Void]
```
samth: Same question for the others.

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -905,51 +905,65 @@
...
+                           [((-Weak-HT a b) a (-> b b)) -Void]
+                           [((-Weak-HT a b) a (-> b b) (-> b)) -Void]))]
 [hash-update (-poly (a b)
-                    (cl-> [((-HT a b) a (-> b b)) (-HT a b)]
-                          [((-HT a b) a (-> b b) (-> b)) (-HT a b)]))]
-[hash-remove (-poly (a b) (cl-> [((-HT a b) Univ) (-HT a b)]
+                    (cl-> [((-HT a b) a (-> b b)) (-Immutable-HT a b)]
+                          [((-HT a b) a (-> b b) (-> b)) (-Immutable-HT a b)]
+                          [((-Immutable-HT a b) a (-> b b)) (-Immutable-HT a b)]
```
samth: Again, specialized first.

typed-racket-lib/typed-racket/base-env/base-special-env.rkt
```diff
@@ -133,53 +133,33 @@
           (cl-> [((-HT a b)) (-seq (-pair a b))]
                 [(-HashTop) (-seq (-pair Univ Univ))]))]
   [(make-template-identifier 'default-in-immutable-hash 'racket/private/for)
-   (-poly (a b)
-          (cl-> [((-HT a b)) (-seq a b)]
-                [(-HashTop) (-seq Univ Univ)]))]
+   (-poly (a b) (-> (-Immutable-HT a b) (-seq a b)))]
```
samth: This seems backwards-incompatible.

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -512,6 +512,12 @@
 ;; value : Type
 (def-type Hashtable ([key Type/c] [value Type/c]) [#:key 'hash]
   [#:frees (λ (f) (combine-frees (list (make-invariant (f key)) (make-invariant (f value)))))])
+(def-type Immutable-Hashtable ([key Type/c] [value Type/c]) [#:key 'hash]
+  [#:frees (λ (f) (combine-frees (list (make-invariant (f key)) (make-invariant (f value)))))])
```
samth: These should not be invariant.

typed-racket-lib/typed-racket/utils/hash-contract.rkt
```diff
@@ -0,0 +1,8 @@
+#lang racket/base
+
+(require racket/contract)
+(provide mutable-hash/c immutable-hash/c weak-hash/c)
+
+(define (mutable-hash/c k v) (and/c (hash/c k v #:immutable #f) (not/c hash-weak?)))
+(define (immutable-hash/c k v) (hash/c k v #:immutable #t))
+(define (weak-hash/c k v) (and/c (hash/c k v #:immutable #f) hash-weak?))
```
samth: I think order matters in `and/c`.

typed-racket-lib/typed-racket/base-env/base-env.rkt
bennn: No they're not, I'll remove them. But is there any hope we could deprecate the `-HT` signatures?

samth: We certainly won't be able to actually remove them for a long time, if ever. I expect enforcing this would break real code.

typed-racket-lib/typed-racket/base-env/base-special-env.rkt
bennn: I don't think there is a backwards here. On Racket 6.5, this program gives me errors about `unsafe-hash-iterate-...` identifiers.

```
#lang typed/racket/base
(for ([(x y) (in-immutable-hash #hash((a . 1)))])
  (void))
```

samth: What about on HEAD (or in the upcoming 6.6)?

bennn: Same on head & the release branch

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
```diff
@@ -398,14 +398,28 @@
...
+            make-Mutable-Hashtable]))
+        (make-H
          (check-below (apply Un kts) kt)
          (check-below (apply Un vts) vt))]
-       [_ (make-Hashtable (generalize (apply Un (map find-stx-type (hash-keys h))))
-                          (generalize (apply Un (map find-stx-type (hash-values h)))))])]
+       [_
+        (let ([kt (generalize (apply Un (map find-stx-type (hash-keys h))))]
+              [vt (generalize (apply Un (map find-stx-type (hash-values h))))])
+          (Un (make-Mutable-Hashtable kt vt)
```
bennn: Q. is there a place I can define a helper function `(make-Hashtable a b)` to use here & in `tc-literal.rkt`?

bennn:
Currently in the middle of making `(HashTable A B)` be a union of Immutable, Mutable, and Weak hash.

Something's wrong tho, I can compile `(define-type x (HashTable Integer Integer))` but I can't do:
```
#lang typed/racket
(for/hash: : (HashTable Integer Symbol)
           ((a (list 1 2 3)))
  (values a 'a))
```

The error is:
```
application: not a procedure;
 expected a procedure that can be applied to arguments
  given: HashTable
  arguments...:
   a
   b
  compilation context...:
   /Users/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-test/b.rkt
  context...:
   /Users/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-lib/typed-racket/base-env/env-lang.rkt:21:26
   /Users/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-lib/typed-racket/env/global-env.rkt:65:0: lookup-type
   ....
```

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
samth: Yes, there are lots of potential places, but you could just put it in `tc-literal`.

bennn:
status update: not ready to merge, still fixing that bug from July

fosskers:
These would be great to have!

bennn:
See #559, that one is rebased and passes its tests.

PR #388, opened 2016-07-08
add a version of any-wrap/c that displays a warning on opaque structs

AlexKnauth:
see https://github.com/racket/typed-racket/pull/385#issuecomment-231354377

This allows passing opaque structs as `Any` into `#:opaque` predicates without wrapping them properly, even though it's unsound. It displays a warning message instead of raising an error.

PR #396, opened 2016-07-11
any-wrap/c displays warning on opaque structures

AlexKnauth:
This includes #385 and #388, but makes it a warning instead of an error to pass an opaque struct as `Any` in all cases, not just in `#:opaque` predicates.

This would also rely on https://github.com/racket/racket/commit/8d698389e76d8a53baca59b0be9b5680dc3f2d75 being merged into the main racket release branch.


samth:
Can you change this to not rely on those other changes?

AlexKnauth:
I guess I can leave out your "Allow cpointer predicates" commit, but then there will be warnings where there shouldn't be. The `math` library right now doesn't cause any of these warnings, but without that commit, it would cause these warnings every time `bigfloat?` is used on a bigfloat.

samth:
That's a good point. For this issue, we therefore have a few options (recapping my previous format):

##### Don't warn in the release

Will break people's code in the future with less warning. Requires no changes to the release branch.

##### Warn in the release, but without https://github.com/racket/racket/commit/8d698389e76d8a53baca59b0be9b5680dc3f2d75

Will provide people with warnings, but the `math` library will cause lots of warnings (whenever bigfloats are used).

##### Warn in the release, with https://github.com/racket/racket/commit/8d698389e76d8a53baca59b0be9b5680dc3f2d75

Will provide users of 6.6 with warnings, but not have excessive `math` library warnings. Requires more code change in the release branch.

I'm interested in the release managers' opinion here: @stamourv @rmculpepper @rfindler @mflatt @mfelleisen

stamourv:
3rd option sounds preferable, since we already have 8d69838. Better than a bunch of spurious warning that would go away in the next release.
I don't feel confident reviewing 8d69838, though. @mflatt?

AlexKnauth:
Was this merged as these three commits?
https://github.com/racket/typed-racket/commit/878e856affc5bc938bd7ce4cac894a6c26f689eb
https://github.com/racket/typed-racket/commit/ee239abeb5359132736ace998cba08c27775bea9
https://github.com/racket/typed-racket/commit/6acef79abf201393a8e3a1c443461019969f088d

Also, does `log-error` actually do anything that users will ever see?

samth:
Yes. And `log-error` prints to STDERR in the default configuration of Racket.

AlexKnauth:
If I run the program
```racket
#lang racket
(log-error "helloo")
```
In DrRacket it succeeds and doesn't print anything.

samth:
Right, DrRacket captures all log output and you can view it in the log pane.

PR #401, opened 2016-07-12
Don't try to name casted values in the contract.

samth:
Relevant to Racket bug [15323](http://bugs.racket-lang.org/query/?cmd=view&pr=15323).


stamourv:
Uh, LGTM?
I imagine a test would be overkill. I trust you've tried it?

samth:
This was merged.

PR #406, opened 2016-07-13
have overlap use a current-seen list

andmkent:
fixes a bug (#403) ```overlap?``` triggered (an infinite loop which continually unfolded recursive types) by keeping around previously seen types and checking against them.

### DIFF
--- a/typed-racket-lib/typed-racket/types/overlap.rkt
+++ b/typed-racket-lib/typed-racket/types/overlap.rkt
@@ -21,6 +21,18 @@
       (void? v)
       (eof-object? v)))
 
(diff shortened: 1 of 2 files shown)


samth:
Looks good. I'd like to have release manager approval to merge this to the release as well. cc @stamourv @rmculpepper

stamourv:
@samth: You do not need release manager approval at this point in the process. You will once testing begins, though.
In case you want my opinion regardless, this fix would prevent a regression, which is good. I'm not confident I understand the ramifications of the changes, though, but if you think it's fine, I'm ok with it.

PR #410, opened 2016-07-19
Fix contracts for structs with the same name (PR 15330)

takikawa:
This is a fix intended to go into the TR release for PR 15330.

There are several ways to fix this. The one I chose was changing the equality to traverse syntax objects by structure, comparing leaves by free-id=? for identifiers and equal? otherwise.

Another option would be to add an (optional) value to these static contracts that is used for equality checking. Then we could stuff the struct predicate identifier in there for struct contracts.

Yet another is to compile struct contracts specially, avoiding using `flat/sc`.

Any opinions on which approach is best?

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/combinators/simple.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/combinators/simple.rkt
@@ -30,16 +30,27 @@
   (display close port))
 
 
+;; check equality of two syntax objects by structural traversal
+;; where identifiers are compared by free-identifier=?
+(define (stx-equal? s1 s2 recur)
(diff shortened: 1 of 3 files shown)


samth:
Given that the previous code is obviously wrong even for non-struct contracts (we can't compare syntax with `syntax->datum`), I think the 3rd option is a non-starter . What happens if we just call `equal?` on the syntax objects (at least for the release)? It will just cause bigger contracts, right?

takikawa:
Should be fine for non-struct and maybe non-opaque contracts, since we usually don't put in arbitrary identifiers into these syntax objects. Calling `equal?` on them is equivalent to reverting the original commit.

PR #411, opened 2016-07-29
Make M/Listof expanders not add new pattern vars

btlachance:
To reproduce, see the snippet below where match complains about `var` not being bound in all branches of the pattern.

``` racket
#lang racket
(require typed-racket/types/match-expanders)

(match #f
  [(or (Listof: _) _) #t])
```


samth:
A test would be good. Otherwise seems good to me.

btlachance:
Updated with tests. Happy to relocate them if you meant something else by "success tests."

samth:
These match expanders have changed to no longer have the bug.

PR #414, opened 2016-08-07
Support for changes mandatory-keyword expansion

mflatt:
These two commits are intended to support a revised expansion of function definitions that have mandatory keyword arguments. The first commit is a more general repair to support `syntax-local-lift-values-expression` in full. The second commit is specific to the keyword expansion.

A branch of Racket with the change to keyword expansion (plus related optimizer changes): mflatt/racket@ad230d2ca09ee45002eaa1bea24a1e8ce933e3a9 (the `mandatory-keywords` branch).


samth:
This looks good to me. Thanks!

PR #418, opened 2016-08-19
unsafe-require/typed #:opaque doesn't warn about opaque structs passed as Any

AlexKnauth:
This changes how the `#:opaque` option in `unsafe-require/typed` behaves. It changes the contract on the predicate to allow opaque structs to be passed as `Any` without displaying a warning.

This fixes an issue raise in https://groups.google.com/forum/#!topic/racket-users/-pgP7-C_3D4, so that it can work if you use `unsafe-require/typed`

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -163,7 +163,7 @@
   (define-syntax-class (clause legacy unsafe? lib)
    #:attributes (spec)
    (pattern oc:opaque-clause #:attr spec
-     #`(require/opaque-type oc.ty oc.pred #,lib . oc.opt))
+     #`(require/opaque-type oc.ty oc.pred #,lib #,@(if unsafe? #'(unsafe-kw) #'()) . oc.opt))
    (pattern (~var strc (struct-clause legacy)) #:attr spec
      #`(require-typed-struct strc.nm (strc.tvar ...)
                              (strc.body ...) strc.constructor-parts ...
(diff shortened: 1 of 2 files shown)


typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -391,9 +393,13 @@
            #,(if (attribute ne)
                  (internal (syntax/loc stx (define-type-alias-internal ty (Opaque pred))))
                  (syntax/loc stx (define-type-alias ty (Opaque pred))))
-           #,(ignore #'(define pred-cnt
-                         (or/c struct-predicate-procedure?/c
-                               (any-wrap-warning/c . c-> . boolean?))))
+           #,(if (attribute unsafe)
+                 (ignore #'(define pred-cnt
```
samth: Just use `(define pred-cnt any/c)`.

samth:
Let's add a test as well.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
AlexKnauth: Shouldn't it still attempt to check that it is a predicate? I was trying to make the change in behavior as minimal as possible for this

samth: If we wanted that, we can just do `(any/c . ->.  boolean?)`. But I don't think we should add checks for `unsafe-`.

AlexKnauth: I'll change it to use `predicate/c` then

samth: No, really, let's not use a contract. For example, when testing the performance impact of contracts, it's nice to be able to use `unsafe-require/typed`. Sometimes adding contracts defeats that.

AlexKnauth:
Why not leave the `predicate/c` check there? It doesn't wrap the argument like `(-> any-wrap/c boolean?)` would, so it only allocates one wrapper ever.

samth:
Adding wrappers for unsafe forms is not right. The point of unsafe is to not have checks at all, not just to avoid tricky checks.

AlexKnauth:
Ok, I've changed it to `any/c`. (I don't think I can change it to not use `require/contract`, because that does other stuff)

AlexKnauth:
Should this be merged?

Issue #289, opened 2016-01-14
unsafe-provide does not always eliminate boundaries


stamourv:
The following program spends ~45% of its time checking contracts, despite all the bindings it gets from typed code coming through `unsafe-provide`.

```
#lang racket

(module m typed/racket
  (require math/array typed/racket/unsafe)
  (unsafe-provide for/array array-ref))

(require 'm)

(define a (for/array ([i (in-range 10000)]) (+ i 1)))

(for ([i (in-range 10000)])
  (array-ref a (vector i)))
```

samth:
There look to be two options here. One, the current behavior, where unsafe-provide is really like plain Racket's provide. Two, we change unsafe-provide to generate new runtime definitons, so that the reference to the array-ref identifier exported by math is not ever expanded in untyped code. The second doesn't seem like a great option, so maybe this should remain a documented limitation of unsafe-provide. But if you have an implementation idea that doesn't produce a new value level definition, then that seems like a better possibility.

stamourv:
I suspect that "using the math library from untyped Racket" is one of the most common use cases for `typed/racket/unsafe`, so solution 1 is not satisfactory, IMO.

Can you elaborate on why you think solution 2 is not a great option?

samth:
I think that if the math library wants to provide an unsafe interface, then that should be part of `math`. Is this issue a problem for implementing that, or just for doing it without changing `math`? We shouldn't be afraid of changing the library to add this interface.

Solution 2 isn't great because it's (a) costly in zo size -- lots of new definitions, (b) potentially bad for performance, since the optimizer now needs to see through more definitions, and (c) much more complex to implement (right now, `unsafe-provide` doesn't need to change anything about type checking, for example, nor does it need to worry about static struct info).

stamourv:
So far, my motivation has been to add unsafety to `math` post-fact, in my own code.
But AFAICT, whether the module with the `unsafe-provide` is in the `math` library or in 3rd party code shouldn't matter, which would make this strategy not work to have `math` provide an unsafe interface.

samth:
This program shows some easy ways to do what you want, I think:

```
#lang racket

(module m1 typed/racket
  (require math/array typed/racket/unsafe)
  (unsafe-provide for/array array-ref))

(module m2 racket
  (require (submod ".." m1))
  
  (define a (for/array ([i (in-range 10000)]) (+ i 1)))
  
  (time (for ([i (in-range 10000)])
          (array-ref a (vector i)))))

[...]

PR #422, opened 2016-08-31
add unsafe-reprovide

bennn:
- adds note to documentation for `unsafe-provide`,
  that only identifiers defined in the current module can get contracts removed
- explains Sam's work-around
- adds `unsafe-reprovide` shorthand for re-defining & re-(unsafe-)providing

plus 2 tests for `unsafe-reprovide`

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/unsafe.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/unsafe.scrbl
@@ -39,6 +39,10 @@ behavior and may even crash Typed Racket.
   This form declares exports from a module with the same syntax as
   the @racket[provide] form.
 
+  @margin-note{@racket[unsafe-provide] only prevents contract generation for
+   identifiers defined in the same module as the unsafe provide.
+   See docs for @racket[unsafe-reprovide].}
+
   Unlike @racket[provide], this form is unsafe and Typed Racket will not generate
   any contracts that correspond to the specified types. This means that uses of the
   exports in other modules may circumvent the type system's invariants.
@@ -77,6 +81,42 @@ behavior and may even crash Typed Racket.
   Like @racket[require/typed/provide] except that this form is unsafe and will not generate
   contracts that correspond to the specified types to check that the values
   actually match their types.
+}
+
+@defform[(unsafe-reprovide id ...)]{
+  The current implementation of @racket[unsafe-provide] can only prevent contract
+   generation for identifiers defined in the same module as the @racket[unsafe-provide] statement.
+  If a typed module imports another typed module's provides and tries to re-provide them unsafely,
+   the re-provided identifiers will still be protected with contracts.
+
+  @examples[#:eval (make-base-eval)
+    (module t typed/racket/base
+      (require math/array typed/racket/unsafe)
+      (unsafe-provide array-ref))
+
+    (require 't racket/contract)
+    (has-contract? array-ref)
+  ]
+
+  One way to work around this limitation is to re-define the typed identifiers
+   before re-providing them.
+  The @racket[unsafe-reprovide] form is a shorthand for re-defining and unsafely
+   providing a sequence of identifiers.
+  Replacing the above @racket[unsafe-provide] with an @racket[unsafe-reprovide]
+   is semantically equivalent to the following:
+
+  @examples[#:eval (make-base-eval)
+    (module t typed/racket/base
+      (require math/array typed/racket/unsafe)
+      (define array-ref1 array-ref)
+      (unsafe-provide (rename-out [array-ref1 array-ref])))
+
+    (require 't racket/contract)
+    (has-contract? array-ref)
+  ]
+
+  @; BG TODO
+  @history[#:added "1.6"]
 }
 
 @close-eval[eval]
--- a/typed-racket-lib/typed-racket/core.rkt
+++ b/typed-racket-lib/typed-racket/core.rkt
@@ -37,7 +37,7 @@
              (;; pmb = #%plain-module-begin
               [(pmb . body2) new-mod]
               ;; perform the provide transformation from [Culpepper 07]
-              [transformed-body (begin0 (remove-provides #'body2) (do-time "Removed provides"))]
+              [transformed-body (begin0 (remove-provides #'body2) (check-unsafe-reprovides #'body2) (do-time "Removed provides"))]
               ;; add the real definitions of contracts on requires
               [transformed-body
                (begin0 (change-contract-fixups (syntax->list #'transformed-body))
--- a/typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/provide-handling.rkt
@@ -12,7 +12,7 @@
          (for-syntax racket/base)
          (for-template racket/base))
 
-(provide remove-provides provide? generate-prov)
+(provide remove-provides provide? generate-prov check-unsafe-reprovides)
 
 ;; Returns #t for safe provides. Returns #f for non-provide forms
 ;; and unsafe provides for which contracts will not be generated.
@@ -27,6 +27,26 @@
   (for/list ([e (in-syntax forms)]
              #:unless (provide? e))
     e))
+
+(define (check-unsafe-reprovides forms)
+  (for ([form (in-syntax forms)])
+    (syntax-parse form
+     #:literal-sets (kernel-literals)
+     [(~and (#%provide . rest) _:unsafe-provide^)
+      (define empty-mpi (module-path-index-join #f #f))
+      (let loop ([stx #'rest])
+        (cond
+         [(identifier? stx)
+          (define b (identifier-binding stx))
+          (when (and (list? b) (not (equal? empty-mpi (car b))))
+            (fprintf (current-error-port) "unsafe-provide: cannot remove contract on imported identifier '~a'\n" (syntax-e stx)))
+          (void)]
+         [(list? (syntax-e stx))
+          (map loop (syntax-e stx))]
+         [else
+          (void)]))]
+     [_
+      (void)])))
 
 (define (freshen-id id)
   ((make-syntax-introducer) id))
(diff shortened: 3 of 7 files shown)


bennn:
re #289

stamourv:
Maybe I'm missing something from the previous discussion (it's been a while), but why not have that be the default behavior for `unsafe-provide`?

bennn:
Because I don't know how to implement this without adding the `define`s, but adding the defines to the module isn't the same semantics as `provide`.

I felt bad saying "the solution is to add defines" without also supplying the macro.

stamourv:
I don't understand the problem with adding the `define`s. And is it a problem for `unsafe-provide` to differ from `provide` in that regard?

bennn:
I do think that `unsafe-provide` should act exactly like `provide` except for not adding contracts. But if Sam agrees that adding `define`s is OK, I'll change this PR and remove `unsafe-reprovide`.

stamourv:
Maybe I'm just being dense, but what's the difference here?

bennn:
(Sam's points from #289)
The extra defines:
- increase zo size
- potentially disrupt the optimizer
- mean that re-provided identifiers aren't `free-identifier=?` to the originals

lexi-lambda:
> mean that re-provided identifiers aren't `free-identifier=?` to the originals

I think this is the key here… this is pretty different behavior from how `provide` works, and making `unsafe-provide` do that would cause a lot of confusion if I ever ran into it.

Also, this won’t work with macros, but I think that’s less of an issue, since `unsafe-provide` doesn’t make much sense with macros, anyway.

rfindler:
Not sure I am completely following here but doesn't contract-out do that?

AlexKnauth:
With `contract-out` it's not that bad because the rest of the world only sees one identifier, the one with the contract. But if `unsafe-provide` is used as an `unsafe-reprovide`-ish form, then there will two publicly available versions of the identifier, since the rest of the world has the safe version provided by the original module.

stamourv:
If you have a module that imports both the original identifier, and the one from the TR module, those *would* be different with TR's regular `provide`. So `unsafe-reprovide`'s behavior would match TR's `provide`'s in that regard, no?

AlexKnauth:
@stamourv With the current implementation of `unsafe-provide`, this `free-identifier=?` call return true:
```racket
#lang racket/base
(module a typed/racket/base
  (provide f)
  (: f : Positive-Fixnum -> Nonnegative-Fixnum)
  (define (f x) (sub1 x)))
(module b typed/racket/base
  (provide f)
  (require (submod ".." a)))
(module c typed/racket/base
  (require typed/racket/unsafe)
  (unsafe-provide f)
  (require (submod ".." a)))

(require (only-in 'b [f bf])
         (only-in 'c [f cf]))

(free-identifier=? #'bf #'cf) ; returns #t
```

stamourv:
I understand that. I was comparing TR's regular `provide` with `unsafe-reprovide` (which could become the semantics for `unsafe-provide`).

lexi-lambda:
I think the difference Alex mentions between `contract-out` and `unsafe-reprovide` is also relevant to the difference between TR’s `provide` and `unsafe-reprovide`. User code would see two identifiers that are not `free-identifier=?`, while TR’s `provide` only exposes a single identifier (even if it’s different from the one inside the module).

AlexKnauth:
@stamourv Okay. With `unsafe-provide` they are be the same identifier, but with `unsafe-reprovide` they would be different. That's an argument for keeping `unsafe-reprovide` as separate from `unsafe-provide`.

stamourv:
Maybe I'm missing something, but having an id and its unsafe equivalent not be `free-identifier=?` seems perfectly reasonable, in the same way as `contract-out`. I understand that this is not the current behavior of `unsafe-provide`, but I'm not convinced that's the right one.

AlexKnauth:
Except, that with `contract-out` the rest of the world only sees one of the identifiers. With an `unsafe-reprovide` form the rest of the world would see both. That's different enough that I think it should be an explicit opt-in, not the default.

stamourv:
They only see both if they require from both the original TR module, and the unsafe-reproviding one, no? I fail to see how that's different from requiring from both a private module (that exports without contract) and a public one (that exports with contract).

bennn:
@samth can you comment on what you think `unsafe-provide` should do?

I would still rather not have `unsafe-provide` make new definitions. @stamourv the difference to me between this scenario (having 2 public API's export the same identifier safely & unsafely) is different from the typical `contract-out` case because the latter uses a private API. If the user's requiring a private module they know there might be trouble. But if the user requires two public modules, one of which happens to unsafely reprovide an identifier from the other, the user hasn't done anything obviously wrong.

Here's also my big-picture take on the situation:
- The core issue is how to let an untyped module unsafely require typed identifiers. `unsafe-provide` seems like a nice hack for doing this
- But we don't have a clean way to let `unsafe-provide` change the "safety" of another typed module's identifiers. The easiest fix is to add `defines`, but that breaks `free-identifer=?` and adds to the bytecode and is overall an extra-ordinary thing for a `provide` to do.
- And maybe we _shouldn't_ let `unsafe-provide` change another module's API design via `unsafe-provide`; maybe there should be a separate form `unsafe-reprovide` to make the distinction clear.

(Sam raised that 3rd point in the other issue, but it didn't sink in for me until talking with Alex on Friday.)

stamourv:
I think we're going in circles, probably because we're starting from different assumptions.
How about we discuss this in person at Strange Loop / RacketCon?

bennn:
@samth @stamourv let's try to close this PR. I think the options are:

1. Merge the current PR, which: describes a "gotcha" in `unsafe-provide`, describes a work-around, adds `unsafe-reprovide` as shorthand for the work-around
2. Same as 1, but remove `unsafe-reprovide`. Just document the gotcha
3. Change the semantics of `unsafe-provide` to fix the gotcha. (Pros & cons listed above.)

My preference is for 2 at this point, but I'll change this PR to implement whichever.

AlexKnauth:
Why would 2 be better than 1? If this is a common pattern with an easily understandable meaning, why shouldn't there be an `unsafe-reprovide` form for it?

bennn:
> Why would 2 be better than 1? If this is a common pattern with an easily understandable meaning, > why shouldn't there be an unsafe-reprovide form for it?

Only because I perceive 1 and 3 as being in conflict. 2 is a band-aid over the problem.

stamourv:
I'm still in favor of 3. The gotcha in `unsafe-provide` doesn't make sense from an interface point of view (IMO), only from an implementation point of view.
But if we're unwilling to go for 3, then I think we at least need *some* solution to the problem, so 1 sounds better than 2.

rfindler:
1a: like one, but rename `unsafe-provide` to something much longer and uglier.

rfindler:
On re-reading this thread, I don't really get the identifier equality issue. An identifier can be `provide`d, and thus it will be with contract in an untyped context and without in typed. That identifier could be `unsafe-provide`d, in which case it would be without contract in both contexts.

In what way does identifier equality enter the picture?

samth:
@rfindler The example in @AlexKnauth's comment explains the identifier equality issue. Right now, `unsafe-provide` is basically just Racket's `provide`, which preserves identifier equality between inside and outside the module (but not like Typed Racket's `provide` for things defined in the current module). `unsafe-reprovide` can't work that way (the contract was attached in some previous module so we can't change it); it has to do more than `provide` usually does.

rfindler:
@samth, I see. But could TR track the original variable and use that one (expanding into a `require` and a `provide` if necessary)?

samth:
Yes, we could do that, which would be a nicer solution. I think the relevant info is already there in the binding of the identifier.

AlexKnauth:
That would limit it to only two (`free-id=?`-distinct) identifiers in the public interface instead of arbitrarily many. However, I still think this is weird enough to be an *opt-in* feature with a separate reprovide form.

rfindler:
Two identifiers that behave differently, right?

stamourv:
@AlexKnauth: I don't understand your comment about 2 vs many.

samth:
@bennn how do you feel about implementing the nicer solution described above?

bennn:
I can try to change `unsafe-reprovide` to do the nicer solution. Will give an update in January.

(If that works out, then I'll think about adding `unsafe-reprovide` vs. changing `unsafe-provide`.)

bennn:
I'm stuck. Here's a program with 2 tests that I think should pass. Test 1 checks that `unsafe-reprovide` is unsafe. Test 2 checks that it doesn't create a new identifier.

```
#lang racket/base

(module A typed/racket/base
  ;; Define a box
  (define b : (Boxof Natural) (box 1))
  (provide b))

(module B typed/racket/base
  ;; Unsafely reprovide the box from A
  (require typed/racket/unsafe)
  (require (submod ".." A))
  (unsafe-reprovide b))

(module C typed/racket
  ;; Check that A and B are using the same identifier
  (require (prefix-in A (submod ".." A)))
  (require (prefix-in B (submod ".." B)))
  (define-syntax (foo stx)
    (with-syntax ((a (free-identifier=? #'Ab #'Bb)))
      #'a))
  (define same-id? (foo))
  (provide same-id?))

(require 'B 'C rackunit)

;; Test 1 : is `b` provided unsafely?
(check-not-exn (lambda () (set-box! b -1)))

;; Test 2 : did `unsafe-reprovide` make a new identifier?
(check-true same-id?)
```

I think the goal is for `unsafe-reprovide` to:
- start with the identifier that `A` gave it (the "external handle" to the box)
- trace back to the "internal identifier" for the box
- provide the internal identifier

And I don't know how to do the tracing back. 

I tried calling `(identifier-binding #'b)`; that gives me this value but I don't know how to use it:

```
(#<module-path-index:((submod ".." A))> b.1 #<module-path-index:((submod ".." A))> b 0 0 0)
```

I also tried `(syntax-local-value #'b)` and that raised an exception `not defined as syntax (after renaming)`.

samth:
I think you need to use `(syntax-local-value/immediate #'b)`, as in the `un-rename` function in `typed-racket/typecheck/renamer` (or maybe you can just use that function directly).

bennn:
Thank you, `un-rename` works great.

I've opened a pull request to change `unsafe-provide` (#657). With that, there's no need for an `unsafe-reprovide` so I'm closing this one.

PR #440, opened 2016-10-11
add typed/file/glob

bennn:
Typed interface to `file/glob`.

This PR is "ready to go", but if we leave it open I'll try to add types + tests for the other `file/*` libraries by the end of the month.

(Otherwise I'll still try to add those, just in a different PR)

### DIFF
--- a/typed-racket-more/typed/file/glob.rkt
+++ b/typed-racket-more/typed/file/glob.rkt
@@ -1,0 +1,16 @@
+#lang typed/racket/base
+
+(require typed/racket/unsafe)
+
+(define-type Glob (U Path-String (Sequenceof Path-String)))
+
+(unsafe-require/typed file/glob
+  [glob (->* [Glob] [#:capture-dotfiles? Boolean] (Listof Path-String))]
+  [in-glob (->* [Glob] [#:capture-dotfiles? Boolean] (Sequenceof Path-String))]
+  [glob-match? (->* [Glob Path-String] [#:capture-dotfiles? Boolean] Boolean)]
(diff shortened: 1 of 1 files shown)


typed-racket-more/typed/file/glob.rkt
```diff
@@ -0,0 +1,16 @@
...
+
+(define-type Glob (U Path-String (Sequenceof Path-String)))
+
+(unsafe-require/typed file/glob
+  [glob (->* [Glob] [#:capture-dotfiles? Boolean] (Listof Path-String))]
+  [in-glob (->* [Glob] [#:capture-dotfiles? Boolean] (Sequenceof Path-String))]
+  [glob-match? (->* [Glob Path-String] [#:capture-dotfiles? Boolean] Boolean)]
+  [glob-capture-dotfiles? (Parameterof Boolean)])
+
+(unsafe-provide glob
```
samth: We definitely don't need both `unsafe-require` and `unsafe-provide`. Presumably no one will use this library from untyped code.

typed-racket-more/typed/file/glob.rkt
```diff
@@ -0,0 +1,16 @@
+#lang typed/racket/base
+
+(require typed/racket/unsafe)
+
+(define-type Glob (U Path-String (Sequenceof Path-String)))
+
+(unsafe-require/typed file/glob
```
samth: Does this need to be unsafe?

bennn: No, but the functions already have contracts.

samth:
Looks good to merge.

PR #445, opened 2016-10-21
Makes free-vars use struct properties

pnwamk:
This PR cleans up the struct property code in `rep-utils.rkt` and uses struct properties for a few things, namely:
1. the `free-vars` and `free-idxs` fields from Reps have been removed and are now computed via struct property functions.
2. the code which parses fold-like functions for reps in `rep-utils.rkt` has been cleaned up and consolidated to use a single set of macros (instead of one per fold-like function).

Also, with these changes in place we're one step closer to removing ubiquitous interning.

### DIFF
--- a/typed-racket-lib/typed-racket/env/init-envs.rkt
+++ b/typed-racket-lib/typed-racket/env/init-envs.rkt
@@ -59,11 +59,11 @@
 
 ;; Compute for a given type how many times each type inside of it
 ;; is referenced
-(define (compute-popularity ty)
-  (when (Type? ty)
-    (hash-update! pop-table ty add1 0))
-  (when (walkable? ty)
-    (Rep-walk compute-popularity ty)))
+(define (compute-popularity x)
+  (when (Type? x)
+    (hash-update! pop-table x add1 0))
+  (when (Rep? x)
+    (Rep-walk compute-popularity x)))
 
 (define (popular? ty)
   (> (hash-ref pop-table ty 0) 5))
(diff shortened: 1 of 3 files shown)


samth:
What's the performance impact?

pnwamk:
@samth I isolated the effect that changing instantiate/abstract had and it was a hair slower =\

So I tossed that change. Now it's just struct-prop functions for free-vars and free-idxs and a cleaner ```rep-utils.rkt```

pnwamk:
(Note: the moving free-vars and free-idxs from being fields to struct props is primarily to get ready for less interning and make them be pay-as-you-go operations and not pre-computed values)

pnwamk:
For performance difference on what is remaining in this PR (just moving free-vars/idxs from fields to generics) we seem to be somewhere around 0%-4% faster across the 4 tests files I tried (math-flonum.rkt, new-metrics.rkt, slowcheck.rkt, and schml-specify-rep.rkt).

pnwamk:
@samth did you get a chance to finish looking at this?

samth:
Use int-err, not error.

Would it be faster to use subtyping for resolvable? Or just a linear sequence of checks?

Otherwise, looks good.

pnwamk:
A sequence of checks is faster. Will adjust that along w/ int-err and merge.

pnwamk:
Similar (but slightly different) changes to these are included in a PR I'm working on which also removes ubiquitous interning... it'll be easier to just submit that PR than to coordinate the changes between the two.

PR #446, opened 2016-10-22
remove contract overhead from static-contract code

pnwamk:
This PR changes numerous ```(provide (contract-out ...))``` instances to ```(provide/cond-contract ...)``` and removes some unneeded generic ```dict``` operations with more specific functions for performance sake.

Reducing this contract overhead in the ```static-contract``` directory and especially changing the particularly hot ```dict-set``` instances to ```free-id-table-set``` calls from ```syntax/private/id-table``` made _huge_ differences when type checking some parts of the Schml compiler (see this standalone [file](https://github.com/pnwamk/tr-performance/blob/master/schml-interp-casts-help.rkt)).

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/combinators/any.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/combinators/any.rkt
@@ -4,16 +4,18 @@
 ;; Allows optimizations as many combinators can be simplified if their arguments are any/sc
 ;; Ex: (listof/sc any/sc) => list?/sc
 
-(require "../structures.rkt" "../constraints.rkt"
+(require "../../utils/utils.rkt"
+         "../structures.rkt"
+         "../constraints.rkt"
          racket/match
-         racket/contract
+         (contract-req)
          (for-template racket/base racket/contract/base)
          (for-syntax racket/base syntax/parse))
 
-(provide
-  (contract-out
-    [any/sc static-contract?])
-  any/sc:)
+(provide/cond-contract
+ [any/sc static-contract?])
+
+(provide any/sc:)
 
 
 ;;Printing
--- a/typed-racket-lib/typed-racket/static-contracts/combinators/case-lambda.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/combinators/case-lambda.rkt
@@ -3,22 +3,24 @@
 ;; Static contract for case->.
 ;; Like case-> doesn't support keyword arguments.
 
-(require "../structures.rkt" "../constraints.rkt"
+(require "../../utils/utils.rkt"
+         "../structures.rkt"
+         "../constraints.rkt"
          racket/list racket/match
-         racket/contract
+         (contract-req)
          (for-template racket/base racket/contract/base)
          (for-syntax racket/base syntax/parse))
 
-(provide
-  (contract-out
-    [case->/sc ((listof arr-combinator?) . -> . static-contract?)]
-    [arr/sc (-> (listof static-contract?)
-                (or/c static-contract? #f)
-                (or/c (listof static-contract?) #f)
-                static-contract?)])
-  case->/sc:
-  arr/sc:
-  (rename-out [arr-combinator? arr/sc?]))
+(provide/cond-contract
+  [case->/sc ((listof arr-combinator?) . -> . static-contract?)]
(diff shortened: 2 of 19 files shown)


pnwamk:
Whoops. Should be fixed by 4bfebbda.

rfindler:
The dict contracts are a lot more expensive than others because of the
dependency. Did you try with contracts that were based on -> instead of ->i
?

samth:
@rfindler Do you mean try changing the `racket/dict` library to use different contracts?

rfindler:
I may have been mixed up here: I'm saying that if uses of contracted operations with `->i` were used and replaced with operations that don't have contracts, it may be good enough to replace them with operations that have `->` based contracts. I know that `free-id-table-set`'s contract is especially expensive.

pnwamk:
That would be interesting to measure -- I did not try that while debugging and developing this PR.

rmculpepper:
The dependent contracts in `racket/dict` are my fault, and as far as I know my libraries are the primary clients of the functionality they support. Maybe we should just say they are too expensive for the benefit they provide and go back to simple `->` contracts instead.

rfindler:
I think they are wonderful contracts and I need to figure out how to get
good performance out of them.

But if you didn't want to wait for me, I understand ....

samth:
Just for context, the performance difference that @pnwamk mentioned was something like changing 20 minute compile times to 4 minute compile times. So this has a big impact on Typed Racket in practice currently.

rfindler:
On Tue, Oct 25, 2016 at 6:12 PM, Sam Tobin-Hochstadt <

> Just for context, the performance difference that @pnwamk
> <https://github.com/pnwamk> mentioned was something like changing 20
> minute compile times to 4 minute compile times. So this has a big impact on
> Typed Racket in practice currently.
>
>
Yes, I believe it. Last time I looked into this, that contract was
significant. I made some tweaks to improve it in a special case, but I
really think that the problem isn't in dependency inherently, but in the
particular abstractions that I chose to implement it.

.... not that I know how to fix it yet, mind you :)

(and I don't want to add anything to anyone's plate, but if that experiment
could be done with -> contracts that have any/c or even something more
specific in there, I'd be curious about the results)

pnwamk:
I threw together a quick experiment examining how various levels of contracted-ness effect the performance of ```dict-set```:

Results:
```
normandy:Experiments pnwamk$ racket dictcontractcomparison.rkt
standard (dependent) contract for dict set
cpu time: 1263 real time: 1266 gc time: 5

simple (non-dependent) contract for dict set
cpu time: 274 real time: 276 gc time: 2

no contract (racket/private/dict) for dict set
cpu time: 104 real time: 104 gc time: 1
```

Code:

```racket
#lang racket

;; define a raw dict-set operation
;; raw-dict-set
(module raw racket/base
  (require racket/private/dict)
  (provide (rename-out [dict-set raw-dict-set])))
(require 'raw)

;; define a dict-set operation w/ a simple contract
;; dict-set
(module simple racket/base
  (require racket/private/dict
           racket/contract)
  (provide (contract-out [dict-set (-> (and/c dict? immutable?) any/c any/c
                                       (and/c dict? immutable?))])))
(require 'simple)

;; define a dict-set operation w/ the current (dependent contract)
;; dep-dict-set
(module dependent racket/base
  (require racket/dict)
  (provide (rename-out [dict-set dep-dict-set])))
(require 'dependent)

(collect-garbage)
(collect-garbage)

(define d (hash))


;; timing tests


(displayln "standard (dependent) contract for dict set")
(time (for ([_ (in-range 1000000)])
        (dep-dict-set d 42 #t)))


(collect-garbage)
(collect-garbage)

(displayln "simple (non-dependent) contract for dict set")
(time (for ([_ (in-range 1000000)])
        (dict-set d 42 #t)))

(collect-garbage)
(collect-garbage)

(displayln "no contract (racket/private/dict) for dict set")
(time (for ([_ (in-range 1000000)])
        (raw-dict-set d 42 #t)))
```

samth:
Thanks for experimenting! What happens for hash-set directly?

pnwamk:
code & results updated to include ``hash-set``

rfindler:
Thank you.

But I apparently wasn't clear. I have done this particular experiment many
many times. My question was meant to be about the larger app you were
running that used those operations.

Sorry about that.

PR #448, opened 2016-10-25
fix intersect, make it (hopefully) commutative

pnwamk:
These changes attempt to make computing type intersections a more commutative operation.

### DIFF
--- a/typed-racket-lib/typed-racket/infer/intersect.rkt
+++ b/typed-racket-lib/typed-racket/infer/intersect.rkt
@@ -10,73 +10,116 @@
 (import infer^)
 (export intersect^)
 
-
 ;; compute the intersection of two types
-;; (note: previously called restrict)
+;; (note: previously called restrict, however now restrict is
+;;  a non-additive intersection computation defined below
+;; (i.e. only parts of t1 will remain, no parts from t2 are added))
 (define (intersect t1 t2)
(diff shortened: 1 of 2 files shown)


pnwamk:
This came up while debugging why the ```wrap``` package was broken after some recent TR commits.

samth:
Can you describe what the new approach is here?

pnwamk:
The two main differences are as follows:

1. When we're intersecting two unions, we do so by intersecting their cartesian product from the start (this produces more consistent outputs which are not order dependent).

2. Previously ```intersect``` tried to build a more specific recursive type only when we were intersecting two μ-types. This has been expanded to include all types which can be recursive (i.e. μ's, Names, and Apps) and there is a little bit of back-channeling to report whether or not a recursive type is actually required.

I tried to add comments describing all of the new behavior (since w/o comments the 'why' we're doing it that way would clearly not be apparent).

pnwamk:
3. There is now an explicit check for the ```Error``` type (since that breaks what would otherwise be reasonable, mostly commutative things like checking for subtyping, etc)

samth:
This might be a good case for some random testing, but that can wait for a separate PR.

samth:
I think the discussion in bullet 1 should appear in a comment, otherwise looks good.

PR #449, opened 2016-10-27
less overhead in contract gen

pnwamk:
This PR introduces lightweight interfaces to hash tables that allow us to treat them as tables/sets which use free-identifier=? as the equality predicate.

Swapping these data structures in for those provided by `syntax/private/id-table` and `syntax/private/id-set` reduce contract generation time from 12.5 seconds to 4.5 seconds in a [test](https://github.com/pnwamk/tr-performance/blob/master/schml-interp-casts-help.rkt) case which stresses contract generation in Typed Racket.

`utils/fid-hash.rkt` contains a lightweight definition for free-identifier=? hash tables

`utils/fid-hset.rkt` contains a lightweight definition for free-identifier=? sets

`utils/hset.rkt` contains a lightweight definition for standard sets

The changes were mostly made in the `static-contract` sub-directory -- I have not investigated how these data structures might affect type checking time if used in other places inside Typed Racket.

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/combinators/simple.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/combinators/simple.rkt
@@ -10,6 +10,7 @@
   "../structures.rkt"
   "../constraints.rkt"
   racket/match
+  (only-in racket/unsafe/ops unsafe-fxior)
   (contract-req))
 
 (provide/cond-contract
@@ -55,13 +56,13 @@
                  (recur (simple-contract-name s1)
                         (simple-contract-name s2))))
           (define (hash-proc sc hash-code)
-            (hash-code (list (syntax->datum (simple-contract-syntax sc))
-                             (simple-contract-kind sc)
-                             (simple-contract-name sc))))
+            (unsafe-fxior (hash-code (syntax->datum (simple-contract-syntax sc)))
+                          (unsafe-fxior (hash-code (simple-contract-kind sc))
+                                        (hash-code (simple-contract-name sc)))))
           (define (hash2-proc sc hash-code)
-            (hash-code (list (syntax->datum (simple-contract-syntax sc))
-                             (simple-contract-kind sc)
-                             (simple-contract-name sc))))]
+            (unsafe-fxior (hash-code (syntax->datum (simple-contract-syntax sc)))
+                          (unsafe-fxior (hash-code (simple-contract-kind sc))
+                                        (hash-code (simple-contract-name sc)))))]
         #:methods gen:sc
          [(define (sc-map v f) v)
           (define (sc-traverse v f) (void))
--- a/typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt
@@ -7,6 +7,7 @@
          "../structures.rkt"
          "../constraints.rkt"
          racket/match
+         (only-in racket/unsafe/ops unsafe-fxior)
          (for-syntax racket/base racket/syntax syntax/stx syntax/parse)
          racket/set
          racket/sequence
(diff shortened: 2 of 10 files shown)


samth:
It's unfortunate that using the built-in id tables and sets is so much overhead. @rmculpepper, do you have any ideas on reducing this overhead? @mflatt, is there a good way to tell where this overhead is coming from? I think the implementation strategies are pretty similar.

rfindler:
This isn't the contracts?

samth:
This is a further improvement on top of the improvement from not using the contracts.

pnwamk:
I did some timing (just doing `free-id-table-set` in a loop) and got the following results:

```
standard contract for free-id-table-set
cpu time: 749 real time: 752 gc time: 6

no contract (syntax/private/id-table) for free-id-table-set
cpu time: 660 real time: 664 gc time: 3

fid-hash-set (i.e. just a raw hash table w/ functions that treat it like a free-id-table)
cpu time: 114 real time: 117 gc time: 0
```

After seeing that difference I made the changes in this PR to Typed Racket to see how much it would affect contract generation.

pnwamk:
(sorry for close/reopen -- buttons are too easy to accidentally click sometimes...)

rmculpepper:
I noticed that `fid-hash-set` etc use `syntax-e` to convert the identifier into a symbol key, which is wrong; it will miss renamed references. Try using `identifier-binding-symbol` instead and see if the gap closes.

Another thing to check is whether the functions like `id-table-set` in `syntax/private/id-table`, which take symbol-producing and comparison functions as arguments, are getting inlined so those calls get specialized to direct calls.

pnwamk:
I updated the ```table-set``` benchmarks above -- fixing my function to be correct made it more or less the same as the versions from Racket's library (or as close as you would expect, perhaps).

Iterating over the the elements of the tables, however, still has significant differences:

```
standard contract for in-free-id-table set
cpu time: 600 real time: 601 gc time: 1

no contract (syntax/private/id-table) for in-free-id-table set
cpu time: 479 real time: 481 gc time: 2

fid-hash-set (nested in-hash-values + in-list)
cpu time: 68 real time: 68 gc time: 0
```

Those numbers are generated by this code (module imports/exports to rename bindings excluded):

```racket
;; loop tests
(define d2 (make-immutable-free-id-table (hash #'a 'a #'b 'b #'c 'c #'d 'd #'e 'e #'f 'f
                                               #'g 'g #'h 'h #'i 'i #'j 'j #'k 'k #'l 'l)))
(define raw-d2 (make-raw-immutable-free-id-table (hash #'a 'a #'b 'b #'c 'c #'d 'd #'e 'e #'f 'f
                                                       #'g 'g #'h 'h #'i 'i #'j 'j #'k 'k #'l 'l)))
(define h2 (fid-hash #'a 'a #'b 'b #'c 'c #'d 'd #'e 'e #'f 'f
                     #'g 'g #'h 'h #'i 'i #'j 'j #'k 'k #'l 'l))

(displayln "\nstandard contract for in-free-id-table set")
(time (for* ([_ (in-range 100000)]
             [(id val) (in-dep-free-id-table d2)])
        (unless (syntax? id) (error 'foo))))


(collect-garbage)
(collect-garbage)

(displayln "\nno contract (syntax/private/id-table) for in-free-id-table set")
(time (for* ([_ (in-range 100000)]
             [(id val) (in-raw-free-id-table raw-d2)])
        (unless (syntax? id) (error 'foo))))

(collect-garbage)
(collect-garbage)


(displayln "\nfid-hash-set (nested in-hash-values + in-list)")
(time (for* ([_ (in-range 100000)]
             [bucket (in-hash-values h2)]
             [id/val (in-list bucket)]
             [id (in-value (car id/val))]
             [val (in-value (cdr id/val))])
        (unless (syntax? id) (error 'foo))))
```

Is there a faster way to iterate over the key/value pairs in a free-id-table? (or is my version cheating again?). Clearly the hash version is uglier (but the time differences are significant enough I'd be willing to deal with that if it came to it).

pnwamk:
@rmculpepper It looks like the biggest bottleneck for us operation wise is that for-loop iteration is unacceptably slow (compared to what you can do working with the underlying data structures). If that isn't reasonably addressable, would it make sense to add performant definitions of `for-each` and `fold` for the `id-tables`? (I'd be willing to take a stab at that -- currently there is only a `for-each` and `map` definition, however they just call off to the equivalent `dict` operations and perform very poorly).

rmculpepper:
`in-free-id-table` is implemented as a function using `make-do-sequence`. You could try implementing it with `define-sequence-syntax` and see how much a difference that makes. I don't know if it's possible for sequence syntax to expand into nested sequence traversals, though, so I'm not sure you'll be able to do as well as your `for*` traversal suggests. See also the comments in `syntax/private/id-table` before `id-table-iterate-first` about recovering from mutation during iteration.

rmculpepper:
Fixing the existing specialized traversals (`for-each` and `map`) sounds like a good idea. Adding `fold` seems reasonable too.

pnwamk:
I'm going to close this for now and look into implementing more efficient iteration functions for `id-tables`.

pnwamk:
@rmculpepper the following is my first attempt at a `define-sequence-syntax` definition for `in-*-id-table`. It is just as fast as the raw hash-table traversal code I wrote earlier (although this initial version doesn't yet do any special reasoning for mutation during iteration, impersonators, etc)

```racket
(define (in-id-table/proc t)
  (for/list ([(k v) (in-id-table t)]) (cons k v)))

(define-sequence-syntax in-id-table
  (lambda () #'in-id-table/proc)
  (lambda (stx)
    (syntax-case stx ()
      [[(key val) (_ table)]
       #'[(key val)
          (:do-in
           ;; outer-id bindings
           ([(h) (id-table-hash table)]
            [(init-idx) (hash-iterate-first (id-table-hash table))])
           #true ;; outer-check
           ;; loop-id's and initial values
           ([idx init-idx]
            [bucket (and init-idx (hash-iterate-value h init-idx))])
           ;; pos-guard
           idx
           ;; inner-ids
           ([(key val) (values (caar bucket) (cdar bucket))]
            [(next-idx next-bucket)
             (if (null? (cdr bucket)) ;; NOTE: hash buckets must never be completely empty!
                 (let ([next-idx (hash-iterate-next h idx)])
                   (values next-idx (and next-idx (hash-iterate-value h next-idx))))
                 (values idx (cdr bucket)))])
           #true ;; pre-guard
           #true ;; post-guard
           ;; recursive call args
           [next-idx next-bucket])]]
      [_ #f])))
```

PR #450, opened 2016-10-28
type for `hash-empty?`

philnguyen:
This commit attempts to add type for `hash-empty?`. I don't really know what I'm doing, just guessing that `base-env.rkt` is the only place that needs editing.

Also, it looks like `hash-count` only needs to know that its argument is `-HashTop` and there's no need for  the `(-poly (a b) ((-HT a b) -Index))` line, but I'm not sure if I would break anything so didn't touch it.


samth:
That looks good, and I agree about hash-count.

AlexKnauth:
Wouldn't that new non-polymorphic type for `hash-count` break programs that use `(inst hash-count ...)`?

stamourv:
Hmm, good point. Bleh.
Should we revert that bit then?

pnwamk:
I wonder what the odds are that someone wrote such an instantiation.

If people only wrote `inst` so that things could typecheck, you would think no such `inst` usages existed (since inference would always succeed). But... people are funny ;-)

rfindler:
I've been suggesting this kind of thing for a long time, so maybe I should stop :) but I think that the questions of backwards compatibility are tricky and can be studied and something good can be done.

In this case, you question Andrew, makes me wonder: what if people who cared about their TR code could provide access to it and then you could answer questions like that. And maybe do other things with it that would influence the design of TR by looking at it.

More than just the pkg server, people could provide links to random git repos or code you could snapshot somewhere and then there could be some measure of how bad a particular backwards compatibility change would be. Or you could think about providing tool support or maybe other ideas would occur if you looked seriously at the code that was out there.

PR #452, opened 2016-10-30
less dynamic dispatch in contract gen

pnwamk:
The main point of this PR is to get rid of `dict` operations in favor of the appropriate data structure-specific variants. This is a noticeable improvement for the type checking times of some programs now that `in-free-id-table` is more performant (racket commit `82204d1` "faster in-*-id-table (#1499)").

It also consolidates a few computations and uses a lightweight implementation of sets which uses hash tables directly to reduce unnecessary overhead.


pnwamk:
This fails at the moment because of a bug w/ `free-id-table-ref!`. The bug has already been [fixed](https://github.com/racket/racket/commit/df2b1dad45a18b6d3302db69f81394d744d6b67e), but the snapshot build it is using today is from 20161026 (the bug fix was on 20161028).

rfindler:
The NWU snapshot build machine is having trouble. Sorry.

rfindler:
FYI: the NWU snapshots completed successfully today. (Finally.)

pnwamk:
Great - thanks for the heads up!

typed-racket-lib/typed-racket/static-contracts/constraints.rkt
```diff
@@ -204,7 +203,9 @@
     [(list (contract-restrict vs rec constraints) ...)
      (contract-restrict (merge-kind-maxes min vs)
                         (free-id-table-union rec)
-                        (apply set-union (set) constraints))]))
+                        (for/fold ([h (set)])
```
pnwamk: it is not (I wrote it that way when comparing sets versus hash tables and forgot about it) -- I'll revert that before pushing.

Issue #460, opened 2016-11-25
Return type of in-query wrong for single-column queries?


typeundeclared:
Using Racket 6.7.

I am attempting to convert a small piece of racket code to typed racket.  For the moment, assume there is a tasks table with a text column 'program'.  I had a piece of code like so:

    (define (fetch-programs db-conn)
      (for/list ([p (in-query db-conn
                              "select program from tasks group by program")])
        p))
    
However, I had problems getting this to type-check in racket.  Trying to make things as simple as possible, I tried this repl:

    (sequence->list (in-query db-conn "select program from tasks group by program"))

I get back the following error:

    in-query: broke its own contract;
     promised a vector,
      produced: "a-sample-string"
      in: an element of
          the range of
          (->*
           (any/c any/c)
           (#:fetch
            any/c
            #:group
            any/c
            #:group-mode
            any/c)
           #:rest
           (listof Any)
           (sequence/c (vectorof Any)))
      contract from: (interface for in-query)
      blaming: (interface for in-query)
       (assuming the contract is correct)
      at: <pkgs>/typed-racket-more/typed/db/base.rkt:72.2
  
In a non-typed-REPL, that operation works (and returns a list of strings).  It appears as though in-query, when run with a query that returns a single column of values, doesn't return a sequence of vectors of length one, but instead returns a sequence of single values.

I'm relatively new to Typed Racket---am I missing something?  Thanks in advance for your help!

wargrey:
Sorry, this is a limitation of Typed Racket when generating contract.

vector is a mutable data, therefore meanwhile, typed db API does not allow vectors as return value.

wargrey:
Sorry, my fault. I have forgot why I left the `in-query` in the API with the wrong annotation.

Typed Racket cannot describe this function since it returns non-fixed multiple values, so you have to change your code to work with other query APIs, say `query-rows` which aways returns a list of vectors.

typeundeclared:
Much appreciated (both the fix and the explanation).  I'll use query-rows going forward.

PR #461, opened 2016-11-27
Bugfix: remove (in-query)

wargrey:
it returns non-fixed multiple values which cannot be described by Typed Racket, or not recommend to do so.

### DIFF
--- a/typed-racket-more/typed/db/base.rkt
+++ b/typed-racket-more/typed/db/base.rkt
@@ -68,13 +68,7 @@
                  [#:group SQL-Group]
                  [#:group-mode (Listof Null-Mode)]
                  SQL-Datum *
-                 (Listof (Vectorof SQL-Datum)))]
- [in-query (-> Connection Statement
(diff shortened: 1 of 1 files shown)


wargrey:
fixed #460

samth:
Can't we just handle the single-value case?

samth:
Actually, could we use `AnyValues` to handle this correctly? Either that, or `(values SQL-Datum ...)`

wargrey:
I think, no.

Actually, `in-query` returns a sequence of `AnyValues` via `vector->values`.
Even though `vector->values` can be annotated with `(-> VectorTop AnyValues)`(We do not have it in Typed Racket right now, nor allowed by `require/typed`), it can only produce one or two values to satisfy the `(Sequenceof a)` and/or `(Sequenceof a b)`, both are not quite useful for real world database client applications.

From the perspective of an application, there will be a further step to asserts the datatypes of selected columns. `match` is one of the best choices, application always knows the exact types and their orders, Typed Racket always duplicates the work.

PR #464, opened 2016-12-01
Add type (: max (Index * -> Index))

tim-brown:
Currently, `(max Index Index)` has the type `Nonnegative-Fixnum` from the case
`(-> Nonnegative-Fixnum Fixnum * Fixnum Nonnegative-Fixnum)`
whereas I believe it should be `Index` -- this includes the case where (max 0 0) is 0 -- which is an Index
added `(-> Index * Index)` to second the varop list for the type of `max`

Is the testing automatic? Or do I need to write a test case --- if so where?


samth:
This looks great, thanks.

samth:
By the way, tests for this would be in typed-racket-test/unit-tests/typecheck-tests.rkt

PR #467, opened 2016-12-04
Add #lang typed/racket/gui/no-check

wargrey:
Okay, the test case for this `#lang` does not work since it will instantiate the gui instance one more time.


samth:
What test case did you try, and where did it fail?

wargrey:
```racket
#lang typed/racket/gui/no-check
 
 (: f (Integer -> Any))
 (define (f x) (add1 (current-eventspace)))
 
 (lambda (#{x : String}) (string-append " " x))
```

It's okay in the local machine, the failure is reported by travis-ci due to the "instantiate the gui instance one more time". That's the reason I deleted this branch, thought it will never pass in the continuous integration environment.

samth:
What about a module like this:

```
#lang racket

(module foo typed/racket/gui/no-check
   ....)
```

That way it will type check, but won't run.

PR #468, opened 2016-12-06
`in-query` returns a sequence of single value instead of result-arity-based multiple values

### LINKED ISSUE #460: Return type of in-query wrong for single-column queries?

wargrey:
Replaced #461 and fixed #460 

Currently, Typed Racket denies sequences that provide 3 or more values,
thus, `in-query` always returns a sequence of vector instead of value/values.
It's okay since applications may predicate column types with `match` after receiving data.

Also, this fix makes the `SQL-Datum` more precision,
since `Any` shadows all opaque types (such as `SQL-Null`) which lead to contract errors.

### DIFF
--- a/typed-racket-more/typed/db/base.rkt
+++ b/typed-racket-more/typed/db/base.rkt
@@ -1,24 +1,26 @@
-#lang typed/racket
+#lang typed/racket/base
 
 (provide (all-defined-out))
 
+(require (for-syntax racket/base))
 (require (for-syntax racket/syntax))
 
 (define-type Data-Source data-source)
 (define-type Simple-Result simple-result)
 (define-type Rows-Result rows-result)
 
-(define-type Null-Mode (U 'preserve-null 'list))
+(define-type Group-Mode (U 'preserve-null 'list))
 (define-type Isolation-Level (U 'serializable 'repeatable-read 'read-committed 'read-uncommitted False))
 (define-type Schema-Option (U 'search-or-current 'search 'current))
 
-(define-type SQL-Datum Any)
+(define-type SQL-Datum (U Boolean String Real Char Bytes SQL-Null))
 (define-type SQL-Type (List Boolean (Option Symbol) SQL-Datum))
 (define-type Statement (U String Prepared-Statement Virtual-Statement Statement-Binding))
 
 (define-type SQL-Field (U String Natural))
 (define-type SQL-Grouping (U SQL-Field (Vectorof SQL-Field)))
-(define-type SQL-Group (U SQL-Grouping (Vectorof SQL-Grouping)))
+(define-type SQL-Group (U SQL-Grouping (Listof SQL-Grouping)))
+(define-type SQL-Dictionary (HashTable SQL-Field (U (Vectorof SQL-Datum) (Listof (Vectorof SQL-Datum)))))
 
 (require/typed/provide
  db/base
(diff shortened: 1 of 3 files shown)


samth:
This makes Typed Racket's `in-query` behave differently from regular `in-query`, right? I think that's not a good idea.

Maybe @rmculpepper has a suggestion for a better API that might be more reasonable to type check.

wargrey:
I've found an interesting situation that `in-query` returns a sequence of single value is useful: design tables only contain fields for indexing with another field to store the full content which can be `read` back as an instance of a prefab struct.

samth:
In that case, we could have things be compatible, and just only support single-valued query results, right?

wargrey:
In my opinion, yes if working with Typed Racket means using a more strict formal method to define the software. I like this approach.

For a new project, if starting with Typed Racket, developers can find their best practice which also satisfy the untyped case, but not vice versa. This situation is not the only example, other is, say, all the arithmetic operators in Typed Racket are only accepts 2 floats.

samth:
I agree that this wouldn't be optimal for people who want to write queries from Typed Racket, and we should provide a better interface. But I think that should be a different interface, and not the same name as one from `db` but with different behavior.

Also, I wish in general people wouldn't write dynamically-valued interfaces like this, since they're almost impossible to type.

rmculpepper:
In case it helps: when used immediately within a `for` rhs, `in-query` knows how many values the `for` clause expects to receive, and it passes that information to `in-query-helper`, which blows up if the statement produces a different number of columns. So that case of `in-query-helper` has the following dependent type (inventing notation):

    in-query-helper : n:Nat Conn Stmt blah blah -> (Sequenceof (Valuesof[n] SqlDatum))

wargrey:
All right, here is the trade-off solution with an example as a lightweight _Data Access Layer_.

If someone does not like this example, there is always a `(in-list (query-rows))`, or the _dbms_ may allow you to write some procedure to combine multiple fields into one.

Anyway, in the future, either a better interface is found or the `(Sequenceof (Valuesof[n] SQL-Datum))` is supported, this solution will be improved.

PR #469, opened 2016-12-06
remove ubiquitous interning

pnwamk:
Major changes:

Interning: All Reps are no longer interned. Right now we only intern unions and some props.

Rep generic operations: we use Racket's generics.

Singletons: Reps can be declared singleton, which creates a single
instance of the rep and all visible operations (even within the
declaring module) are defined in terms of that singleton instance
(e.g. predicate is defined as (λ (x) (eq? x singleton-instance)), etc).

Custom constructors: A custom constructor can be specified for a Rep,
which allows for simple normalization or other invariants to be
enfored universally when a Rep is created (they should be lightweight).

Union: Unions now only enforce the following invariants on their
members: (1) No duplicates (by virtue of using a hash-based set), (2)
Any and Nothing do not appear in unions, (3) Nested unions are
flattened

Contract generation: When generating contracts for unions, we
(recursively) reduce the types so that there is no obvious overlap
between types w.r.t. subtyping. See union.rkt for 'normalize-type'

Subtyping changes: Subtyping has been designed to reduce dispatch time
w/ a switch since we no longer cache all subtyping calls (we only
cache subtyping results for unions)

Testing: More subtype tests have been added (we are still relying too
much on typecheck-tests.rkt and not the more granular unit tests).
Typecheck-tests has been changed to check for type-equivalence (i.e.
subtyping in both directions) instead of equal?)


samth:
I sort of wish this could be split into the renaming and singleton-using commit, and the other stuff. Not sure how to do that easily, though.

typed-racket-lib/typed-racket/rep/rep-utils.rkt
```diff
@@ -24,159 +25,245 @@
...
 
-(define (hash-name name)
-  (if (identifier? name)
-      (hash-id name)
-      name))
 
 (begin-for-syntax
+  ;; defines a rep transformer where the tranforming function is
+  ;; a parameter (e.g. free vars)
+  (define (rep-transform self f-id struct-fields body)
```
samth: Can you explain this more?

typed-racket-lib/typed-racket/rep/rep-utils.rkt
```diff
@@ -24,159 +25,245 @@
...
-           c))))
+(define-values (next-uid! get-uid-count)
+  (let ([state 0]
+        [finalized? #f])
+    (values (λ () (if finalized?
+                      (int-err "next-uid! called after uid count finalized!")
+                      (begin0 state (set! state (add1 state)))))
+            (λ () (set! finalized? #t) state))))
+
+(define-syntax (define-switch stx)
```
samth: Seems like a lot of this could be a separate file (with a little documentation).

typed-racket-lib/typed-racket/rep/rep-utils.rkt
```diff
@@ -208,31 +295,25 @@
...
+       (~optional [#:frees . (~var frees-spec (freesspec #'(flds.ids ...)))])
+       ;; #:for-each spec (how to traverse this structure for effect)
+       (~optional [#:for-each . (~var for-each-spec (generic-transformer #'(flds.ids ...)))])
        ;; #:fold spec (how to transform & fold this structure)
-       (~optional [#:fold . (~var fold-spec (foldspec #'var.name
-                                                      #'var.mexpdr
-                                                      #'(flds.ids ...)))])
-       (~optional [#:type-mask . type-mask-body])
-       ;; is this a Type w/ a Top type? (e.g. Vector --> VectorTop)
-       (~optional [#:top top-pred:id])
```
samth: How do we now get this info?

typed-racket-lib/typed-racket/rep/rep-utils.rkt
```diff
@@ -242,141 +323,156 @@
...
+                       uid-id)))]
+         [else
+          (syntax/loc stx
+            (begin
+              (define uid-id (next-uid!))
+              struct-def
+              constructor-def
+              mexpdr-def
+              provides ...
+              (provide uid-id)))]))]))
```
samth: Why does `uid-id` get provided?

typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt
```diff
@@ -160,9 +161,9 @@
       [(? resolvable?)
        (tc/funapp f-stx args-stx (resolve-once f-type) args-res expected)]
       ;; a union of functions can be applied if we can apply all of the elements
-      [(Union: (and ts (list (? Function?) ...)))
+      [(Union: ts) #:when (for/and ([t (in-hset ts)]) (Function? t))
```
samth: We could probably avoid traversing the hset twice. Maybe not worth it right now, but something to think about.

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -400,6 +400,8 @@
           [else (int-err "Two conflicting definitions: ~a ~a" def other-def)]))
       (dict-update h (binding-name def) merge-def-bindings #f)))
   (do-time "computed def-tbl")
+  ;; check that all parsed apps are sensible:
+  (check-registered-apps!)
```
samth: There are plenty of other places where types get parsed. Are we sure that this is after all of them in this code? And the check probably needs to be added elsewhere.

typed-racket-lib/typed-racket/types/abbrev.rkt
```diff
@@ -12,12 +12,18 @@
...
-         (types union numeric-tower prefab)
+         (types numeric-tower prefab)
          ;; Using this form so all-from-out works
          "base-abbrev.rkt" "match-expanders.rkt"
 
+         ;; signature env req here is so it is statically required by
+         ;; the code loaded during typechecking, otherwise we get
+         ;; a `reference to a module that is not available` error
+         ;; from references generated by init-envs
+         (env signature-env)
```
samth: Which identifier is missing from `signature-env`?

typed-racket-lib/typed-racket/types/prop-ops.rkt
```diff
@@ -70,69 +71,62 @@
...
 
 ;; does p imply q? (but only directly/simply)
 (define (implies-atomic? p q)
   (match* (p q)
     ;; reflexivity
-    [(_ _) #:when (or (prop-equal? p q)
-                      (prop-equal? q -tt)
-                      (prop-equal? p -ff)) #t]
+    [(_ _) #:when (or (equal? p q)
+                      (TrueProp? q)
```
samth: We should change the order of these tests, since the first one is now expensive (and also add `(eq? p q)` at the beginning).

typed-racket-lib/typed-racket/types/prop-ops.rkt
```diff
@@ -70,69 +71,62 @@
...
     [(_ _) #f]))
 
+(define (implies? p q)
+  (FalseProp? (-and p (negate-prop q))))
+
+(define (prop-equiv? p q)
+  (and (implies? p q)
+       (implies? q p)))
+
 ;; intersect-update
```
samth: Why are we using a mutable alist and not a hash table?

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -384,397 +396,560 @@
 ;; Instances, and Structs (Prefabs?)
 (define/cond-contract (subtype* A t1 t2)
   (-> (listof (cons/c Type? Type?)) Type? Type? (or/c #f list?))
-  (early-return
```
samth: What do you have against `early-return`?

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -384,397 +396,560 @@
...
+   (match t2
+     [(? ChannelTop?) A]
+     [(Channel: elem2) (type-equiv? A elem1 elem2)]
+     [(Evt: evt-t) (subtype* A elem1 evt-t)]
+     [_ (continue)])]
+  [(case: Class (Class: row inits fields methods augments init-rest))
+   (match t2
+     [(? ClassTop?) A]
+     [(Class: row* inits* fields* methods* augments* init-rest*)
+      ;; TODO: should the result be folded instead?
```
samth: What does this comment mean?

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -384,397 +396,560 @@
...
+      (define (sub t1 t2) (subtype* A t1 t2))
+      ;; check that each of inits, fields, methods, etc. are
+      ;; equal by sorting and checking type equality
+      (define (equal-clause? clause clause* [inits? #f])
```
samth: This definition can be moved further out, I think.

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -374,6 +372,20 @@
...
 
+(define union-super-cache (make-weak-hasheq))
+(define union-sub-cache (make-weak-hasheq))
+
+(define/cond-contract (cache-set! cache t1 t2 result)
+  (-> hash? Type? Type? boolean? void?)
+  (hash-set! (hash-ref cache t1 (λ () (make-weak-hash))) t2 (box-immutable result)))
+
+(define/cond-contract (cache-ref cache t1 t2)
+  (-> hash? Type? Type? (or/c #f (box/c boolean?)))
```
samth: Shouldn't the result type be `(box/c any/c)`?

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -384,397 +396,560 @@
...
+     [(? VectorTop?) A]
+     [(Vector: elem2) (type-equiv? A elem1 elem2)]
+     [(Sequence: (list seq-t)) (subtype* A elem1 seq-t)]
+     [_ (continue)])]
+  [(case: Weak-Box (Weak-Box: elem1))
+   (match t2
+     [(? Weak-BoxTop?) A]
+     [(Weak-Box: elem2) (type-equiv? A elem1 elem2)]
+     [_ (continue)])]
+  [else: (continue)])
```
samth: Why this last clause?

typed-racket-lib/typed-racket/types/union.rkt
```diff
@@ -1,57 +1,52 @@
...
 
-(provide/cond-contract
- [Un (() #:rest (c:listof Type?) . c:->* . Type?)])
+;; t1 ∪ t2
+;; But excludes duplicate info w.r.t. subtyping
+;; can be useful in a few places, but avoid using
+;; in hot code when Un (or similar) will suffice.
+(define (union t1 t2)
+  (cond
+    [(subtype t1 t2) t2]
```
samth: Can we count how often these two clauses happen?

typed-racket-lib/typed-racket/types/union.rkt
```diff
@@ -1,57 +1,52 @@
...
-            (hash-set! Un-cache args type)
-            type])]))
+;; Recursively reduce unions so that they do not contain
+;; reduntant information w.r.t. subtyping. We used to maintain
+;; this properly throughout typechecking, but this was costly.
+;; Not we only do it as we are generating contracts, since we
+;; don't want to do redundant runtime checks, etc.
+(define (normalize-type t)
+  (match t
+    [(Union: ts) (make-Union (list->hset (foldr merge '() (append-map flat (hset->list ts)))))]
```
samth: This allocates so many intermediate lists.

typed-racket-lib/typed-racket/utils/struct-interface.rkt
```diff
@@ -0,0 +1,74 @@
...
+                      vec))
+             vec)
+           (define-values (prop-name pred get-struct-interface)
+             (make-struct-type-property 'name))
+           (define (accessor struct)
+             (unsafe-vector-ref (get-struct-interface struct) idx))
+           ...)))]))
+
+
+#| delete this line to uncomment example w/ tests
```
samth: Put the tests in a `module+ test`

typed-racket-lib/typed-racket/utils/utils.rkt
```diff
@@ -272,3 +274,20 @@ at least theoretically.
   (syntax-parser
     [(_ x:id val:expr)
      #'(app (λ (_) val) x)]))
+
+(define-syntax (assert stx)
+    (syntax-case stx ()
+      [(_ expr)
+       #`(unless expr #,(quasisyntax/loc stx (error 'assert "failed!")))]))
+
+(define-syntax (def/match stx)
```
samth: Isn't this `define/match`?

typed-racket-test/fail/issue-169-1.rkt
```diff
@@ -1,5 +1,5 @@
 #;
-(exn-pred #rx"expected: \\(Listof Nothing\\).*given: \\(Listof Pos\\*\\)")
+(exn-pred #rx"expected: \\Null.*given: \\(Listof Pos\\*\\)")
```
samth: Why did this error message change?

typed-racket-test/unit-tests/contract-tests.rkt
```diff
@@ -128,262 +128,267 @@
...
-              ;; Adapted from PR 13815
-              (t (-poly (a) (-> a a)))
-              (t (-poly (a) (-mu X (-> a X))))
-              (t (-poly (a) (-poly (b) (-> a a))))
-              (t (-poly (a) (-App (-poly (b) (-> a a)) (list -Number) #'#f)))
+  (test-suite
+   "Contract Tests"
+   (t (-> (-poly (A B) (-> (Un A (-mu X (Un A (-lst X)))) (Un A (-mu X (Un A (-lst X))))))
+          (-> -Symbol (-mu X (Un -Symbol (-lst X)))))))
+  #;(test-suite
```
samth: Why are these commented out?

typed-racket-test/unit-tests/subtype-tests.rkt
```diff
@@ -369,24 +552,28 @@
...
-   [FAIL
-    (-prefab '(foo #()) -String) (-prefab '(foo #(0)) (-opt -String))]
+(define values-tests
+  (test-suite
+   "SomeValues subtyping"
+   (test-check "(Values Number) (Values Univ)"
+               subval
+               (-values (list -Number))
+               (-values (list Univ)))))
+;[FAIL (make-ValuesDots (list) -Symbol 'a) (make-ValuesDots (list (-result -String)) -String 'a)]
```
samth: Why is this commented out?

typed-racket-lib/typed-racket/utils/utils.rkt
pnwamk: not exactly but I don't think its used -- I'll remove it

typed-racket-lib/typed-racket/rep/object-rep.rkt
```diff
@@ -6,37 +6,37 @@
...
+(def-pathelem SyntaxPE () [#:singleton -syntax-e])
+(def-pathelem ForcePE () [#:singleton -force])
+(def-pathelem FieldPE () [#:singleton -field])
 
-(def-pathelem CarPE () #:base)
-(def-pathelem CdrPE () #:base)
-(def-pathelem SyntaxPE () #:base)
-(def-pathelem ForcePE () #:base)
 ;; t is always a Name (can't put that into the contract b/c of circularity)
 (def-pathelem StructPE ([t Type?] [idx natural-number/c])
```
pnwamk: The comment is actually just wrong if I remember correctly (during a refactoring I pushed Name up a level and Names weren't the only thing to enter that field).

typed-racket-lib/typed-racket/rep/rep-utils.rkt
```diff
@@ -24,159 +25,245 @@
...
+    [(identifier? x)
+     (and (identifier? y) (free-identifier=? x y))]
+    [else (equal? x y)]))
 
-;; structural type info for simple/straightforward types
-;; (i.e. we store the list of field variances)
-(define-values (prop:structural structural? Type-variances)
-  (make-struct-type-property 'structural))
+(define id-table (make-free-id-table))
+(define (normalize-id id)
```
pnwamk: Yes - it's so equal? works on ids. Yes, it means we keep around free identifiers forever. 

1. There shouldn't be that many ids.
2. Its a lot faster to use default equal? and hash functions:
https://gist.github.com/pnwamk/7546a47621647da881c20a3ea183bfdc

typed-racket-lib/typed-racket/rep/rep-utils.rkt
pnwamk: switches are building vectors that store the appropriate λ's in their corresponding field. The field index is bound to uid-id, so having it in scope makes it easy for the macro to simply expand into a reference to that id.

typed-racket-lib/typed-racket/types/abbrev.rkt
pnwamk: `init-envs` produces syntax which refers to `register-signature!`, which is defined in `signature-env.rkt`. Without this I get those "module not available" errors from the references to `register-signature!` since `signature-env.rkt` is not loaded.

typed-racket-lib/typed-racket/types/subtype.rkt
pnwamk: No. I'll add a comment. A box returned means we have a cached answer, yes or no (a `boolean?`). Returning `#f` directly means we don't have a cached answer.

typed-racket-lib/typed-racket/types/subtype.rkt
pnwamk: It means the body uses `andmap` and does not thread `A` through the various recursive calls to `subtype*`, and the person who wrote this way back when was unsure if that was the correct choice.

typed-racket-lib/typed-racket/types/subtype.rkt
pnwamk: The switch macro does not enforce coverage -- there has to be an else clause.

typed-racket-lib/typed-racket/types/union.rkt
pnwamk: It's used for printing and contract generation and (thankfully!) did not impact overall performance when I added it. I could tweak it to be less.... wasteful.

typed-racket-test/fail/issue-169-1.rkt
pnwamk: `(Listof Nothing) ≡ Null` and now TR collapses accessible instances of `Nothing`

typed-racket-lib/typed-racket/types/abbrev.rkt
samth: Aaah, I just realized that `signature-env` is a parameter with an immutable table in it. Can this be fixed to be sensible?

typed-racket-lib/typed-racket/types/subtype.rkt
samth: What does `(continue)` do then?

typed-racket-test/fail/issue-169-1.rkt
samth: Ah, that's probably a little confusing. So

```
(if ...
    null
    (cons (error 'fail) blah))
```
has type `Null`?

typed-racket-lib/typed-racket/types/subtype.rkt
pnwamk: Nothing per se... the code got simpler, there were not intermediate bindings, it felt like overkill when `cond` worked just fine and didn't add extra right creep.

typed-racket-lib/typed-racket/types/abbrev.rkt
pnwamk: I'm not sure what you're suggesting be fixed.

typed-racket-lib/typed-racket/types/subtype.rkt
pnwamk: The control flow here is slightly annoying.

Currently the switch will return #t or #f if the subtyping can be determined, or it calls the continuation `continue` to finish checking the remaining subtyping possibilities (which are based on the rhs argument).

I could return a sentinel value and check for the "we haven't figured out yes or no" case... or we can do this continuation calling.

samth: Ok, now I understand.

typed-racket-test/fail/issue-169-1.rkt
pnwamk: Correct. I would cast my vote for "not confusing", personally.

pnwamk: Right now (6.7) the type of that expression is `(Pairof Nothing (Listof Nothing))`.

In this PR, the type of that expression is simply `Nothing`.

typed-racket-lib/typed-racket/rep/rep-utils.rkt
pnwamk: hello-world.rkt, 1 LOC, 86 unique free identifiers
new-metrics.rkt, 508 LOC,  457 unique free identifiers
math-flonum.rkt,  1326 LOC, 801 unique free identifiers
schml-intern-casts-helper, 2098 LOC, 1147 unique free identifiers

typed-racket-lib/typed-racket/rep/rep-utils.rkt
pnwamk: it's explicit in the definition of subtype again

typed-racket-lib/typed-racket/rep/rep-utils.rkt
pnwamk: yes

typed-racket-lib/typed-racket/rep/rep-utils.rkt
samth: I guess we'll have to deal, but we could consider an approach where we made the table weak.

typed-racket-test/unit-tests/subtype-tests.rkt
pnwamk: accident -- fixed now

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -384,397 +396,560 @@
...
-                  (andmap equal? names names*)
-                  (andmap sub types types*))]
+       [_
+        (or
+         ;; then we try a switch on t1
+         (subtype-switch
+          t1 t2 A
+          ;; if we're still not certain after the switch,
+          ;; check the cases that need to come at the end
+          (λ ()
```
pnwamk: done

typed-racket-lib/typed-racket/rep/rep-utils.rkt
samth: yuck, but ok

typed-racket-lib/typed-racket/rep/object-rep.rkt
```diff
@@ -6,37 +6,37 @@
...
-  [#:fold (f) (make-Path (map f elems) name)]
-  [#:walk (f) (for-each f elems)])
+  [#:fmap (f) (make-Path (map f elems) name)]
+  [#:for-each (f) (for-each f elems)]
+  [#:custom-constructor
+   (cond
+     [(identifier? name)
+      (if (is-var-mutated? name)
+          -empty-obj
+          (make-Path elems (normalize-id name)))]
```
samth: Found it

typed-racket-lib/typed-racket/rep/rep-utils.rkt
```diff
@@ -24,159 +25,245 @@
...
+      (hash-id name)
+      (rec name)))
 
 
-;; Is this a type that can be a 'back-edge' into the type graph?
-;; (i.e. could blindly following this type lead to infinite recursion?)
-(define-values (prop:resolvable resolvable?)
-  (let-values ([(prop predicate _) (make-struct-type-property 'resolvable)])
-    (values prop predicate)))
+;; This table maps types (or really, the sequence number of the type)
```
pnwamk: Fixed

typed-racket-lib/typed-racket/rep/rep-utils.rkt
pnwamk: Done

typed-racket-lib/typed-racket/rep/rep-utils.rkt
pnwamk: Done

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
pnwamk: Okay, I'll have to think more about where else this should also go.

typed-racket-lib/typed-racket/types/union.rkt
pnwamk: I cleaned this up to allocate less.

typed-racket-lib/typed-racket/types/prop-ops.rkt
pnwamk: Done

typed-racket-lib/typed-racket/utils/struct-interface.rkt
pnwamk: This file got deleted (it was old and I had forgotten about it)

pnwamk:
I'm not able to reproduce the current error travis-ci is hitting =\

typed-racket-lib/typed-racket/types/prop-ops.rkt
pnwamk: fixed - now uses hashes and less code

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
pnwamk: I added another check after pass 2. At a glance that seems like it should cover all the Apps we could run into while type checking a file.

samth:
What about at the top level?

pnwamk:
I just tried several examples at the top level w/ both TR provided and user defined App types and they all err'd as expected:
image

rfindler:
​Not to be obtuse and there could easily be something I'm missing, but the
error message in the first example says "Pairof" and the program says
"Pair", which suggests to me that the error message isn't as expected?

pnwamk:
Ah, thanks for pointing that out @rfindler -- I'll have to look at why that's happening.

rfindler:
Looks like the last one is wrong too, so probably it isn't specific to
Pairof/Pair.

pnwamk:
Things I've identified so far:

1. Although the error is clearly not being reported correctly in some cases (e.g. when it is reported in terms of the inner type application type), type application errors are reported regardless of whether or not there is a nested type application error:
image

So it appears we're not missing errors, but just incorrectly reporting them (at least from what I've observed so far).

2. This behavior was not introduced by this PR -- you get the same, incorrect error message in the releases 6.7 and 6.6 (that's all I've tried so far).

I'll make an GH issue regarding this inconsistency.

samth:
Is this perhaps just the bad behavior of printing that we already know about, where it chooses the first matching type name?

pnwamk:
@samth ah, yes I think that's it. I just created two type aliases which were identical except for the name:

```racket
(define-type (Foo A) Integer)
(define-type (Bar A) Integer)
```

And errors for both are reported in terms of `Foo`.

samth:
I think we can fix that by associating syntax objects with types when they're parsed, now that we have removed interning, but let's address that after this PR.

typed-racket-lib/typed-racket/rep/rep-utils.rkt
samth: This was wrong, the identifiers will be live for all of type checking anyway.

typed-racket-lib/typed-racket/rep/prop-rep.rkt
```diff
@@ -68,19 +68,9 @@
...
-     [(list) -ff]
-     [(list p) p]
-     [ps (make-OrProp ps)])])
+  [#:for-each (f) (for-each f ps)])
 
 (def-prop AndProp ([ps (listof (or/c OrProp? TypeProp? NotTypeProp?))])
   [#:frees (f) (combine-frees (map f ps))]
   [#:fmap (f) (apply -and (map f ps))]
-  [#:for-each (f) (for-each f ps)]
-  [#:custom-constructor
```
samth: Why aren't these able to be built-in to the constructor?

typed-racket-lib/typed-racket/types/prop-ops.rkt
```diff
@@ -166,30 +139,32 @@
...
+    (for*/list ([(o t) (in-hash ts+)]
+                [p (in-value (-is-type o t))]
+                #:when (not (FalseProp? p)))
+      p))
+  (define neg-props
+    (for*/list ([(o t) (in-hash ts-)]
+                [p (in-value (-not-type o t))]
+                #:when (not (FalseProp? p)))
+      p))
+  (if (or (ormap TrueProp? pos-props)
```
samth: Actually, can't all these loops be combined (including the `for` above)?

typed-racket-lib/typed-racket/infer/infer-unit.rkt
```diff
@@ -790,37 +791,38 @@
...
-                     (let ([entry (hash-ref subst v #f)])
-                       ;; Make sure we got a subst entry for a type var
-                       ;; (i.e. just a type to substitute)
-                       ;; If we don't have one, there are no constraints on this variable
-                       (if (and entry (t-subst? entry))
-                           subst
-                           (hash-set subst v (t-subst Univ)))))])
-       ;; verify that we got all the important variables
-       (extend-idxs subst))]))
+  (for/list ([m (in-list (cset-maps C))])
```
samth: Can we thread `multiple-substitutions?` to here and avoid doing all this work when it's not needed?

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-list.rkt
```diff
@@ -78,21 +78,29 @@
         ;; TODO fix double typechecking
         [_ (tc/app-regular #'form expected)])))
   ;; special case for `list'
-  (pattern (list . args)
-    (let ()
+  (pattern
```
samth: I think this code could have some much simpler special cases based on the expected type (for example, if it's `(List A B C)` just use those as the expected type for each list element, or if it's `(U (List A B C) (List D E F))` then try inference with each union element.

typed-racket-lib/typed-racket/types/prop-ops.rkt
pnwamk: I don't think these can be reasonably combined. I did, however, swap the for/lists to be for/folds (no reverse call) and put a check at the beginning to see if the input prop list is even long enough to warrant all of this potential simplification (Or's really don't get that large in our system, we get lots of small ones). I think that's a reasonable tidying up of this function (I'll push those changes a bit).

typed-racket-lib/typed-racket/infer/infer-unit.rkt
pnwamk: done

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-list.rkt
pnwamk: I implemented the `(List ...)` simplification, but the union case I think is best to stick to what we're currently doing. It's not the case that we could simply try some inferences and not others, we need to get all possible substitutions and try them out as we do now.

pnwamk:
I addressed the last batch of comments.

The only thing now is that `special-env-typecheck-tests.rkt` is broken in a funny way. I made `tc-result-equal/test?` use subtyping/logical equivalence as it should now, but there is some weird phase issue w/ how `special-env-typecheck-tests.rkt` is set up. What happens is it calls `check-below`, and `check-below` doesn't recognize the `tc-results` structs that it gets from `special-env-typecheck-tests.rkt`. In other words, if all our structs were prefabs this wouldn't be an issue, but they aren't, and I haven't figured out how to pipe the info to `check-below yet`.

pnwamk:
Okay -- fixed up the testing issue. Things should be pretty good to go (assuming these last few tests running as I write this pass).

pnwamk:
On the defs in `prop-ops` -- I ran into a situation where my "optimized" versions were running significantly slower on a particular benchmark, and about the same on all the other benchmarks. In particular, different versions of the function `compact` that seem like they should definitely be more efficient (using hasheq's instead of hash, using less dynamic dispatching, etc) were 2 or 3x slower than the previous definition (i.e. the one that is in the latest push). I don't have time to dig into why at the moment (finals!!!) but my initial suspicion is that this more "naive" version is getting more inlining (or other compiler love) than the other version, and since it is a hot piece of code for some programs, it makes a huge difference in those cases.

typed-racket-lib/typed-racket/rep/rep-utils.rkt
```diff
@@ -50,14 +49,22 @@
...
-  (if (identifier? name)
-      (hash-id name)
-      (rec name)))
 
+;; function for fetching the interned copies of Reps
+;; should be robust to GC occurring at any time (since we
+;; not only check for the interned weak-box, but also then
+;; check that it contains a non-#f value)
+(define (intern-table-ref! table val)
+  (define intern-box (hash-ref table val #f))
```
samth: This code isn't thread-safe, which isn't a problem currently but could be in the future.

typed-racket-lib/typed-racket/typecheck/check-below.rkt
```diff
@@ -88,8 +88,8 @@
       [(p p) #t]
       [(p #f) #t]
       [((PropSet: p1+ p1-) (PropSet: p2+ p2-))
-       (and (implies? p1+ p2+)
-            (implies? p1- p2-))]
+       (and (implies-atomic? p1+ p2+)
```
samth: Why is this change safe?

typed-racket-lib/typed-racket/types/prop-ops.rkt
```diff
@@ -66,208 +63,181 @@
...
   (match* (p1 p2)
-    [((TypeProp: o t1) (TypeProp: o t2))
-     (not (overlap? t1 t2))]
-    [((TypeProp: o t1) (NotTypeProp: o t2))
-     (subtype t1 t2)]
-    [((NotTypeProp: o t2) (TypeProp: o t1))
-     (subtype t1 t2)]
-    [(_ _) (or (FalseProp? p1)
-               (FalseProp? p2))]))
+    [((TypeProp: o1 t1) (NotTypeProp: o2 t2))
```
samth: These changes make this function more conservative, right?

typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -73,6 +73,9 @@
 
 
   (define (check-tc-results result golden #:name name)
+    (define (prop-equiv? p q)
```
samth: Is this code unreferenced?

samth:
Is the big algorithmic change in 343d1e0 just reverting changes from earlier?

pnwamk:
That was the intent, yes (I'll double check after Thursday morning for
sure).

samth:
What's the status here?

typed-racket-test/unit-tests/typecheck-tests.rkt
pnwamk: no its used twice in that scope

typed-racket-lib/typed-racket/typecheck/check-below.rkt
pnwamk: it was reverting a change back to the way its always been for now

pnwamk:
Okay - making progress. Should have it on the next try ;-)

pnwamk:
Just a side note: This PR was in a working state at commit aff78d -- the only issue was a pretty significant regression in a new benchmark I found (Sarah's parser from a month ago or so).

So, at the moment I'm trying to see what the smallest change to aff78d is that fixes that regression w/o introducing any new issues.

Sorry for the noise.

samth:
Great! Looking forward to this landing.

pnwamk:
NOTE: I ended up scratching a previous approach and just building off the last successfully building commit (aff78d) after I found the heart of the problem - sorry for any confusion.

Heart of the previous perf problem (the only real known issue with aff78d ): Basically '-and' and '-or' for the longest time were (a) written to fail fast, and (b) were were written in a way where the ORDER the props were examined in could have a HUGE impact on the size of the resulting prop (that is, once we started checking if ors were subsets of one another in `implies-atomic?`. 

Solution: I reverted back to the original approaches so we get (a) back, and for (b) I tried to address the issue in both: 

1. `-or` now is written so that order is more or less irrelevant

2. `-and` does a more thorough job ordering its props now, taking the size of disjunctions into account so we consistently eliminate redundant ors.

Additionally, I added interning for `Path`, `TypeProp`, `NotTypeProp`, and `OrProp` since they contain _lots_ of copies of the same info (i.e. they're really the thing that can grow exponentially when type checking, but they don't actually contain an exponential number of atomic parts) and now we can do `eq?` and `memq` checks in our hot logical code (see `implies-atomic?`).

And shoot me if the tests don't pass. I think they're all fine on my machine.

rfindler:
On Fri, Dec 16, 2016 at 9:11 AM, Andrew Kent <notifications@github.com>
wrote:

> And shoot me if the tests don't pass.
>

Can I buy you ice-cream instead?

pnwamk:
yes... ice-cream is a fine alternative I suppose =)

pnwamk:
# Performance Summary
Things are generally slightly faster or about the same, memory usage is down or about the same in all cases.

# Benchmark Programs
https://github.com/pnwamk/tr-performance

# Time Plot
image

# Max Memory Usage Plot
image

# Raw Data
PR Performance
/usr/bin/time -l raco make -v new-metrics.rkt
"new-metrics.rkt":
  making #<path:/Users/pnwamk/Repos/tr-performance/new-metrics.rkt>
 [output to "compiled/new-metrics_rkt.zo"]
        3.37 real         2.85 user         0.47 sys
 214298624  maximum resident set size

/usr/bin/time -l raco make -v math-flonum.rkt
"math-flonum.rkt":
  making #<path:/Users/pnwamk/Repos/tr-performance/math-flonum.rkt>
 [output to "compiled/math-flonum_rkt.zo"]
        4.16 real         3.67 user         0.47 sys
 208060416  maximum resident set size

/usr/bin/time -l raco make -v slowcheck.rkt 
"slowcheck.rkt":
  making #<path:/Users/pnwamk/Repos/tr-performance/slowcheck.rkt>
 [output to "compiled/slowcheck_rkt.zo"]
        2.53 real         2.12 user         0.39 sys
 189218816  maximum resident set size

/usr/bin/time -l raco make -v parser.rkt
"slow-tree.rkt":
  making #<path:/Users/pnwamk/Repos/tr-performance/slow-tree.rkt>
 [output to "compiled/slow-tree_rkt.zo"]
        3.67 real         3.26 user         0.37 sys
 177647616  maximum resident set size

/usr/bin/time -l raco make -v schml-interp-casts-help.rkt 
"schml-interp-casts-help.rkt":
  making #<path:/Users/pnwamk/Repos/tr-performance/schml-interp-casts-help.rkt>
 [output to "compiled/schml-interp-casts-help_rkt.zo"]
       23.21 real        22.19 user         0.98 sys
 396488704  maximum resident set size



HEAD Performance

/usr/bin/time -l raco make -v new-metrics.rkt
"new-metrics.rkt":
  making #<path:/Users/pnwamk/Repos/tr-performance/new-metrics.rkt>
 [output to "compiled/new-metrics_rkt.zo"]
        3.69 real         3.15 user         0.50 sys
 231137280  maximum resident set size

/usr/bin/time -l raco make -v math-flonum.rkt
"math-flonum.rkt":
  making #<path:/Users/pnwamk/Repos/tr-performance/math-flonum.rkt>
 [output to "compiled/math-flonum_rkt.zo"]
        3.98 real         3.47 user         0.48 sys
 215490560  maximum resident set size

/usr/bin/time -l raco make -v slowcheck.rkt
"slowcheck.rkt":
  making #<path:/Users/pnwamk/Repos/tr-performance/slowcheck.rkt>
 [output to "compiled/slowcheck_rkt.zo"]
        2.51 real         2.09 user         0.39 sys
 193814528  maximum resident set size

/usr/bin/time -l raco make -v parser.rkt 
"slow-tree.rkt":
  making #<path:/Users/pnwamk/Repos/tr-performance/slow-tree.rkt>
 [output to "compiled/slow-tree_rkt.zo"]
       14.60 real        14.10 user         0.45 sys
 197877760  maximum resident set size
 
/usr/bin/time -l raco make -v schml-interp-casts-help.rkt 
"schml-interp-casts-help.rkt":
  making #<path:/Users/pnwamk/Repos/tr-performance/schml-interp-casts-help.rkt>
 [output to "compiled/schml-interp-casts-help_rkt.zo"]
       23.00 real        21.66 user         1.30 sys
 600612864  maximum resident set size

pnwamk:
cleaned up prop-ops a little -- performance is the same and the code is clearer now. I think its ready =D

samth:
Great!

pnwamk:
I remembered to check out build w/ full contract checking on -- found a bug that a30d798
fixes (it wasn't returning "full-tc-results").

pnwamk:
merged (commit 8e7f390).

samth:
Woo!

PR #471, opened 2016-12-14
sync with change to `for` expansion (goes with racket/racket PR)

mflatt:
In racket/racket#1534, the expansion of `for` for a generic sequence has changed to receive
and use an extra result from the internal `make-sequence` function. These appear to be the needed changes for TR.

The public interface of `make-do-sequence` has also changed to allow a function that returns one more result. I don't think that anyone will ever want to provide that extra result, so I left the type as-is.

### DIFF
--- a/typed-racket-lib/info.rkt
+++ b/typed-racket-lib/info.rkt
@@ -2,7 +2,7 @@
 
 (define collection 'multi)
 
-(define deps '(("base" #:version "6.4.0.5")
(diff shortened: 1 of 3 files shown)


samth:
This looks good, but should probably feature a change to info.rkt to depend on the appropriate base version (which should be incremented).

mflatt:
Updated dependency - thanks!

samth:
Looks good to merge, then.

PR #473, opened 2016-12-17
fix intermittent error in 8e7f390

pnwamk:
This PR (1) removes some unnecessary nondeterminism from prop-ops and (2) fixes the issue where the `nested-poly` integration test would occasionally fail. Basically, the correctness of type checking currently relies on the `var-name-table` in `type-rep.rkt` being *persistent*... which seems odd and wrong, but that's definitely the issue. I removed the table entirely and that caused the error to consistently happen, and it seems like that program w/ its nested polymorphic applications is the only program I've run into that relied on that particular table being persistent.

A better fix I think would be to further examine how the table is used and remove this requirement --- my initial assumption was that the table was just for error reporting and printing (so that users saw types w/ the same polymorphic variables that appeared in the program) but clearly its doing a little more than that.

### DIFF
--- a/typed-racket-lib/typed-racket/rep/type-rep.rkt
+++ b/typed-racket-lib/typed-racket/rep/type-rep.rkt
@@ -72,7 +72,7 @@
  ("../types/overlap.rkt" (overlap?))
  ("../types/resolve.rkt" (resolve-app)))
 
-(define var-name-table (make-weak-hash))
+(define var-name-table (make-hash))
 
 ;; Name = Symbol
 
--- a/typed-racket-lib/typed-racket/types/prop-ops.rkt
+++ b/typed-racket-lib/typed-racket/types/prop-ops.rkt
@@ -5,8 +5,6 @@
          (prefix-in c: (contract-req))
          (rep type-rep prop-rep object-rep values-rep rep-utils)
          (only-in (infer infer) intersect)
-         compatibility/mlist
-         racket/set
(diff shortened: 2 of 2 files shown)


samth:
Is it just is depending on the ordering of that table?

pnwamk:
No - sorry that was not a good description I gave.

So the behavior is: when the name table is a `weak-hash` the error will occur intermittently. If, however, it is a normal `mutable-hash` table the error never occurs. And if the table is gone entirely then error _consistently_ happens.

I believe this is because with the `weak-hash` occasionally a GC will remove  a hash entry from the name table that is still needed for type checking the program in question.  i.e. GC collects the entry `[τ ↦ sym]`  because there is no reachable `τ`, but there is a reachable `τ'` that is `equal?` to `τ` and that is still relying on the `[τ ↦ sym]` entry being present in the name table.

samth:
OK, that makes sense.

For the nondeterminism, it's good to fix this now, but if you try other orders for those lists, can you trigger the failure consistently? If just reversing it doesn't work, try setting the random seed and then shuffling.

pnwamk:
Again sorry for not being clear. The pull request that changes `make-weak-hash` to `make-hash` fixes the bug we saw. 

The other changes (in prop-ops.rkt) are proactive so there is not nondeterminism in the `prop-ops` code that _could be_ the cause of issues in the future (especially since the deterministic version performs just as well!).

PR #474, opened 2016-12-19
be less eager about bottom propagation

pnwamk:
Moving to eager propagating of bottom works for most cases,
but in some cases flattening types such as (Pairof Bottom Any)
to Bottom made things like type inference break for some cases
(since (Listof Nothing) == Null, and (Listof A) did not structurally
like up like it used to). Perhaps w/ a little more effort
inference and any other potential issues could work better
with propagating bottom, but for now we'll be slightly less
aggressive about it.

i.e. this fixes `pfds`, which commit 8e7f390 broke.


pnwamk:
fixed and merging

samth:
Sounds good.

Issue #429, opened 2016-09-15
Known type information should be available in the propositions (pos/neg filters).


jsmaniac:
### What version of Racket are you using?

6.4.0.5 (20160914-62f5b2c)
### What program did you run?

The following code builds a `(stx-car/c nested-c)` contract, which returns the result of calling `nested-c` (usually a boolean) on the `stx-car` of `v`. I wanted to add some type propositions to make it clear that the result is non-false iff `v` is a `(Syntaxof (Pairof Pos Any))`, where `Pos` is the positive proposition for `nested-c`. The goal is to tell TR that `((stx-car/c number?) v)` is true iff `v` is `(Syntaxof (Pairof Number Any))`.

```
#lang typed/racket

(: stx-car/c (∀ (Result Pos Neg) (→ (→ Any Result : #:+ Pos #:- Neg)
                                    (→ Any (U #f Result)
                                       : #:+ (Syntaxof (Pairof Pos Any))))))
(define ((stx-car/c nested-c) v)
  (if (syntax? v)
      (if (pair? (syntax-e v))
          (let ([result (nested-c (car (syntax-e v)))])
            (if result
                (begin (ann v (Syntaxof (Pairof Pos Any)))
                       result)
                #f))
          #f)
      #f))
```
### What should have happened?

The program should typecheck. The information concerning `v` is known, since the expression `(ann v (Syntaxof (Pairof Pos Any)))` typechecks witout any problem, but that information is not propagated to the `#:+` filter.
### If you got an error message, please include it here.

The error message shows that TR has all the information needed in small pieces, but it does not combine them. Also, since the path-elem is restricted to `car` and `cdr`, one can only write `#:+ (Type @ (car-or-cdr …) v)`, but not e.g. `#:+ (Type @ (car syntax) v)`, so it's impossible to simply tell TR to use the filters it has inferred, because they can't be expressed syntactically.

```
Type Checker: type mismatch;
 mismatch in proposition
  expected: (((Syntaxof (Pairof Pos Any)) @ v) | Top)
[...]

PR #475, opened 2016-12-19
WIP: propagate expected props further down AST for typechecking

pnwamk:
Previously expected propositions were erased frequently and without
any comments explaining why.  Instead of checking propositions at the
leaves of the AST, we would erase expected props at let's and ifs and
typecheck the remaining portion of the AST while propagating up ALL
logical info we learned. Then, once all the logical info from the
leaves of the AST had amassed, we would check if it entailed the
logical expected props that we erased. The problem is that you can
easily get exponential blow-up in the size of propositions using this
strategy, even when we are not interested in the propositions whatsoever.

This PR instead sends expected types _and_ props all the way down when
typechecking, so we can verify any expected logical conclusions at
leaves (and with _much_ smaller propositions to worry
about). Additionally, if an expected prop is present, we simply verify
it is satisfied and _do not_ propagate up any logical info (this is
what gets rid of the old exponential growth).

The added test `slow-parser.rkt` is an example of a program that used
to have typecheck time which grew exponentially in the number of match
clauses, but now checks remarkably faster with what appears to be
linear (or similar) scaling w.r.t. the number of clauses.

NOTE: I've labeled this a WIP since it could easily break obscure
things if I haven't handled everything correctly... so I need to do
some thorough testing still.

### DIFF
--- a/typed-racket-lib/typed-racket/infer/infer-unit.rkt
+++ b/typed-racket-lib/typed-racket/infer/infer-unit.rkt
@@ -220,12 +220,12 @@
       (substitute (make-F var) v ty*))))
 
 
-(define/cond-contract (cgen/prop context s t)
+(define/cond-contract (cgen/prop context p q)
   (context? Prop? Prop? . -> . (or/c #f cset?))
-  (match* (s t)
-    [(e e) (empty-cset/context context)]
-    [(e (TrueProp:)) (empty-cset/context context)]
-    [((FalseProp:) e) (empty-cset/context context)]
+  (match* (p q)
+    [(p p) (empty-cset/context context)]
+    [(p (TrueProp:)) (empty-cset/context context)]
+    [((FalseProp:) q) (empty-cset/context context)]
     ;; FIXME - is there something to be said about the logical ones?
     [((TypeProp: o s) (TypeProp: o t)) (cgen/inv context s t)]
     [((NotTypeProp: o s) (NotTypeProp: o t)) (cgen/inv context s t)]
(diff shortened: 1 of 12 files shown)


samth:
How do other performance tests fare?

Also, can we test this on as many TR-using pkgs as possible, to avoid the problem with `pfds` that we just saw?

pnwamk:
Performance: I have observed largely no change in any of the benchmarks except the `parser.rkt` example, which, as I stated, had a HUGE improvement. I think this is probably why the previous approach seemed okay: for many programs the "potential blow up" just doesn't register. Also -- I have not measured how this affects any larger scale benchmarks (math library, pict3d, plot, etc).

Also -- maybe there is more trimming that could be done still -- this is definitely just my first try at this poking around trying to keep the size of props down.

Other tests: I'll definitely run a few other libraries through the ringer (pfds, plot, pict3d). We don't have an "extended" test script that does this for us yet, do we? That might be nice to add....

pnwamk:
Oh - one other note. This PR _greatly_ improves the completeness of `check-below` on logical information -- it uses the available info in the environment as well as the expressions logical propositions. Because of this, there may be a nice set of programs w/ non-trivial props that used to have a difficult typechecking but now trivially do (e.g. issue #429).

pnwamk:
Tests done in addition to standard typed-racket-test batch:
- [ ] pfds
- [ ] pict3d
- [ ] plot

pnwamk:
merged via be155fa3e3236caa93ef927e205c45b00d6d6e6e

PR #476, opened 2016-12-22
Add typed/images/compile-time.rkt

wargrey:
copied from `images/compile-time` with few modifications.

### DIFF
--- a/typed-racket-test/succeed/pr476-compile-time-images.rkt
+++ b/typed-racket-test/succeed/pr476-compile-time-images.rkt
@@ -1,0 +1,8 @@
+#lang typed/racket
+
+(require typed/images/compile-time)
+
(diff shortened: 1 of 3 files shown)


wargrey:
So just put a tests into typed-racket-more/succeed?

samth:
Yes, exactly.

wargrey:
Okay, it's done.

typed-racket-test/succeed/pr476-compile-time-images.rkt
```diff
@@ -0,0 +1,8 @@
+#lang typed/racket
+
+(require "../../typed-racket-more/typed/images/compile-time.rkt")
```
samth: I think this should just be `(require typed/images/compile-time)`.

PR #478, opened 2016-12-29
Use existing contract combinators.

stamourv:
To cooperate better with the contract system.

This introduces a dependency to a module that (it looks like) we want to keep thin. I'd imagine that anything that requires that module would also depend on the contract system anyway, so the point is moot?

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -865,11 +865,11 @@
     [_ #f]))
 
 (module predicates racket/base
-  (require racket/extflonum)
+  (require racket/extflonum (only-in racket/contract/base >=/c <=/c))
   (provide nonnegative? nonpositive?
            extflonum? extflzero? extflnonnegative? extflnonpositive?)
-  (define nonnegative? (lambda (x) (>= x 0)))
-  (define nonpositive? (lambda (x) (<= x 0)))
+  (define nonnegative? (>=/c 0))
+  (define nonpositive? (<=/c 0))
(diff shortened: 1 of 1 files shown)


stamourv:
Good point. That would help some.

stamourv:
I'd forgotten to push the change.
@samth: Thoughts?

samth:
Where else is the predicates sub module used?

stamourv:
It's required in all residual TR programs, AFAICT. Hence the precaution.

samth:
In that case, I think we should find a different strategy for doing this.

stamourv:
Other modules that are also required by all residual TR programs already do require `racket/contract` (e.g., `utils/evt-contract.rkt`, which could perhaps just require `racket/contract/base`). So I think having this module also require `racket/contract/base` wouldn't incur much overhead, but I'm not confident in my grasp of the module loading machinery.

samth:
I think those other modules are only required by the contract-defs submodule. Is that the case for primitives as well?

stamourv:
The only other use of the `predicates` (that `extra-requires` other things in `extra-requires` may not have) is in the top-level module of type-contract.rkt, so not in residual code.

samth:
If an empty TR file won't depend on it, then I'm fine with the change.

stamourv:
The `require` to  the`predicates` submodule does show up in the expansion of an empty TR file as part of the `#%contract-defs` submodule, but it doesn't look like this submodule is required as part of executing an empty TR module.

samth:
Yes, that's what I mean. Sounds good.

PR #480, opened 2017-01-01
Add type annotation for (vector-set-performance-stats)

wargrey:


typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -3026,7 +3026,12 @@
 
 [current-command-line-arguments (-Param (-vec -String) (-vec -String))]
 [current-thread-initial-stack-size (-Param -PosInt -PosInt)]
-;vector-set-performance-stats! TODO complicated
+[vector-set-performance-stats!
+ (let ([place-stats (list -Fixnum -Fixnum -Fixnum -Nat -Nat -Nat -Nat -Nat -Fixnum -Fixnum -Nat -Nat)]
```
samth: I think using `-Integer` for everything is probably sufficient.

wargrey: if so, is `(-vec -Integer)` okay? Clients will not be limited to pass a fixed length vector.

wargrey: Or better `(Un (-vec -Int) (make-HeterogeneousVector (list -Int ...))` since `(-vec a)` does not accept `(vector a ...)`.

PR #485, opened 2017-01-06
Make keyword conversion actually expand types to all cases

pnwamk:
## Related PR

This PR is more or less an updated version of https://github.com/racket/typed-racket/pull/62 (thank you @endobson for initially proposing a correct fix for this that ended up only needing a slight tweak!)

## Problem Description

Keyword functions are a little tricky. This PR (and the preceding one) were addressing issues checking the _body_ of kw functions.

Basically, a function with keyword arguments such as `inc`:

```racket
(define (inc x #:n [n 1])
   (+ x n))
```

actually expands into a more complex function with 3 arguments that looks something resembling the following:

```racket
(define (inc-expanded n* n-given? x)
   (let ([n (if n-given? n* 1)]) (+ x n)))
```

and calls to `inc` are converted to match this form:

```racket
(inc 42) => (inc-expanded #f #f 42)

(inc 42 #:n 2) => (inc-expanded 2 #t 42)

```

Note that each optional keyword argument has a boolean flag argument that signals whether or not the caller provided that keyword argument.

Without diving into all of the reasons as to why, basically our handling that typechecked the body of functions such as `inc` has been incorrect for quite some time. An example program that should not typecheck (but has for some time) is the following:

```racket
(: f (Number [#:y Boolean] -> Number))
(define (f x #:y [y #f] #:z [z 'this-can-be-anything])
   (if y "y is truthy" x))
```

This problem was discovered and a fix was presented in https://github.com/racket/typed-racket/pull/62

This fix, however, basically expanded the function type into _every possible combination_ of function types w.r.t. the optional arguments and their accompanying boolean flags.

The body for `inc`, for example, would be checked against the following function type:

```racket
(case->
  [-> False False Number Number]
  [-> Number True Number Number])
```

Unfortunately, this sort of _complete_ expansion of the type resulted in astronimically large types for libraries that use a lot of keyword arguments (see `plot`).

As a tweak to this approach, this PR noticed that the value for the `n*` in `inc` is _only_ rachable in code when `n-given?` is `#t`, and so, assuming the kw-expansion protocol always only accesses `n*` if `n-given?` is `#t`, we can actually safely check the body of the function against the following simpler type:

```racket
(-> Number Boolean Number Number)
```

## Performance issue w/ original

Here example of a file that currently takes a ridiculously long time to type check w/ the original changes because expanding into every possible optional argument combination was prohibitively costly for type checking: `plot-lib/plot/private/no-gui/plot3d.rkt` ([link](https://github.com/racket/plot/blob/master/plot-lib/plot/private/no-gui/plot3d.rkt))

### DIFF
--- a/typed-racket-lib/typed-racket/rep/type-rep.rkt
+++ b/typed-racket-lib/typed-racket/rep/type-rep.rkt
@@ -462,11 +462,17 @@
   [#:for-each (f) (f ty)])
 
 
+(define (keyword-sorted/c kws)
+  (or (empty? kws)
+      (= (length kws) 1)
+      (apply keyword<? (map Keyword-kw kws))))
+
+
 (def-rep arr ([dom (listof Type?)]
               [rng SomeValues?]
               [rest (or/c #f Type?)]
               [drest (or/c #f (cons/c Type? (or/c natural-number/c symbol?)))]
-              [kws (listof Keyword?)])
+              [kws (and/c (listof Keyword?) keyword-sorted/c)])
   [#:frees
    [#:vars (f)
     (combine-frees
@@ -506,6 +512,7 @@
    (when drest (f (car drest)))
    (when rest (f rest))
    (for-each f kws)])
+
 
 ;; arities : Listof[arr]
 (def-type Function ([arities (listof arr?)])
--- a/typed-racket-lib/typed-racket/types/kw-types.rkt
+++ b/typed-racket-lib/typed-racket/types/kw-types.rkt
@@ -7,71 +7,63 @@
          "../utils/tc-utils.rkt"
          "../base-env/annotate-classes.rkt"
          "tc-result.rkt"
-         racket/list racket/set racket/dict racket/match
+         racket/list racket/set racket/match
          racket/format racket/string
          syntax/parse)
 
 ;; convert : [Listof Keyword] [Listof Type] [Listof Type] [Option Type]
 ;;           [Option Type] [Option (Pair Type symbol)] boolean -> Type
 (define (convert kw-t plain-t opt-t rng rest drest split?)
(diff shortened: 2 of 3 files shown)


pnwamk:
I took a stab at making this correction not destroy type checking time.

Basically, each optional kw argument comes w/ two real arguments we consider when type checking the body: the `kw-arg` of type `τ` (i.e. the actual kw argument) and the `user-provided-kw-arg?` flag argument of type `Boolean` that signals whether or not the user provided the optional argument.

When type checking the _body_ of these keyword functions, the original PR took into consideration the fact that if the `kw-arg` is not provided, and thus `user-provided-kw-arg?` is `#false`, then the kw protocol for Racket will also pass `#false` for `kw-arg`, and to account for this we generated every possible combination of optional keyword arguments and types in a (potentially huge) case lambda. 

However, we can observe that in reality, the kw protocol could pass _any_ value for the unused `kw-arg`, because the value will _never_ be used. Instead the protocol checks and sees that `user-provided-kw-arg?` is `#false` and completely ignores the value of `kw-arg`.

If we assume that the kw protocol is correct, and that if `user-provided-kw-arg?` is `#false` then the value of `kw-arg` is never used, then for type checking the _body_ we can simply ignore the fact that `kw-arg` is sometimes `False` and assume it is of type `τ`, since the `False` value is "unreachable". So then to typecheck the body we just record that `kw-arg` is of type `τ` and `user-provided-kw-arg?` is of type `Boolean` and we're home free!

Anyway, I haven't done a deep dive on this and there could be some totally incorrect assumptions above and this was a silly attempt to address the issue.

pnwamk:
plot, pfds, and pict3d build successfully w/ these changes

samth:
What's the performance impact for these changes (including on those library builds)?

samth:
Yay!

pnwamk:
merged via 3b80ae71f9d84abb717f92d16723a8f1f6d27347

Issue #489, opened 2017-01-12
`for/flvector` and `for/extflvector` documented but not provided?


AlexKnauth:
The forms `for/flvector`, `for*/flvector`,`for/extflvector`, and `for*/extflvector` are documented as being provided by `typed/racket` and `typed/racket/base`, but they are not provided there.

Should they be removed from the documentation? Or should they be provided along with the `:` versions, `for/flvector:` etc, which are provided?

AlexKnauth:
@samth Why was this issue closed? #490 addressed a separate issue, not this one.

samth:
I'm confused. Why doesn't that address this issue?

AlexKnauth:
Because those four identifiers are still documented but not provided.

AlexKnauth:
Actually, the  typechecker can now handle the expansions of `for/flvector:` etc, so it would make sense to provide the renamed versions without the colons.

However, this will conflict with `(require racket/flonum)`, so if people have both that and `(require typed/racket)` in their program, this would cause an error.

bestlem:
I am using 8.2 or 8.3 on macOS with #lang typed/racket 

And I get the error
for/flvector: unbound identifier in: for/flvector

And for/flvector is documented in 12 Legacy Forms ot typed racket https://docs.racket-lang.org/ts-reference/Legacy_Forms.html

How do I get this to work?

PR #490, opened 2017-01-12
Move unsupported `for*/` forms to the unsupported section

AlexKnauth:


stamourv:
`for/flvector:` (with a colon) and co are defined and exported. `for/flvector` (no colon) and co are not.
The right solution (IMO) would be to export those under both names.

AlexKnauth:
That's about https://github.com/racket/typed-racket/issues/489, which is separate from this pull request.

stamourv:
Ah. I saw the two as being together.
If the typechecker can't deal with the expansion of `for/flvector` and co, then the change you're proposing makes sense.

AlexKnauth:
This isn't about moving `for/flvector`; this is about moving the `for*/` variants so that they're together with their `for/` partners.

stamourv:
Oops. That's what I meant. See my previous statement with `*` added where necessary.

AlexKnauth:
So should this be merged?

stamourv:
If the typechecker can't deal with their expansion, yes.

PR #493, opened 2017-01-14
regression test for issue #33

### LINKED ISSUE #33: The `list` special case uses the expected type poorly
Example:

```
#lang typed/racket

(: x (U (List Symbol Any)
        (List String Symbol)))
(define x (list "foo" 'bold))
```

Produces this error message:

```
/tmp/union.rkt:5:10: Type Checker: type mismatch
  expected: (U (List Symbol Any) (List String Symbol))
  given: (List String Any)
  in: (list "foo" (quote bold))
  context...:
```

Note that using a literal list there for `x` will work.

bennn:
closes #33

### DIFF
--- a/typed-racket-test/succeed/issue-33.rkt
+++ b/typed-racket-test/succeed/issue-33.rkt
@@ -1,0 +1,17 @@
+#lang typed/racket/base
+
+;; regression test for issue #33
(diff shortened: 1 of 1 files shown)


samth:
Looks good.

PR #498, opened 2017-01-26
Should TR optimizer docs recommend real->double-flonum instead of exact->inexact?

AlexKnauth:
Since its usage in the guide is to create a value of type `Float` on which TR can apply optimizations, `real->double-flonum` seems like the the better choice. The type of `exact->inexact` can sometimes produce single flonums or complex numbers, but `real->double-flonum` always returns a `Float`.

For converting exact complex numbers into inexact complex numbers I don't see another option; I'm not sure what to do there.


stamourv:
Good idea. These docs predate the existence of `real->double-flonum` (or even single-flonums being enabled by default) :)
You make a good point re complex numbers, though. There should probably be an equivalent conversion function that always returns a complex number with double parts.

pnwamk:
Was about to merge this -- any reason to hold off, @AlexKnauth?

AlexKnauth:
No.

Issue #415, opened 2016-08-14
(let) requires explicit type annotations for struct accessors


wargrey:
### What version of Racket are you using?

I am not sure what exactly the broken condition is. It also happens to occur in version 6.5 in few (normal) situations. In version 6.6, it occurs more often, almost every newly written `let`s require explicit type annotations for struct accessors.
### What program did you run?

```
(struct selector ([ids : (Listof Keyword)]) #:prefab)
(struct subject ([id : (U Keyword (Pairof Keyword (Listof Keyword)))]) #:prefab)

(define s (selector '(#:id)))
(define e (subject '#:123))

;;; Okay
(let ([ids : (Listof Keyword) (selector-ids s)]
      [id : (U Keyword (Pairof Keyword (Listof Keyword))) (subject-id e)])
  (cond [(null? ids) #true]
        [(keyword? id) (and (null? (cdr ids)) (eq? (car ids) id))]
        [else (set=? (list->set ids) (list->set id))]))

;;; Broken
(let ([ids (selector-ids s)]
      [id (subject-id e)])
  (cond [(null? ids) #true]
        [(keyword? id) (and (null? (cdr ids)) (eq? (car ids) id))]
        [else (set=? (list->set ids) (list->set id))]))

;;; Okay
(define ids (selector-ids s))
(define id (subject-id e))
(cond [(null? ids) #true]
      [(keyword? id) (and (null? (cdr ids)) (eq? (car ids) id))]
      [else (set=? (list->set ids) (list->set id))])
```
### What should have happened?

All three test cases should be evaluated to `false`. 
### If you got an error message, please include it here.

No more error messages.

wargrey:
Same behavior shows up within `in-list`.
But it seems that old code has not been found broken.

andmkent:
Minor update -- the behavior is correct when the structs are not prefab.

wargrey:
Oh, that is informative.
All structs in my current application are prefab structs because I need the serialization mechanism.

wargrey:
@pnwamk

Hello, what's the status of this issue?
This is really an awkward thing, since prefab struct is the most efficient serialization mechanism that out of box in Typed Racket, it can play an important role in in-process communication.

Now the big problem is, `match` loses all the type information about prefab structures, even worse, `assert` deals nothing with it without following a silly `(define v : Any instance)` or `(define v : Any (struct-field instance))`

Thanks in advance.

wargrey:
```racket
#lang typed/racket

[...]

PR #500, opened 2017-02-07
Fix Prefab flds

pnwamk:
Description in terms of Typed Racket usage: Fixes bugs w.r.t. let binding and matching on the fields of prefabs (see, for example, https://github.com/racket/typed-racket/issues/415).

Description in terms of Typed Racket internals: Fixes how prefab field path elements are handed w.r.t. `path-type` and `update`.

### DIFF
--- a/typed-racket-test/succeed/prefab.rkt
+++ b/typed-racket-test/succeed/prefab.rkt
@@ -56,10 +56,10 @@
 
 
 
-(struct msg ([str : String]) #:prefab)
+(struct msg ([str : String] [maybe-str : (Option String)]) #:prefab)
 
 ;; tests let bindings fields of prefab structs
-;; works as expected
+;; works as expected (requires path-type.rkt to be correct w.r.t. prefabs)
 (: test-let-binding (-> Any String))
 (define (test-let-binding x)
   (cond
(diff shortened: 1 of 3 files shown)


typed-racket-test/succeed/prefab.rkt
```diff
@@ -53,3 +53,45 @@
...
+
+
+(struct msg ([str : String] [maybe-str : (Option String)]) #:prefab)
+
+;; tests let bindings fields of prefab structs
+;; works as expected (requires path-type.rkt to be correct w.r.t. prefabs)
+(: test-let-binding (-> Any String))
+(define (test-let-binding x)
+  (cond
+    [(msg? x)
```
samth: This example doesn't seem sound to me. What if we just read an instance of MSG from a string with the wrong kind of fields?

pnwamk: This doesn't sound like an issue with this PR... it sounds like a fundamental issue with `prefab` structs. For example, this code works in v6.8:

```racket
#lang typed/racket/base
 
(struct fubar ([id : Byte]) #:prefab)
 
(define rval : Any (read (open-input-string "#s(fubar NotANumber)")))
 
 
(when (fubar? rval)
  (displayln (+ 1 (fubar-id rval))))
```

And produces nondeterministic, silly output.

pnwamk: @samth thoughts? 

To summarize/recap... currently prefabs have the following properties in TR:

1. There are types which claim to describe them.
2. They can easily expose unsound/unsafe behavior the way they are currently implemented.
3. They are a huge pain to program with (because there are `Path` objects which can refer to their fields, but we are unable to lookup or update the types for those `Path`s (leading to the frustrating behaviors noted in https://github.com/racket/typed-racket/issues/415)).

This PR relieves (3) and does not affect (1) or (2).

samth: If we had sensible types for prefabs (fixing issues 1/2) would the resulting accessors have the paths that this PR adds?

pnwamk: Yes. Immutable fields in sane prefab structs would certainly be things we wanted paths to be able to reference (and I believe we're already currently ommitting mutable fields), and the additions this PR makes (making update/path-type able to reach into prefab types) certainly would make sense in that setting.

My only concern is more programs being created after this PR that would then be broken by fixing prefabs... but I'm not positive this would be an issue, it's just a kneejerk concern at the moment...

pnwamk:
This issue needs revisited and a more wholistic solution (i.e. actually dealing with the underlying prefab issues). Closing this PR (see https://github.com/racket/typed-racket/issues/621 for further work/discussion on the underlying prefab issues mentioned here).

PR #505, opened 2017-03-01
fix types of `exact?` and `inexact?`, closes #502

### LINKED ISSUE #502: `exact?` and `inexact?`'s types mismatch with untyped definitions

Racket `v6.8.0.2`

### What program did you run?

The following programs will result in run-time errors, because `exact?` and `inexact?`'s domain are `Any`, but the untyped implementations expect `number?`. I looked into `base-env.rkt` but didn't figure out where to fix them. I'm guessing they're defined as predicates for `Exact-Number` somewhere...

```{racket}
(exact? #t)
(inexact? '())
```

philnguyen:
The current types for `exact?` and `inexact?` mismatch with the untyped implementation and cause runtime errors as in issue #502.

This commit attepmts to fix the domains and make the types more precise, knowing that the tested value is a number.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env-numeric.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env-numeric.rkt
@@ -728,8 +728,10 @@
 [complex? (make-pred-ty N)]
 ;; `rational?' includes all Reals, except infinities and NaN.
 [rational? (asym-pred Univ B (-PS (-is-type 0 -Real) (-not-type 0 -Rat)))]
-[exact? (make-pred-ty -ExactNumber)]
-[inexact? (make-pred-ty (Un -InexactReal -InexactImaginary -InexactComplex))]
+[exact?   (-> N B : (-PS (-is-type 0 -ExactNumber)
+                         (-is-type 0 (Un -InexactReal -InexactImaginary -InexactComplex))))]
+[inexact? (-> N B : (-PS (-is-type 0 (Un -InexactReal -InexactImaginary -InexactComplex))
+                         (-is-type 0 -ExactNumber)))]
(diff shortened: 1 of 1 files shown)


AlexKnauth:
Does `-InexactComplex` already include `-InexactReal` and `-InexactImaginary`?

philnguyen:
I checked in the repl, looks like it does not. The original types for `inexact?` also had that union.

stamourv:
This change would accomplish your goal.
However, it would turn some errors that are currently run-time errors (e.g., `(exact? '())`) into type errors, which is not backwards compatible.
Not sure what the right resolution is here. @samth?

samth:
Yeah, I worry about that too. One possibility is to try it and see if it breaks any packages, and revert if so.

Maybe @rfindler has thoughts.

rfindler:
I would probably leave it alone.

I find that racket's inheritance of predicates like `positive?` that raises an error on non-`real?`s to be annoying, tho, which suggests that if I were starting over, I would probably have set the types the way @philnguyen wants them to be.

For me, this doesn't meet the bar to be backwards incompatible. 

Think of it this way @philnguyen : imagine you have an TR code lying around that you prepared as a demo for someone when you meet them at a conference. A few years later, after you've updated Racket a few times and, say, moved to a machine that doesnt' run racket v6.x anymore (older versions of Racket no longer run on my mac laptop, sadly so this does happen) you whip out that code to show someone who missed that epic epoch of your career. It was a cool demo after all. And what used to happen was neato things. And what happens now is "type error: exact? ... blah blah blah" and you mumble something and try to fix it, but there are like 10 places in the code ... and your awesome demo just makes you look dumb.

.... no one wants that. :)

philnguyen:
I see. I agree to just leave it alone then :)

PR #508, opened 2017-03-06
doc: comment on values vs. All

bennn:
An `(All (A) ....)` doesn't quantify over `(Values ....)` types


bennn:
Rendered:
screen shot 2017-03-06 at 11 03 53

PR #510, opened 2017-03-15
add linear expression objs, inequality props, and refinements

pnwamk:
# In this PR
This PR adds some of the basic atoms for refinement types, namely the following:

- Linear Expression objects (see object-rep.rkt)

- Linear Inequality propositions (see prop-rep.rkt)

- a solver for systems of linear inequalities (see logic/ineq.rkt)

- intersection types now carry any logical refinement proposition (tt when there is none) -- I chose to use intersection types for this role to avoid duplicating code. There are places in TR where, for example, we currently match on the literal type `(Vectorof τ)`, and these parts of the type checker should also work with refinements, intersections, etc. Now there just needs to be 1 additional line added (for intersections) instead of 2 or more (for intersections, refinements, etc).

- user-level syntax for the above abstractions

- tests for the above abstractions

# Not in this PR

- functions w/ dependent arguments

- the necessary tweaks for type inference w/ refinements to work on real programs (namely plumbing objects through the constraint generation process)

- actually producing refinement types by default anywhere when typechecking

# Other misc functionality

- the types of symbolic objects (and not just identifiers) can now also stored in the lexical environment (e.g. if an identifier is an alias for a linear expression, we have to keep the type of the linear expression that it aliases around so occurrence typing of more specific integer types still works for the identifier -- see type-env-structs.rkt, lexical-env.rkt, etc)

- `combine-props` in tc-metafunctions.rkt works harder to prove things (needed for refinement subtyping)

- there is now two contradiction functions: `atomic-contradiction?` and `contradiction?`. the lighter weight one is used for normalizing propositions, the stronger one is used for `combine-props` (i.e. when we're actually trying to prove/derive things)

- subtyping now considers refinement propositions and calls the logic portion of TR when necessary

- printing can print refinement types (printer.rkt)

- the parser (private/parse-type.rkt) has a new, compositional grammar for logical propositions. This is currently only used for parsing the `ψ` in `(Refine [x : τ] ψ)` but will also be used later for parsing function types.

- There is a new low-level mechanism for instantiating/abstracting term-level bindings in Reps (i.e. its like tc-subst but doesn't reason about the semantics of the types w.r.t. subtyping, overlap, etc -- see type-rep.rkt `subst-names`)

# To Do (before possibly ready for merge)

- [x] performance impact?
- [x] subtyping shouldn't cache for refinements
- [x] more tests (subtyping (logic, w/ envs), substitution, more?)
- [x] document syntax for propositions


samth:
Initial thoughts (before reading the code):

1. What are the performance impacts?
2. Do the things not in the PR impact our confidence in the correctness/testing of the things here?
3. What's the syntax, and is it documented?

pnwamk:
3. Syntax: The new syntax is not documented in this PR -- it is currently as follows:

Refinements are written `(Refine [x : τ] p)` where `τ` is valid type syntax (unchanged) and `p` is defined by:
```
o ::= x | (car o) | (cdr o)
t ::= o | (* n o)
l ::= n | t | (+ t t ...)
p ::= Top | Bot | (: l t) | (! l t) | (and p p ...) | (or p p ...) | (not p) 
          | (when p p) | (unless p p) | (if p p p) 
          | (<= l l) | (< l l) | (> l l) | (>= l l) | (= l l)
```

2. Correctness: This PR should be mostly testable and self contained (although it does need a little more testing at the moment). The internal data structures that are added area straightforwardly testable, and the tweaks to subtyping and the logic are as well.

1. Performance is largely unaffected (which was sort of what I expected since the new code isn't hit in general by non-refinement types, etc), with a minor caveat -- most of the micro benchmarks I use for TR were unaffected, except:

`math-flonum.rkt` got slightly slower (3.45 to 3.7sec) -- this seems to primarily stem from the fact that our `normalize-id` function (which maps free-identifier=? ids to the same id) now uses a weak hash table instead of a standard free-id-table. This is necessary now since we gensym identifiers for refinements (and later we will for dependent function checking).

`schml-interp-cast.rkt` got slightly faster (23seconds to 21 seconds by my measurement)... no idea why (might be that the stronger logic used for extending the environment more quickly rules some cases out as uninhabitable... but I'm not sure)

pnwamk:
e076284 doesn't seem to affect typechecking type either

pnwamk:
I've added additional testing.

Unless there's things we want to change, the only other possible addition would be to add documentation for Refinement types, unless we want to add that with updated function syntax (which might be a more natural place to put the grammar for propositions... I dunno)

pnwamk:
Oh, two more things I need to add to this PR:
- [ ] well-formedness of the refinements (e.g. lexp expect integers, car/cdr expect pairs, vector-length expects vectors, etc)
- [ ] contract generation!

typed-racket-lib/typed-racket/private/parse-type.rkt
```diff
@@ -255,22 +279,158 @@
...
+                   (type-of-sym-obj-in-lexical-env (attribute o.val)))
+           #:attr val (attribute o.val)))
+
+(define-syntax-class symbolic-object
+  #:description "symbolic object"
+  #:attributes (val)
+  (pattern (:+^ ~! . l:linear-expression-body)
+           #:attr val (attribute l.val))
+  (pattern (:*^ ~! n:exact-integer o:symbolic-object-w/o-lexp)
+           #:fail-unless (check-non-lexp-obj-subtype-of (attribute o.val) -Int)
```
samth: Feels like it would be clearer to just bind the type and then check it.

typed-racket-lib/typed-racket/private/parse-type.rkt
```diff
@@ -456,7 +627,7 @@
                            (let loop ([ty t*])
                              (match ty
                                [(Union: _ elems) (andmap loop elems)]
-                               [(Intersection: elems) (andmap loop elems)]
+                               [(Intersection: elems _) (andmap loop elems)]
```
samth: Are occurrences in refinements really under a constructor?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -789,7 +845,7 @@
             (match (resolve ty)
               [(Function: _) #t]
               [(Union: _ elems) (andmap loop elems)]
-              [(Intersection: elems) (ormap loop elems)]
+              [(Intersection: elems _) (ormap loop elems)]
```
samth: Why ignore the refinement here?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -828,7 +884,7 @@
...
```
samth: Also here.

typed-racket-lib/typed-racket/rep/object-rep.rkt
```diff
@@ -6,43 +6,289 @@
...
+        [(? Object? p)
+         (values c (terms-set ts p (add1 (terms-ref ts p))))]
+        [(? name-ref/c var)
+         (define p (-id-path var))
+         (values c (terms-set ts p (add1 (terms-ref ts p))))])))
+  (make-LExp* const terms))
+
+
+;; LExp-add1
+(define/cond-contract (-lexp-add1 l)
```
samth: Why are add1 and sub1 treated specially?

typed-racket-lib/typed-racket/rep/object-rep.rkt
```diff
@@ -6,43 +6,289 @@
...
+     (apply -obj* (multiply-OptObjects o1 o2) os)]))
+
+
+;; multiply-Objects
+;; 1. if either is empty, the result is empty
+;; 2. if one is an object the other a constant-LExp?, then
+;; the result is the object scaled by the constant
+;; 3. if two non-constant objects are multiplied, the
+;; result is empty (since we do not represent non-linear
+;; expressions currently)
```
samth: I wonder if representing "non-linear" objects differently would help with error messages.

typed-racket-lib/typed-racket/rep/prop-rep.rkt
```diff
@@ -75,11 +79,75 @@
       [else (-id-path i)]))
   (make-NotTypeProp o t))
 
-(def-prop OrProp ([ps (listof (or/c TypeProp? NotTypeProp?))])
+
+;; an inequality between two linear inequalities
+;; NOTE: we keep these in a normal form where
```
samth: Something's missing here.

typed-racket-lib/typed-racket/rep/rep-utils.rkt
```diff
@@ -36,19 +36,30 @@
...
+           (exact-integer? (car x))
+           (exact-integer? (cdr x)))))
 
 (define (name-ref=? x y)
   (cond
     [(identifier? x)
      (and (identifier? y) (free-identifier=? x y))]
     [else (equal? x y)]))
 
-(define id-table (make-free-id-table))
```
samth: Why is a free-id-table not good here?

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -124,7 +140,8 @@
                 [args exact-nonnegative-integer?]
                 [struct? boolean?])
   #:base
-  [#:custom-constructor
+  [#:custom-constructor/contract
```
samth: Why are we writing all these contracts twice?

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -759,63 +788,156 @@
...
+(define/provide (Intersection-w/o-prop t)
+  (match t
+    [(Intersection: _ (TrueProp:) _) t]
+    [(Intersection: (list t) prop _) t]
+    [(Intersection: ts prop tset) (make-Intersection ts -tt tset)]))
+
+;; constructs {v ∈ t | prop(v)}
+(define/provide refine/prop
+  (case-lambda
+    [(t prop)
```
samth: This is for the case where v is not free in prop?

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -1363,3 +1487,71 @@
...
+    (match tgt
+      [(cons nm1 (Path: flds nm2))
+       (cons (inc-lvl nm1) (make-Path flds (inc-lvl nm2)))]
+      [(cons nm1 rst)
+       (cons (inc-lvl nm1) rst)])))
+
+
+;; name substitution -- like subst-rep in
+;; tc-subst but much simpler
+(define/cond-contract (subst-names rep targets)
```
samth: This is ok, but we really want to be moving away from substitution.

typed-racket-lib/typed-racket/static-contracts/combinators/proposition.rkt
```diff
@@ -0,0 +1,164 @@
...
+         (values "#<" ">")))
+   (display open port)
+   (fprintf port "λ")
+   (display " (" port)
+   (recur arg port)
+   (display ") " port)
+   (recur body port)
+   (display close port)])
+
+(struct flat-named-lambda/sc static-contract (name arg body)
```
samth: Purpose statements for these structs would be helpful.

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
```diff
@@ -31,33 +33,57 @@
...
             [(cons p ps)
              (match p
-               [(TypeProp: (Path: lo x) pt)
-                #:when (and (not (is-var-mutated? x))
-                            (identifier-binding x))
-                (let* ([t (lookup-type/lexical x Γ #:fail (λ (_) Univ))]
-                       [new-t (update t pt #t lo)])
-                  (and (not (Bottom? new-t))
-                       (loop ps negs (env-set-type Γ x new-t))))]
+               [(TypeProp: (Path: pes (? identifier? x)) pt)
```
samth: Why isn't this subsumed by the case below?

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
```diff
@@ -31,33 +33,57 @@
...
-                            [_ Γ]))])
-                 (and Γ (env-replace-props Γ (append atoms (env-props Γ)))))]))]
+               [(? NotTypeProp?)
+                (loop ps (cons p negs) new Γ)]
+               [_ (loop ps negs new Γ)])]
+            [_ (let loop ([negs negs]
+                          [new new]
+                          [Γ Γ])
+                 (match negs
+                   [(cons (NotTypeProp: (Path: pes (? identifier? x)) pt) rst)
```
samth: Same as above.

typed-racket-lib/typed-racket/types/abbrev.rkt
```diff
@@ -264,3 +264,15 @@
 (define-syntax-rule (-object . ?clauses)
   (make-Instance (-class . ?clauses)))
 
+
+(define-syntax (make-refine stx)
```
samth: How is this different from the regular constructor?

typed-racket-lib/typed-racket/types/printer.rkt
```diff
@@ -149,15 +171,32 @@
     [(CdrPE:) 'cdr]
     [(ForcePE:) 'force]
     [(StructPE: t i) `(,(type->sexp t)-,i)]
-    [(SyntaxPE:) 'syntax]
```
samth: Why are we dropping the syntax accessor?

typed-racket-lib/typed-racket/types/printer.rkt
```diff
@@ -149,15 +171,32 @@
...
     [else `(Invalid Path-Element: ,(struct->vector pathelem))]))
 
 ;; object->sexp : Object -> S-expression
 ;; Print an Object (see object-rep.rkt) to the given port
 (define (object->sexp object)
   (match object
     [(Empty:) '-]
-    [(Path: pes n) (append (map pathelem->sexp pes) (list (name-ref->sexp n)))]
+    [(Path: pes n)
+     (let ([pes (map pathelem->sexp pes)])
```
samth: The old code was shorter. :)

typed-racket-lib/typed-racket/logic/ineq.rkt
```diff
@@ -0,0 +1,920 @@
...
+;;
+;; Decides satisfiability, implication, etc... of/between systems of linear inequalities
+;; of the form ([(a_1x_1 + a_2x_2 + ... + a_c) <= (b_1x_1 + b_2x_2 + ... + b_c)] ...)
+;; *****************************************************************************
+
+
+(define (deduplicate xs)
+  (hash-keys (for/hash ([x (in-list xs)]) (values x #t))))
+
+(define-match-expander lexp:
```
pnwamk: speed -- we don't need any fancy features, just a pair.

typed-racket-lib/typed-racket/logic/ineq.rkt
```diff
@@ -0,0 +1,920 @@
...
+       (loop c (terms-set h var (+ coeff (terms-ref h var))) zxs-rst)]
+      [(cons constant zxs-rst)
+       (loop (+ constant c) h zxs-rst)])))
+
+; lexp-coeff
+(define/cond-contract (lexp-coeff l x)
+  (-> lexp? any/c exact-integer?)
+  (terms-ref (lexp-terms l) x))
+
+(define-syntax (check-true stx)
```
pnwamk: I can put them in a test module. If they're in a test module, then it's fine if our test module depends on rackunit? (i.e. that won't add unwanted runtime deps, right?)

typed-racket-lib/typed-racket/logic/ineq.rkt
```diff
@@ -0,0 +1,920 @@
...
+    [else
+     (match exp
+       [(lexp: c h)
+        (lexp (* c a)
+              (terms-scale h a))])]))
+
+(module+ test
+  (check-equal? (lexp-set (lexp* 17 '(42 x)) 'x 0)
+                (lexp* 17)))
+; lexp-set
```
pnwamk: personal preference (I like the different color flagging function defs) -- I can add descriptions as well.

typed-racket-lib/typed-racket/logic/proves.rkt
```diff
@@ -0,0 +1,20 @@
...
+ ("../typecheck/tc-subst.rkt" (instantiate-rep/obj)))
+
+(provide (rename-out [implies-in-env?-for-stupid-infer-unit
+                      implies-in-env?]
+                     [instantiate-rep/obj-for-stupid-infer-unit
+                      instantiate-rep/obj]))
+
+;; the units for infer currently make it impossible to do
+;; lazy requires
+(define (implies-in-env?-for-stupid-infer-unit env p q)
```
pnwamk: They're lazy required to avoid cycles -- they can't appear at the top level w/o evaluating the lazy reference, so I eta expanded them.

typed-racket-lib/typed-racket/private/parse-type.rkt
pnwamk: I guess not... I'll fix these

typed-racket-lib/typed-racket/rep/object-rep.rkt
pnwamk: add1 is used for negation and expressing strict comparisons -- sub1 is useful when we recognize something like a strict comparison (e.g. while printing) and we want to remove the +1 to print it more cleanly w/ < or > instead of <= or >=

typed-racket-lib/typed-racket/rep/rep-utils.rkt
pnwamk: we now "gensym" identifiers for temporary idents when working w/ refinements, etc -- we don't want them to hang around forever, so we use a weak-hash-table that uses free-identifier=? for comparisons

typed-racket-lib/typed-racket/rep/type-rep.rkt
pnwamk: constructors can have a different interface than the underlying object (e.g. different number of arguments, different restrictions, etc)

typed-racket-lib/typed-racket/rep/type-rep.rkt
pnwamk: or if `prop` is already in the "nameless-local" representation form and we just want to put it back together without extra instantiating/abstracting

typed-racket-lib/typed-racket/rep/type-rep.rkt
samth: Right, but in almost all cases the "obvious" contract is the right one, and we're just duplicating.

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
pnwamk: most types are stored w/ their associated identifier.

Occasionally, however, we want to record the type of something w/o an underlying identifier (i.e. it has a DeBruijn index, its a linear expression, etc), and so when this occurs we store the type of the whole Object in the type environment.

It might be nice to make the lexical environment ONLY talk about the types of objects (and not identifies), but that would be a larger refactoring... which I remember considering... and I can't remember why I didn't just do it.

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
pnwamk: same as above

typed-racket-lib/typed-racket/types/abbrev.rkt
pnwamk: There is not regular constructor of this form. The exposed constructor for intersections just intersects a bunch of types really simply. this provides a nice constructor to add a refinement to an arbitrary type, and then the helpers in `type-rep.rkt` can worry about what needs to be done to do this.

typed-racket-lib/typed-racket/types/printer.rkt
pnwamk: they're not the same -- can you spot the different?

pnwamk: ha ha -- identifiers used to be printed as `(x)`, now they're printed as `x`

typed-racket-lib/typed-racket/base-env/base-types-extra.rkt
```diff
@@ -10,11 +10,19 @@
               ...
               (provide nm) ...)]))
 
+(define-syntax (define-other-props stx)
```
pnwamk: oops -- I need to move `Top` and `Bot` here.

typed-racket-lib/typed-racket/types/printer.rkt
samth: ha

typed-racket-lib/typed-racket/infer/infer-unit.rkt
```diff
@@ -524,17 +525,22 @@
...
         ;; find *an* element of elems which can be made a subtype of T
-        [((Intersection: ts) T)
+        [((Intersection: ts raw-prop) T)
          (cset-join
           (for*/list ([t (in-list ts)]
                       [v (in-value (cg t T))]
                       #:when v)
             v))]
         
         ;; constrain S to be below *each* element of elems, and then combine the constraints
```
pnwamk: currently it's ignored (which is sound -- we're just generalizing the type a little). I suppose a simple and slightly better thing to do would be extend the lexical environment w/ the proposition so it's in scope for the recursive calls where it might be useful.

typed-racket-lib/typed-racket/logic/ineq.rkt
```diff
@@ -0,0 +1,920 @@
...
+
+;; Leq->leq
+(define (Leq->leq l)
+  (match l
+    [(LeqProp: (LExp: c1 ts1) (LExp: c2 ts2))
+     (leq (lexp c1 ts1) (lexp c2 ts2))]))
+
+
+;; *****************************************************************************
+;; Fourier-Motzkin elimination implementation
```
pnwamk: this file is currently FME and the few functions that create the API the rest of TR uses to perform SAT queries. I guess we could break it up further... I don't think it's obscenely big as is ¯\_(ツ)_/¯

typed-racket-lib/typed-racket/private/type-contract.rkt
pnwamk: Okay, so I'm fixing the other ones -- but here we're ensuring a type is a function... and if it's a function, it should obviously be so, not only in a refinement prop. So I think it's fine to ignore the proposition here (it's sound and the argument for "complete" here seems silly)

typed-racket-lib/typed-racket/private/type-contract.rkt
pnwamk: same comment as above -- it's sound to ignore the refinement prop and we don't gain anything (as far as I can see) from including it

bennn:
> What are the performance impacts?

I compiled the Typed Racket programs from the [gradual-typing-performance](https://github.com/nuprl/gradual-typing-performance) repo on Racket v6.8 and this branch.
- compile times are 0.5 to 3 seconds slower
- runtimes are almost the same
  - but for some reason `fsmoo` and `take5` (object oriented programs) are consistently 50 milliseconds slower

Here are the raw timings (compile times first, scroll down for running times)
https://docs.google.com/spreadsheets/d/1S9PccVnGmzh8y0GhWbsjU6-JQuPoZ_6dTQ9nTw21stk/edit?usp=sharing

Here are the programs. If you unpack this archive, edit the Makefile to point to your "control" and "experimental" Racket binaries, and run `make` you should be able to reproduce this little experiment.
[refine.tar.gz](https://github.com/racket/typed-racket/files/860321/refine.tar.gz)

mfelleisen:
@benn Thanks for keeping an eye open on performance. In my capacity as PLT’s CPS, I herewith appoint you to ‘performance czar’ of Typed Racket.

typed-racket-lib/typed-racket/infer/infer-unit.rkt
pnwamk: Ah -- this needs to happen after inference gets objects.

typed-racket-lib/typed-racket/types/abbrev.rkt
samth: See definition of `refine/prop`.

pnwamk:
This PR is pretty much wrapped up. All the intended features are in. More vanilla programs can typecheck (e.g. we do a more thorough job proving expected propositions now) and there is a flag for typechecking with linear integer arithmetic reasoning.

Here are the current performance impacts (based on a small sample of files/programs):

image

Things are a little slower compared to HEAD, but we're still doing well compared to TR before some of the recent performance work that went into the 6.8 release. Hopefully we can identify and further mitigate any performance overhead that is added here, but for the moment we think this is reasonable for this PR.

PR #514, opened 2017-03-17
fix port-> functions to match new kw args in impl

pnwamk:
fixes types of `port->*` functions to match new keyword argument added by Racket commit samth/racket@1309cf164930cca4dfed4956db8ff1048ded3d7f


samth:
Sounds good.

PR #519, opened 2017-03-27
update path-add-extension type with new optional argument

bennn:
This seems to work --- is adding 2 new cases to the `cl->*` the right way to support the optional argument?


samth:
That looks right to me.

bennn:
Looks like the build failed because Travis doesn't have commit https://github.com/racket/racket/commit/2cbd44d64d815faeb939421cf448fbe2ae7635d0 yet. Guess I'll restart the build tomorrow evening.

PR #521, opened 2017-03-29
fix bug, make funapp code more readable

pnwamk:


typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt
```diff
@@ -69,48 +80,64 @@
       ;; any kind of dotted polymorphic function without mandatory keyword args
       [(PolyDots: (list fixed-vars ... dotted-var)
                   (Function/arrs: doms rngs rests drests (list (Keyword: _ _ #f) ...) #:arrs arrs))
-       (handle-clauses
-        (doms rngs rests drests arrs) f-stx args-stx
+       (for/first-valid-inference
+        #:in-arr-fields ([dom doms]
```
samth: Might even be worth moving the `in-list` here instead of in the macro.

pnwamk:
@samth took your `in-list` advice and removed all of the non-helpful intermediate list building #allocationisnotfree

PR #525, opened 2017-03-30
add contracted unit tests to travis

pnwamk:
Currently we run unit tests with our internal development contracts on as a part of our DrDr testing... but we do this without actually recompiling Typed Racket. This leads to the following:

1. The test usually takes 30 or more minutes (lately it's been 50+ minutes)
2. we often don't find out about incorrect development contracts until DrDr builds, instead of finding our as part of the PR review process

This PR moves this testing to our Travis CI builds, which seems to add about 6 or 7 minutes to the overall run time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,9 @@ script:
 - raco setup -j 1 math
 - racket -l typed-racket-test -- --math
 - racket -l typed-racket-test/test-docs-complete
+- echo ';; please rebuild me' >> typed-racket-lib/typed-racket/utils/utils.rkt
+- PLT_TR_CONTRACTS=1 raco setup -D typed-racket && raco setup -D typed && raco setup -D typed-racket-test
+- racket -l typed-racket-test -- --unit
 - echo "done"
 
 after_script:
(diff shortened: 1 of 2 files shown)


pnwamk:
@samth perhaps this is a more reasonable alternative to our current DrDr 30+ minute option (well, right now 50+ minutes -- either way, rebuilding w/ contracts and running unit tests seems much, much faster)

Also, I believe I've turned on the daily builds for TR, so we should be guaranteed at least a build a day -- so if something is pushed to Racket that breaks our development contracts we'll find out in a day or two.

.travis.yml
```diff
@@ -27,6 +27,9 @@ script:
 - raco setup -j 1 math
 - racket -l typed-racket-test -- --math
 - racket -l typed-racket-test/test-docs-complete
+- echo ';; please rebuild me' >> typed-racket-lib/typed-racket/utils/utils.rkt
```
bennn: Could you just `touch` the file?

pnwamk: raco setup will not rebuild because a file was `touch`ed

samth:
I think we should keep the test in DrDr, although maybe we need to disable
it until we can do something about the timeout.

PR #526, opened 2017-03-31
Fixes #434 (Type of literal char is not the literal char itself)

### LINKED ISSUE #434: Type of literal char is not the literal char itself

6.6.0.5
### What program did you run?

``` racket
#lang typed/racket
(ann #\nul #\nul)
(ann #\a #\a)
(ann #\b '#\b)
```

The type of a char `#\c` should be `#\c`.

```
Type Checker: type mismatch
  expected: #\nul
  given: Char in: #\nul
```

jsmaniac:
Using the solution from Alex Knauth.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-literal.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-literal.rkt
@@ -28,7 +28,7 @@
 ;; tc-literal: racket-value-syntax [type] -> type
 (define (tc-literal v-stx [expected #f])
   (define-syntax-class exp
-    (pattern (~and i (~or :number :str :bytes))
+    (pattern (~and i (~or :number :str :bytes :char))
              #:fail-unless expected #f
              #:fail-unless (let ([n (syntax-e #'i)])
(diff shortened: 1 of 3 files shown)


pnwamk:
This may not be necessary just yet, since it looks like this modification only affects typechecking when there is an _expected type_, but perhaps we should add another case to the `generalize` function (found here https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/types/generalize.rkt) which checks if the type is a subtype of `Char` and upcasts to it if so.

jsmaniac:
@pnwamk I had added some tests to check that the generalisation occurs, but I wasn't sure where to modify that and if it needed modification (assuming that the tests would catch it if the generalisation did not occur).

I just pushed a case for `-Char` to that file, is that what you meant?

Also, are the tests in `typed-racket-test/succeed/` executed or at least compiled on Travis? I couldn't find in the log any reference to the test file I added.

pnwamk:
Yes, the change you made was the one I was alluding too.

The `typed-racket-test/succeed/` directory is included in the integration tests (i.e. `racket -l typed-racket-test -- --int`)... which is not obvious without digging around in the test framework unfortunately =)

LGTM

pnwamk:
merged via PR #527

PR #529, opened 2017-04-06
existential objects for function args

pnwamk:
This change (the latest one 	6b29b47 ... the others are from in the other refinement PR) adds existential objects for arguments in function applications so we can more accurately track detailed information about the types of things. In this PR they are only added when the linear-integer flag is on, and they are erased at the exit points for abstraction typechecking (i.e. so repeated applications of functions don't produce contradictory information) (note to self -- we might be able to get away with only erasing them when we've finished typechecking an abstraction AND there is no expected type (if there was an expected type, we shouldn't be "leaking" info anyway, we just check if its below the type))

With these changes, all these typecheck (and are added to our test suite):
```
(assert<= (min one two three) three)
(assert< (min one two) three)
(assert<= ((λ () (min one two three))) three)
(assert< ((λ () (min one two))) three)
(assert<= (let ([x (min one two three)]) x) three)
(assert< (let ([x (min one two)]) x) three)
(assert< (- (min one two) 1) three)
```

(note: the `assert`s are typecheck asserts -- the term will not typecheck if the inequality is not proven valid by the type checker)

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt
@@ -6,8 +6,9 @@
          (env tvar-env)
          (for-syntax syntax/parse racket/base)
          (types utils subtype resolve abbrev
-                substitute classes prop-ops)
+                substitute classes prop-ops type-table)
          (typecheck tc-metafunctions tc-app-helper tc-subst)
+         racket/sequence
          (rep type-rep)
          (r:infer infer))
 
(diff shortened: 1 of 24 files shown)


pnwamk:
w.r.t. perf -- I ran four benchmarks (2 small, 2 multi-file programs). 3 were unaffected, new-metrics.rkt mysteriously seemed to incur about a 3% overhead. I've poked and poked and don't really understand the overhead. I wonder if new-metrics is a relatively simple file to typecheck and the fact that a few things are no longer tail calls, or that we're checking the values of some parameters regularly is thus visible in that setting ¯\_(ツ)_/¯

samth:
Lots more parameter de-referencing could certainly be a performance hit. How necessary is that?

pnwamk:
tried replaced it w/ unbox-ing a boxed value, didn't seem to make a difference

typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt
```diff
@@ -55,7 +56,23 @@
...
-  (match-define (list (tc-result1: argtys (PropSet: argps+ argps-) argobjs) ...) args-res)
+  (define-values (argtys argprops argobjs)
+    (cond
+      [(with-linear-integer-arithmetic?)
+       (for/lists (_1 _2 _3) ([stx (in-syntax args-stx)]
+                              [res (in-list args-res)])
+         (match res
+           [(tc-result1: t (PropSet: p+ p-) (? Empty?))
+            (define o (-id-path (gen-existential-id)))
+            (add-existential-obj! stx o)
```
samth: Why is this done with mutation?

pnwamk: The other option would be to have a function like `single-value` (i.e. that typechecks an expression asserting it returns a single value) that indicates the expression should be assigned a existential object if it does not have a non-empty object, and then that existential object would be what is added to the type table initially (meaning we don't need this function to add it after the fact)

samth: Ah, ok, this is only about the type table.

pnwamk:
these changes were incorporated in 8aa05bebff09f6a947ee5b9e4f9ca709ff82630c

PR #531, opened 2017-04-11
Use code-inspectors to prevent optimization in unsafe contexts.

thinkmoore:
The optimizer should only run when the code being compiled could
directly access racket/unsafe/ops. This prevents unsoundness in Typed
Racket from giving untrusted code access to dangerous operations.

### DIFF
--- a/typed-racket-lib/typed-racket/tc-setup.rkt
+++ b/typed-racket-lib/typed-racket/tc-setup.rkt
@@ -26,9 +26,12 @@
 
 (define (maybe-optimize body)
   ;; do we optimize?
-  ;; only if the current module requested optimization and the
+  ;; only if the current module requested optimization,
+  ;; PLT_TR_NO_OPTIMIZE is not set, and the
   ;; current code inspector has sufficient privileges
-  (if (and (optimize?) (authorized-code-inspector?))
+  (if (and (optimize?)
+           (not (getenv "PLT_TR_NO_OPTIMIZE"))
+           (authorized-code-inspector?))
       (begin
         (do-time "Starting optimizer")
         (begin0 (stx-map optimize-top body)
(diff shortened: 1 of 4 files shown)


typed-racket-lib/typed-racket/tc-setup.rkt
```diff
@@ -17,9 +17,18 @@
...
+(define (authorized-code-inspector?)
+  ;; Check whether the current code inspector has sufficient
+  ;; privileges to access unsafe operations.
+  (with-handlers ([exn:fail? (lambda (e) #f)])
+    (module->namespace 'racket/unsafe/ops)
+    #t))
+
 (define (maybe-optimize body)
   ;; do we optimize?
-  (if (and (optimize?) (not (getenv "PLT_TR_NO_OPTIMIZE")))
```
samth: I think we should keep the environment variable -- it might be useful for other things.

thinkmoore: Added this back in and made note of it in the documentation.

mflatt:
I'm not sure, but I think
```
(inspector-superior?
 (current-code-inspector)
 (make-inspector
  (variable-reference->module-declaration-inspector
   (#%variable-reference))))
```
might be a more appropriate test. That should check whether the current code inspector allows access to the enclosing TR module, which I think should be the same as whether the current code inspector allows access to unsafe operations. (Otherwise, references generated by the TR optimizer won't work.)

thinkmoore:
I originally had written (a hackier) version of that check. The reason I switched to trying to access racket/unsafe/ops was that it has the side effect that if for some reason typed racket was loaded in a context that didn't have access to racket/unsafe/ops, it meant that typed racket would silently turn off optimization and still work, which seemed nice.

mflatt:
Ok. I don't see any problem with that strategy.

PR #533, opened 2017-04-16
move refine and linear integer stuff in docs

pnwamk:
Moves the refinement and linear integer arithmetic docs addition to the experimental section of the docs for now.


stamourv:
If I'm reading this correctly, this commit should be included in 6.9, correct?
(Otherwise, the docs would show those as non-experimental in 6.9, but experimental in 6.9.0.1, which is strange.)

samth:
Yes, that's the idea.

PR #534, opened 2017-04-21
provide for/flvector, for/extflvector, etc. as aliases for : versions

### LINKED ISSUE #489: `for/flvector` and `for/extflvector` documented but not provided?

AlexKnauth:
Since the type-checker can now handle their expansions.

This fixes #489. However, one problem is that this will cause an error for programs that use both `(require racket/flonum)` and `(require typed/racket)`. Is this okay, or should we keep it the way it is with the `:` versions of these forms?

One of the programs that does this is the math library. It would have to be fixed by adding `(except-in racket/flonum for/flvector for*/flvector)`.


AlexKnauth:
Would the better solution be to provide them from a new `typed/racket/flonum` library?

samth:
We should not be changing things in ways that break the math library, since
that suggests that other code will break as well. A separate library seems
like a good idea.

AlexKnauth:
Okay, I've changed it to provide them from `typed/racket/flonum` and `typed/racket/extflonum` instead.

samth:
Looks good.

PR #540, opened 2017-05-10
Fixes type of udp-open-socket

jsmaniac:
Many thanks to s455wang on IRC for finding the bug, and to @apg for finding the incriminated line & fix.

I put the post precise type I could, although it really should be `(In-Range 1 65535)`, preferably bound to `Port-Number`, when we get more precise numeric types (ISTR that there is a PR for that or at least someone working on it?). Should I change it to `Exact-Nonnegative-Integer`, or even to `Integer` for convenience?

Here are the docs for `udp-open-socket`: https://docs.racket-lang.org/reference/udp.html?q=udp-open#%28def._%28%28lib._racket%2Fudp..rkt%29._udp-open-socket%29%29

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -2738,19 +2738,19 @@
 
 ;; Section 15.3.2 (racket/udp)
 [udp-open-socket (->opt [(-opt -String) (-opt -Int)] -UDP-Socket)]
-[udp-bind! (-> -UDP-Socket (-opt -String) -Nat -Void)]
-[udp-connect! (-> -UDP-Socket (-opt -String) (-opt -Nat) -Void)]
+[udp-bind! (-> -UDP-Socket (-opt -String) -Int -Void)]
+[udp-connect! (-> -UDP-Socket (-opt -String) (-opt -Int) -Void)]
 
-[udp-send-to (->opt -UDP-Socket -String -Nat -Bytes [-Nat -Nat] -Void)]
-[udp-send (->opt -UDP-Socket -Bytes [-Nat -Nat] -Void)]
-[udp-send-to* (->opt -UDP-Socket -String -Nat -Bytes [-Nat -Nat] B)]
-[udp-send* (->opt -UDP-Socket -Bytes [-Nat -Nat] B)]
-[udp-send-to/enable-break (->opt -UDP-Socket -String -Nat -Bytes [-Nat -Nat] -Void)]
-[udp-send/enable-break (->opt -UDP-Socket -Bytes [-Nat -Nat] -Void)]
+[udp-send-to (->opt -UDP-Socket -String -Int -Bytes [-Int -Int] -Void)]
+[udp-send (->opt -UDP-Socket -Bytes [-Int -Int] -Void)]
+[udp-send-to* (->opt -UDP-Socket -String -Int -Bytes [-Int -Int] B)]
+[udp-send* (->opt -UDP-Socket -Bytes [-Int -Int] B)]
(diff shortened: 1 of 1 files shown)


apg:
I don't know enough about typed-racket, but it seems as though the `port-number?` Contract is defined as `(between/c 1 65535)` which seems like the appropriate way to specify this? So, `In-Range` seems most correct to me.

jsmaniac:
@apg `In-Range` doesn't exist — yet ;-) . Typed Racket has types for common integer sizes (Byte, Fixnum, Index, Integer), but not something as precise as `In-Range`, for now.

Also thanks for mentioning the `(between/c 1 65535)`, I had accidentally based myself on the range for `listen-port-number?` instead of `port-number?`. Just fixed that right now.

samth:
Typically we just use `Integer` for types like this, even if there's a more constrained range. It's often painful to enforce the subrange statically.

jsmaniac:
@samth I thought since the port will often be a constant (possibly stored in a configuration variable of some sort), it wouldn't be so much of a problem. But since fixnum is already too large, it makes sense to put `-Nat` for convenience.

Updated just now.

stamourv:
@jsmaniac: As @samth said, we usually use `Integer` for those, not `Natural`. Enforcing even just the sign statically can be painful.

jsmaniac:
@stamourv Oops, sorry. I had intended to put `Integer`, but mindlessly copied the `-Nat` from `udp-bind!` just below.

Morale: I shouldn't fix types before coffee in the morning.

Hopefully I didn't get this wrong this time :) .

stamourv:
Actually, could be worth relaxing that one (and the one below) too.

jsmaniac:
@stamourv I can push that change too, but actually there's a whole load of UDP-related functions which use `-Nat`:

```
[udp-bind! (-> -UDP-Socket (-opt -String) -Int -Void)]
[udp-connect! (-> -UDP-Socket (-opt -String) (-opt -Int) -Void)]

[udp-send-to (->opt -UDP-Socket -String -Nat -Bytes [-Nat -Nat] -Void)]
[udp-send (->opt -UDP-Socket -Bytes [-Nat -Nat] -Void)]
[udp-send-to* (->opt -UDP-Socket -String -Nat -Bytes [-Nat -Nat] B)]
[udp-send* (->opt -UDP-Socket -Bytes [-Nat -Nat] B)]
[udp-send-to/enable-break (->opt -UDP-Socket -String -Nat -Bytes [-Nat -Nat] -Void)]
[udp-send/enable-break (->opt -UDP-Socket -Bytes [-Nat -Nat] -Void)]

[udp-receive! (->opt -UDP-Socket -Bytes [-Nat -Nat] (-values (list -Nat -String -Nat)))]
[udp-receive!* (->opt -UDP-Socket -Bytes [-Nat -Nat] (-values (list (-opt -Nat) (-opt -String) (-opt -Nat))))]
[udp-receive!/enable-break (->opt -UDP-Socket -Bytes [-Nat -Nat] (-values (list -Nat -String -Nat)))]

;; […]

[udp-send-to-evt (->opt -UDP-Socket -String -Nat -Bytes [-Nat -Nat]
                        (-evt -Void))]
[udp-send-evt (->opt -UDP-Socket -Bytes [-Nat -Nat]
                     (-evt -Void))]
[udp-receive!-evt
 (->opt -UDP-Socket -Bytes [-Nat -Nat]
        (-evt (-lst* -Nat -String -Nat)))]
```

Should I update all of those (I feel a bit unsure about relaxing a whole bunch of functions)?

I guess all of these would be good candidates for @bennn's `trivial`.

stamourv:
@jsmaniac: Yes, may as well update them too. That's backwards compatible.

jsmaniac:
@stamourv Done! IIRC you can squash the two commits directly while merging on GitHub, so I left them separate for clarity.

stamourv:
Merged, thanks!

PR #544, opened 2017-05-12
Fixes 539: Type of literal regexp is not the literal regexp itself

jsmaniac:
I haven't tried this locally. If the (large number) of new tests all succeed, then it should be fine.

I was a bit out of inspiration for negative tests. I checked that none of the `Regexp`, `PRegexp`, `Byte-Regexp` and `Byte-PRegexp` have the wrong subtyping relationship. Suggestions for other negative tests are welcome!


jsmaniac:
In https://github.com/racket/typed-racket/blob/master/typed-racket-test/unit-tests/typecheck-tests.rkt, the new tests trigger errors because they are using `-Regexp`, `-PRegexp`, `-Byte-Regexp` and `-Byte-PRegexp`, which are not defined (see the log https://travis-ci.org/racket/typed-racket/builds/231491850#L346).

I tried adding the following after [line 294](https://github.com/racket/typed-racket/blob/master/typed-racket-test/unit-tests/typecheck-tests.rkt#L294), but the tests still fail:

```
(only-in typed-racket/types/abbrev -Regexp -PRegexp -Byte-Regexp -Byte-PRegexp)
```

What do i need to do to make the `-Regexp`, `-PRegexp`, `-Byte-Regexp` and `-Byte-PRegexp` tests available to tests in `typecheck-tests.rkt` ?

pnwamk:
In the `ann` type check tests you want actual Typed Racket syntax for a type (e.g. `Regexp`) not the internal identifier `-Regexp`.

jsmaniac:
In generalize.rkt, we could either use `[(? (lambda (t) (subtype t -Base-Regexp))) -Base-Regexp]` (to generalize literals which are `regexp?` to `-BaseRegexp`), or instead generalize to `-Regexp`, i.e. `[(? (lambda (t) (subtype t -Base-Regexp))) -Regexp]`.

I tried to see if it made any difference in the way the type is printed when entering a literal `#rx"abc"` at the REPL, but it makes no difference.

I also expected `(let ([x #rx"foo"]) (set! x #px"foo"))` to fail in the second case (generalizing to `-Base-Regexp`), but it still works (so it seems something else is further generalizing `-Base-Regexp` to `-Regexp` anyway).

I feel this will matter only if/when we decide to make the `Base-Regexp` type "public" (for now it seems to be used internally only).

I think if we expose Basic-Regexp at some point, the user could always force it with `(ann #rx"abc" Basic-Regexp)` in both cases.

@pnwamk tentatively suggested on Slack to use the more general `-Regexp`, so I went with that (it's also backwards-compatible, in that currently the type of `#rx"abc"` is `-Regexp`), but I thought I would mention the choice here for future reference.

typed-racket-lib/typed-racket/types/generalize.rkt
```diff
@@ -28,6 +28,10 @@
         [(? (lambda (t) (subtype t -FloatComplex))) -FloatComplex]
         [(? (lambda (t) (subtype t -SingleFlonumComplex))) -SingleFlonumComplex]
         [(? (lambda (t) (subtype t -Number))) -Number]
+        [(? (lambda (t) (subtype t -Base-Regexp))) -Regexp]
```
samth: I think either this should generalize to `-Base-Regexp`, or they should both generalize to `-Regexp`.

jsmaniac: @samth The current implementation in TR generalises `#rx` to `-Regexp`, and `#px` to `-PRegexp`:

```
(let ([x #px"foo"]) (set! x #rx"foo") x)
error: eval:1:20: Type Checker: type mismatch;
  mutation only allowed with compatible types
   expected: PRegexp
   given: Regexp
   in: (set! x #rx"foo …

(let ([x #rx"foo"]) (set! x #px"foo") x) ;; works fine
```

Although the likelihood of this actually breaking anything is exceedingly low, changing `#px` so that it generalises to `-Regexp`  is in principle a breaking change:

```
;; works for now, fails if we make #px generalise to -Regexp
(let ([x #px"foo"]) (ann x PRegexp))
```

Changing it the other way round (so that `#rx` generalises to `-Base-Regexp`) would not break anything for now, as `-Base-Regexp` somehow magically turns into `-Regexp` anyway (I don't know what code is responsible of this). But if we change that "second generalisation", e.g. when making `Base-Regexp` available, then it would be a breaking change:

```
;; works for now, fails if we make #rx generalise to -Base-Regexp
(let ([x #rx"foo"]) (set! x #px"bar") x)
```

I'm willing to make the first breaking change, though (generalising `#px` to `-Regexp`), as it will most likely reduce surprises, and it is unlikely that there is some code relying on this in the wild. But I'd prefer confirmation before going down that road :)

samth:
After this discussion (and sorry for the delay) the current approach seems right. Can you rebase so it merges properly and the tests can re-run?

jsmaniac:
@samth Done! I don't have time to test locally this morning, so I'll let Travis do its job. Thanks for taking care of the merge!

samth:
Unfortunately something seems to have gone wrong in the travis settings. Can you push again to this branch to trigger a build?

jsmaniac:
@samth I force-pushed, it seems Travis properly started this time.

Issue #541, opened 2017-05-12
(∩ A-Struct A) gives nonsensical type (Rec rec179895 rec179895)


jsmaniac:
### What version of Racket are you using?
6.9.0.1

### What program did you run?
```racket
#lang typed/racket
(struct Parent ())
(struct Child ())
(define-type τ (∀ (A) (→ (∩ Child A) Any)))
;; At the REPL:
;(:type (τ Parent))
```

This prints the following type, which is wrong, and contains a self-referential recursive type which does not make any sense:

```racket
(-> (Rec rec179895 rec179895) Any)
```

### What should have happened?

The type `(τ Parent)` should be equivalent to `(→ Child Any)`.

-----

Since this is an issue with intersection types and polymorphism, it might be related to #532 .

pnwamk:
fixed via #549

Issue #532, opened 2017-04-12
Type Checker: I is not bound as a type in: here


jsmaniac:
### What version of Racket are you using?
6.9.0.1

### What program did you run?

I ran the program below, and typed at the REPL `(:print-type w1)`.

```
#lang typed/racket

(struct (A) I ([v : A]) #:transparent)
(struct (A) O ([v : A]) #:transparent)

(: worklist
   (∀ (A ...)
      (→ (List (Listof (∩ A (I Any))) ...)
         (List (→ (∩ A (I Any)) (∩ A (O Any))) ...)
         (List (Listof (∩ A (O Any))) ...))))

(define (worklist roots processors)
  (error "Not Implemented Yet"))

(define w1
  (inst worklist
        (U (I Number) (O String))
        (U (I Float) (O Boolean))))
```

### What should have happened?

The type of `w1` should be printed. If `I` was not bound as a type, the location should not be "here", but something more precise :) .

### If you got an error message, please include it here.

```
../../.racket/snapshot/pkgs/typed-racket-lib/typed-racket/utils/tc-utils.rkt:49:43: Type Checker: I is not bound as a type in: here
```

The error also occurs when typing the following code at the REPL, but this code does not cause any problem if I delete the definition of `w1` (which has nothing to do with the code below):

```
(ann (λ ([x : (I Number)])
         (O (number->string (I-v x))))
       (→ (∩ (U (I Number) (O String)) (I Any))
          (∩ (U (I Number) (O String)) (O Any))))
```

Also note that if I hover the `worklist` in the definition of `w1` with my mouse, I can see that it is given the following type (the `#(struct:I (Any))` and `#(struct:O (Any))` look suspicious):

```
(-> (List
     (Listof (∩ (U (I Number) (O String)) #(struct:I (Any))))
     (Listof (∩ (U (O Boolean) (I Flonum)) #(struct:I (Any)))))
    (List
     (-> (∩ (U (I Number) (O String)) #(struct:I (Any)))
         (∩ (U (I Number) (O String)) #(struct:O (Any))))
     (-> (∩ (U (O Boolean) (I Flonum)) #(struct:I (Any)))
         (∩ (U (O Boolean) (I Flonum)) #(struct:O (Any)))))
    (List
     (Listof (∩ (U (I Number) (O String)) #(struct:O (Any))))
     (Listof (∩ (U (O Boolean) (I Flonum)) #(struct:O (Any))))))
```

### Context:

[...]

PR #549, opened 2017-05-13
fix intersection bug

pnwamk:
intersections with resolvable types would occasionally generate
spurious weird types (e.g. μx.x) when a type name
is not yet fully defined -- this patches that problem by
using resolve-once instead of resolve and checking the result
for #f before proceeding to compute the intersection

See #541 and #532.


samth:
I think this needs some tests (maybe just from the bug reports).

PR #555, opened 2017-05-21
Fix overlap mask error

### LINKED ISSUE #542: mask-accessor: contract violation   expected: mask?   given: #f
6.9.0.1

### What program did you run?
```racket
#lang typed/racket
(struct (A) S ([f : A]))
(define-type T (∩ (S Nonnegative-Integer) (S Nonpositive-Integer)))
```

The type `T` should be defined as `(S Zero)`. Note that the bug is not related to arithmetic types, it also occurs with other types like `String` and `Symbol`.

```
.racket/snapshot/pkgs/typed-racket-lib/typed-racket/types/overlap.rkt:40:0: mask-accessor: contract violation
  expected: mask?
  given: #f
```

Might be related to #447.

pnwamk:
fixes #542

### DIFF
--- a/typed-racket-lib/typed-racket/types/current-seen.rkt
+++ b/typed-racket-lib/typed-racket/types/current-seen.rkt
@@ -7,7 +7,7 @@
 
 
 ;;************************************************************
-;; Current Seen Continuation Mark
+;; Current Seen Continuation Mark for Subtyping
 ;;************************************************************
 ;;
 ;; Prevents infinite loops when subtyping calls outside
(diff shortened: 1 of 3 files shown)


pnwamk:
This PR changes calls to `resolve-once` (which can return #f if a type name is not yet defined) to `resolve` (which always returns a Type), and reduces allocation by extending the `seen` list for overlap only when needed.

Sorry for the whitespace changes -- remember to add `?w=1` to the diff URL to ignore whitespace changes!

pnwamk:
tidy diff link: https://github.com/racket/typed-racket/pull/555/commits/7ef2b54e313323984dbe372ed152dcfbd9ac5ee2?w=1

pnwamk:
there's a `#:no-order` on that clause, so we get it both ways (https://github.com/racket/typed-racket/pull/555/files#diff-5fe8f1857d5552b04caf253b9d9782d7R55)

samth:
Ah, I see now.

PR #556, opened 2017-05-28
WIP: tidy TR arrow rep, add dependent domains

pnwamk:
This WIP PR more or less has the following goals:

1. have cleaner representation of arrow types inside of TR, use interface functions for arrow types when it's more clear than manually tearing apart structs with a `match`, etc...
2. be more explicit about when function types are dependent
3. Within TR's implementation, use macros w/ clear names for bound variables instead of raw De Bruijn indices
4. add user-level syntax which uses clear variable names for bound variables and deprecate/discourage using the old raw De Bruijn index syntax in actual TR code
5. add dependent domains to our dependent functions

...well to be perfectly honest, the real goal is item 5, but our arrow representation and binder code is pretty complex right now, so I'm trying to make things easier to reason about so I can add dependent domains w/o just making everything slower and even more complex.

If anyone wants to comment regarding how we're doing things in this PR, it would be wonderful to get said comments earlier as opposed to later, since this PR will likely affect a lot of files and I'm not exactly swimming in free time this summer =)

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env-indexing-abs.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env-indexing-abs.rkt
@@ -231,7 +231,7 @@
 
    ;; flvector ops
 
-   [flvector? (make-pred-ty -FlVector)]
+   [flvector? (pred-> -FlVector)]
    [flvector (->* (list) -Flonum -FlVector)]
    [make-flvector (cl->* (-> index-type -FlVector)
                          (-> index-type -Flonum -FlVector))]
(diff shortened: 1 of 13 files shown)


pnwamk:
Currently (with just the first commit) things are all sorts of broken overall, but I have done the following:

i. added new Rep structs to represent arrows (see `Arrow`, `ArrowStar`, and `ArrowDep` in type-rep.rkt)
ii. added cleaner and more consistent code for instantiate and abstract operations (also in type-rep.rkt)
iii. upgraded the macros which are widely used throughout TR's implementation for creating function types to use actual names for bound function arguments in dependent types (i.e. now they won't use De Bruijns, so we should be guaranteed that things are correctly scoped, have the right indices assigned to them, etc)

pnwamk:
I've decided to do this in two steps:
1. refine our current arrow representation, making where things are dependent explicit (but not adding interdependent function arguments)
2. add interdependent function arguments

PR #557, opened 2017-06-05
require-typed: more informative error when struct field is missing :

bennn:
Today I made a mistake writing a `require/typed`, I was missing a colon in the struct fields. Something like:

```
(require/typed "file.rkt"
  (#:struct posn ([x Integer] [y Integer])))
```

The error message was confusing:
```
<pkgs>/typed-racket-lib/typed-racket/base-env/prims-contract.rkt:168.7: require-typed-struct: expected constructor-term
  at: (((x Integer) (y Integer)) #:type-name posn (submod ".." server))
  within: (require-typed-struct posn () ((x Integer) (y Integer)) #:type-name posn (submod ".." server))
  in: (require-typed-struct posn () ((x Integer) (y Integer)) #:type-name posn (submod ".." server))
  ....
```

After this commit, the error message is better:
```
<pkgs>/typed-racket-lib/typed-racket/base-env/prims-contract.rkt:168.7: require-typed-struct: expected the identifier `:'
  at: Integer
  in: (require-typed-struct posn () ((x Integer) (y Integer)) #:type-name posn (submod ".." server))
  parsing context: 
   while parsing struct field
    term: ((x Integer) (y Integer))
    location: <pkgs>/typed-racket-lib/typed-racket/base-env/prims-contract.rkt:169.29
  ....
```

- - -

Though, I'm not sure this is the right fix.

The problem is that [`require-typed-struct`](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/base-env/prims-contract.rkt#L449) is getting called on a syntax object with an empty list of type variables and malformed struct fields, like in the 2nd error message above:

```
(require-typed-struct posn () ((x Integer) (y Integer)) #:type-name posn (submod ".." server))
```

So, one other way to fix would be to check [here](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/base-env/prims-contract.rkt#L168) if the list of type variables is empty, and if so then omit the `()`

Maybe there's a 3rd way via `syntax-parse`? I don't know.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -449,8 +449,8 @@
   (define ((rts legacy) stx)
     (syntax-parse stx #:literals (:)
       [(_ name:opt-parent
-          (~optional (~seq (tvar:id ...)) #:defaults ([(tvar 1) '()]))
-          ([fld : ty] ...)
+          (~seq (tvar:id ...))
+          (~describe "struct fields" ([fld : ty] ...))
           (~var input-maker (constructor-term legacy #'name.nm))
           (~optional (~seq #:type-name type:id) #:defaults ([type #'name.nm]))
           unsafe:unsafe-clause
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -111,6 +111,10 @@
   (pattern nm:id #:with parent #'#f)
   (pattern (nm:id parent:id)))
 
+(define-syntax-class typed-field
+  #:attributes (field type)
+  #:literals (:)
+  (pattern [field:id : type]))
 
 (define-values (require/typed-legacy require/typed -unsafe-require/typed)
  (let ()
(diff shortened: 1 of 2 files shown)


bennn:
Travis rejected. Switching to option 2.

samth:
This looks like an improvement, although perhaps a dedicated clause for this case to produce a specific error would be even better.

bennn:
Ok I added a syntax class for "typed fields" and used it for structs and signatures.

I'm happy with this fix, much better than the last two. Any ideas for other code I can test it on?

samth:
What's the new error message?

bennn:
```
"require-typed-struct-missing-colon.rkt":
  making #<path:/Users/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-test/fail/require-typed-struct-missing-colon.rkt>
<pkgs>/typed-racket-test/fail/require-typed-struct-missing-colon.rkt:10.2: require/typed: expected the identifier `:'
  at: Integer
  in: (require/typed (submod ".." server) (#:struct posn ((x Integer) (y Integer))))
  parsing context: 
   while parsing typed-field
    term: (x Integer)
    location: <pkgs>/typed-racket-test/fail/require-typed-struct-missing-colon.rkt:11.20
   while parsing struct-clause
    term: (#:struct posn ((x Integer) (y Integer)))
    location: <pkgs>/typed-racket-test/fail/require-typed-struct-missing-colon.rkt:11.4
   while parsing clause
    term: (#:struct posn ((x Integer) (y Integer)))
    location: <pkgs>/typed-racket-test/fail/require-typed-struct-missing-colon.rkt:11.4
  compilation context...:
   /Users/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-test/fail/require-typed-struct-missing-colon.rkt
  context...:
```

samth:
Looks good to me.

PR #559, opened 2017-06-15
add types for Immutable-HashTable, Mutable-HashTable, Weak-HashTable

bennn:
The old 'HashTable' type is now the union of the other 3 hash types, `Immutable-HashTable` `Mutable-HashTable` and `Weak-HashTable`.

- all operations that used to work on 'HashTable's still work, but some now have more specific outputs
- `#hash` literals have type `ImmutableHash`
- `immutable?` and `hash-weak?` are filters

Minor notes:

- renamed internal identifiers containing 'Hashtable' to all use 'HashTable'
- add Racket guide/reference 'secref' functions

- - - 

~This PR has 2 "serious issues"~ EDIT: fixed

1. `(type->contract (HashTable k v))` now makes an `or/c` with 3 possibilities. What's the best way to fix this?
2. The type for `immutable?` is no good. It works to filter `(HashTable k v)` but not `(U #f (HashTable k v))` etcetera. But I couldn't think of a better type .... ideally I would flip the masks --- if immutable?` is true then zero-out the mask bits for `Mutable-HashTable` and `Weak-HashTable`. (That approach should scale to vectors & sets too.)


samth:
For issue 1, can you just add an explicit case [here](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/private/type-contract.rkt#L387) for the union type. See, for example, how `Listof:` is used there, even though that isn't a real type constructor either.

samth:
For issue 2, can you change the type of `immutable?` to be roughly `(Any -> Boolean : #:+ (not (mutable-ht a b)))`?

bennn:
> For issue 2 ..... `(Any -> Boolean : #:+ (not (mutable-ht a b)))`

I like that type, but I don't know how to bind `a` and `b`.

(Here's what I tried: `(asym-pred Univ B (-PS (-not-type 0 (-Mutable-HT a b)) ()))`)

samth:
Does polymorphism not work for that?

bennn:
Hmm, polymorphism works to bind `a` and `b` in the type,
but then it doesn't filter unless I instantiate `immutable?`

... I guess I prefer this to the current solution.

Example:

```
#lang typed/racket/base

(: f (-> (HashTable Integer Integer) Integer))
(define (f h)
  (cond
   [(immutable? h)
    (ann h (Immutable-HashTable Integer Integer))
    0]
   [else
    1]))
;; Type error

#lang typed/racket/base

(: g (-> (HashTable Integer Integer) Integer))
(define (g h)
  (cond
   [((inst immutable? Integer Integer) h)
    (ann h (Immutable-HashTable Integer Integer))
    0]
   [else
    1]))
```

pnwamk:
Quick thought: `HashtableTop` as it currently exists is no longer sensible with the changes in this PR. There should be a `MutableHashTableTop` and a `WeakHashTableTop`, and then we would define `HashtableTop` as `(U (ImmutableHashTable Any Any) MutableHashTableTop WeakHashTableTop`.

After that is worked out, a predicate could say it is not a mutable or weak hash table (by saying it is not even it's respective Top) without weird type variable/polymorphism issues.

samth:
That sounds like a problem with type argument inference, if an explicit
inst works.

pnwamk:
Another quick thought: if we're giving `immutable?` a weird type (i.e. specifically calling out Hashtables) when other data (e.g. vectors) can be immutable as well... is that a sign this is not the right approach? Should we be using a simple `Refinement`? Or an intersection of some sort?

samth:
I think the idea is we'll do the same thing for vectors and boxes.

pnwamk:
I guess I just mean to say my knee-jerk is that it's odd we're giving `immutable?` a type with the hashtable case baked into the predicate's latent properties.

samth:
That's because immutable? and the relationship between mutable and
immutable versions of the basic data structures are all bad designs.

pnwamk:
Voila!

```racket
;; type-rep.rkt

;; ...

(def-structural Immutable-HashTable ([key #:covariant] [value #:covariant])
  [#:mask mask:immutable-hash])

(def-type Mutable-HashTableTop ()
  [#:mask mask:mutable-hash]
  [#:singleton -Mutable-HashTableTop])
(def-structural Mutable-HashTable ([key #:invariant] [value #:invariant])
  [#:mask mask:mutable-hash])

(def-type Weak-HashTableTop ()
  [#:mask mask:weak-hash]
  [#:singleton -Weak-HashTableTop])
(def-structural Weak-HashTable ([key #:invariant] [value #:invariant])
  [#:mask mask:weak-hash])

;; ...

```

```racket
;; base-abbrev.rkt

(define/decl -HashTableTop (Un (make-Immutable-HashTable Any Any)
                                                       -Mutable-HashTableTop
                                                       -Weak-HashTableTop))
```

That should also simplify the type of things like `hash-weak?` (which looks pretty brutal right now).

And... it might be nice to make the `mask:*` names clearer -- they're rarely used, so we're not saving a lot of space by making them more opaque for future readers (and forgetful future me).

samth:
I agree with @pnwamk that adding `Top` variants of the types is a better solution here.

bennn:
Thank you both, I'll have an update this weekend

bennn:
The new type for `immutable?` is great. Doesn't need an `inst` and we'll be able to add sets and vectors later.

```
[immutable?
 (asym-pred Univ B
            (-PS (-and (-not-type 0 -Mutable-HashTableTop)
                       (-not-type 0 -Weak-HashTableTop))
                 (-not-type 0 (-Immutable-HT Univ Univ))))]
```

Technically, `(immutable? x) ==> #true` should also imply that `x` is not `#f` or `#t` or .... but I don't see an easy way to support that.

- - -

Two questions about tests:
- [ ] The file `pr390-variation-4.rkt` raises an exception on Travis about instantiating `racket/gui/base` twice in the same process. Is there an easy fix? Should I move the test to somewhere in `typed-racket-more`?
- [ ] I don't know how to test `type->contract` (see in-line note below). Any ideas?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -450,8 +465,12 @@
             (apply or/sc (append other-scs (map t->sc (nbits->base-types nbits)))))]
        [(? Union? t)
         (match (normalize-type t)
-          [(Union: (? Bottom?) elems) (apply or/sc (map t->sc elems))]
-          [(Union: base elems) (apply or/sc (t->sc base) (map t->sc elems))]
+          [(Union-all: elems)
+           (define-values [hash-elems other-elems] (partition hash/chaperone? elems))
+           (define maybe-hash/sc (hash-types->sc hash-elems))
+           (if maybe-hash/sc
+             (apply or/sc maybe-hash/sc (map t->sc other-elems))
```
bennn: This is the new behavior in `type->contract` that I'd like to test.

`(U #f (Immutable-HashTable String String) (Mutable-HashTable String String))` should make the contract `(or/c #f (hash/c (or string? string?) (or string? string?)))`


(Maybe this is the wrong approach? Maybe `type->contract` should be very simple, then have a `simplify-contract` function?)

pnwamk:
What if we just had a type for those things for which `immutable?` is (or can be) true:

```racket
(define/decl -Immutable (U -String
                           -ByteString
                           -VectorTop
                           (make-Immutable-HashTable Any Any)
                           -BoxTop))
```

This is a little conservative currently because some of our types are still coarse, but I think it would still be correct and would easily adapt as we added finer grained types for the other immutable/mutable variants.

Then we could just do this for the type of `immutable?`:

```racket
[immutable? (make-pred-ty -Immutable)]
```

Would that work?

AlexKnauth:
I would hope that `(immutable? x)` returning false on a `(U String OtherType)` would not imply that `x` was an `OtherType`...

pnwamk:
@AlexKnauth ah yes, thanks. The `-Immutable` type I suggested could be in the 'then' proposition, but the 'else' proposition would have to be something more customized (until perhaps `-Immutable` is exactly the values that produce `#t` for `immutable?`)

bennn:
Ok I can try that.

(I'm surprised, I thought a positive `(U A B)` filter would change the type to `(U A B)` on success ... so my HashTables would be `(U ImmutableHash String .....)`.)

samth:
A positive proposition (the term "filter" is dead!) intersects with the previous type of the variable.

typed-racket-lib/typed-racket/private/type-contract.rkt
samth: `type->contract` should indeed be simple, but the intelligence is in the static-contract data structures and their contract generation methods.

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -920,6 +946,17 @@
      (t:subtype -Boolean t)]
     [_ #f]))
 
+;; hash/chaperone? : Type -> Boolean
```
samth: What does this have to do with chaperones?

typed-racket-lib/typed-racket/typecheck/tc-literal.rkt
```diff
@@ -134,18 +134,21 @@
...
-               [vs (hash-map h (lambda (x y) (tc-literal y v)))])
-          (make-Hashtable
-            (check-below (apply Un ks) k)
-            (check-below (apply Un vs) v)))]
-       [_ (let* ([h (syntax-e #'i)]
-                 [ks (hash-map h (lambda (x y) (tc-literal x)))]
-                 [vs (hash-map h (lambda (x y) (tc-literal y)))])
-            (make-Hashtable (generalize (apply Un ks)) (generalize (apply Un vs))))])]
+     (let ([h (syntax-e #'i)])
+       ;; TODO assert that it is immutable?
```
samth: Syntax is always immutable.

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -682,6 +671,16 @@
                  #:break (not A))
         (subtype* A elem1 seq-t))]
      [_ (continue<: A t1 t2 obj)])]
+  [(case: Immutable-HashTable (Immutable-HashTable: key1 val1))
+   (match t2
+    [(Immutable-HashTable: key2 val2)
+     (subtype-seq A (subtype* key1 key2) (subtype* val1 val2))]
```
pnwamk: the rest of this file has a vertically-aligned style for the `subtype-seq` monad macro---might be nice to keep that style consistent... maybe it doesn't matter. I always associated the vertical style with it being like a `do` block a-lá Haskell.

pnwamk:
General question: has there been any discussion about if/how we might leverage these types for a stricter hash API (i.e. have `hash-set` only accept immutable hashes, etc) in some way/shape/form? Or have we tried changing it to see how much code it broke? and if that code was erroneous or our type checker was just too dumb to see the reasoning?

Just curious.

bennn:
The only discussion I remember was Sam saying "please keep backwards compatibility" --- which I totally agree with. I don't want to change the hash API for the same reason that I often use type `Real` when I'm thinking type `Index`.

I'd be interested in adding a `typed/racket/hash` module that would shadow identifiers from `#lang typed/racket`. 
... though that's not a good name, because the same convention wouldn't work for vectors.

I have not tested to see how much code would break with the new types.

samth:
This is looking good. Let's address improved APIs in future PRs.

samth:
I restarted the Travis build, but this is good to merge.

PR #564, opened 2017-06-27
Remove opaque monad type from typed/rackunit

jackfirth:
The "monad" struct in RackUnit's internals is never exposed to users and there's no requirement to use it with `fold-test-results`. The "seed" parameter in `fold-test-results` can be any value and is more accurately described by a parametric type. I'm not quite sure what the correct change to make it parametric is, but this change is necessary to remove the dependency on `rackunit/private/monad`. The monad implementation of `rackunit/text-ui` will hopefully go away soon.

### DIFF
--- a/typed-racket-more/info.rkt
+++ b/typed-racket-more/info.rkt
@@ -22,4 +22,4 @@
 
 (define pkg-authors '(samth stamourv))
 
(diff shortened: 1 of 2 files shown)


jeapostrophe:
LGTM

PR #565, opened 2017-06-27
patch: check duplicates before making an or/c for hash keys

bennn:
The contract for `(U (I-Hash k1 v1) (M-Hash k2 v2) (W-Hash k3 v3))`
 is now `(hash/c (or/c k1 k2 k3) (or/c v1 v2 v3))`
 ONLY WHEN the key and value types are distinct.
The contract should no longer include duplicate key or value types.

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -384,17 +384,22 @@
       (define (hash-types->sc hts)
         (if (or (null? hts) (null? (cdr hts)))
           #false ;; too few types, don't merge
-          (let-values ([(key-scs val-scs)
-                        (for/lists (ks vs)
-                                   ([ht (in-list hts)])
-                          (match ht
-                           [(or (Immutable-HashTable: k v)
-                                (Mutable-HashTable: k v)
-                                (Weak-HashTable: k v))
-                            (values (t->sc k) (t->sc v))]
-                           [_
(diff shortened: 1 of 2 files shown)


bennn:
Should fix http://drdr.racket-lang.org/41016/racket/share/pkgs/redex-test/redex/tests/run-tests.rkt

I wasn't able to reproduce that error with `raco test lw-test.rkt`, but the unit test in this PR seems morally the same.

pnwamk:
The issue is not simply duplicates, but overlapping types/contracts, right? Would it be better to just union the keys and values and turn those into contracts (since our union contract generation already worries about overlap)?

i.e. something like this
```racket
(define (hash-types->sc hts)
  (cond
    ;; too few types, don't merge
    [(or (null? hts) (null? (cdr hts))) #false]
    ;; otherwise union kts and vts for hash/c
    [else
     (define-values (kts vts)
       (for/lists (_1 _2) ([ht (in-list hts)])
         (match ht
           [(or (Immutable-HashTable: k v)
                (Mutable-HashTable: k v)
                (Weak-HashTable: k v))
            (values k v)])))
     (hash/sc (t->sc (apply Un kts))
              (t->sc (apply Un vts)))]))
```

samth:
Yes, that seems like something we need to worry about.

bennn:
That's a better implementation.

(I was looking to find what `Union` _did_, saw `remove-duplicates` and then decided on `member`. Didn't think to just make a union :/ )

PR #566, opened 2017-06-27
simplify arrows a little, less list allocation

pnwamk:


pnwamk:
On my machine this passes all the unit, integration, and math tests from the TR test suite. The TravisCI integration test fails because it breaks a htdp file which has some rather... intimate requires from TR's internals (and there's a PR to patch the specific part that it broke---maybe it would be nice to rewrite it using require/provide forms if possible?).

PR #568, opened 2017-06-28
Remove typed/rackunit references to check-around

jackfirth:
This is an internal implementation detail of rackunit. The `current-check-around` parameter is set to `check-around` by default, so there's no need to explicitly set it in test-begin.

### DIFF
--- a/typed-racket-more/info.rkt
+++ b/typed-racket-more/info.rkt
@@ -22,4 +22,4 @@
 
 (define pkg-authors '(samth stamourv))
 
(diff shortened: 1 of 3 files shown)


jackfirth:
Could someone with admin access restart the travis build? It timed out.

pnwamk:
this looks pretty minor/benign... @jackfirth is this good to squash+merge?

jackfirth:
Yup, all good.

Issue #563, opened 2017-06-24
Occurrence typing on polymorphism types yields error


yjqww6:
### What version of Racket are you using?
6.9
### What program did you run?
```
#lang typed/racket

(define-type P (∀ (x) (Option (x -> x))))

(define (g [a : P])
  (if a #t #f))
```
### If you got an error message, please include it here.
```
Poly*: arity mismatch;
 the expected number of arguments does not match the given number
  expected: 2 plus an optional argument with keyword #:original-names
  given: 3
  arguments...:
```

It looks like the error message comes from here https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/types/subtract.rkt#L43

PR #571, opened 2017-07-01
fix trivial bug in subtract

pnwamk:


pnwamk:
fixes https://github.com/racket/typed-racket/issues/563

PR #575, opened 2017-07-09
types for mutable and immutable vectors

### LINKED ISSUE #598: Unsound static contracts optimization
The static contracts optimizer converts `(or/sc _ ... any/sc _ ...)` to `any/sc`. This is only sound if the `_ ...` contracts are flat.

I'm planning to fix this in #575,
but posting the issue to get feedback on my "only sound if" claim above.

Tested on 6.7 and HEAD

### What program did you run?

```
#lang typed/racket/base

(module u racket/base
  (define (f b)
    (set-box! b "hello"))
  (provide f))

(define-type Maybe-Box (U #f (Boxof Integer)))

(require/typed 'u (f (-> Maybe-Box Void)))

(define b : Maybe-Box (box 4))
(f b)

(if (box? b)
  (+ 1 (unbox b))
  (error 'deadcode))
```

Contract violation at the `set-box!` call.

```
+: contract violation
  expected: number?
  given: "hello"
  argument position: 2nd
  other arguments...:
   1
```

### LINKED ISSUE #743: infer: very slow constraint generation / solving

6.90.0.30

### What program did you ~run~ try to compile?

```
#lang typed/racket/base

;; Adapted from the 'math' library.

(require racket/fixnum)

(define-predicate exact-zero? Zero)

(: vector-ref! (All (A B) ((Vectorof (U A B)) Integer (-> A) (Any -> Boolean : B) -> A)))
(define (vector-ref! vs i thnk nothing?)
  (define n (vector-length vs))
  (cond [(and (fixnum? i) (i . fx>= . 0) (i . fx< . n))
         (define v (vector-ref vs i))
         (if (nothing? v)
             (let ([v  (thnk)])
               (vector-set! vs i v)
               v)
             v)]
        [else
         (raise-range-error 'vector-ref! "Vector" "" i vs 0 n #f)]))

(define-type (Containerof T) (U (Listof T) (Vectorof T)))

(: container-ref!
   (All (A B) ((Containerof (U A B)) Integer (-> A) (Any -> Boolean : B) -> A)))
(define (container-ref! c i t p)
  (if (vector? c)
    (vector-ref! c i t p)
    (raise-user-error 'cref!)))

(define num-global-bs 2)
(define local-bs : (Containerof Exact-Rational)
  (make-vector num-global-bs 0))

(: bs-ref! (Integer (-> Exact-Rational) -> Exact-Rational))
(define (bs-ref! n thnk)
  (container-ref! local-bs (- n num-global-bs) thnk exact-zero?))
```

This takes at least 5 minutes to compile. (I'm waiting for it to finish. EDIT: killed it, too slow & too much memory to wait.)

### Test case

Here's a small test case for `typed-racket-test/unit-tests/infer-tests.rkt`:

```
 (infer-t (Un (-pair -Rat (-val null))
              (-vec -Rat))
          (Un (-pair (Un (-v a) (-v b)) (-val null))
              (-vec (Un (-v a) (-v b))))
          #:vars '(a b))
```

bennn:
New types:
- Immutable-Vector (covariant)
- Mutable-Vector (invariant)
- Immutable-Vectorof (covariant)
- Mutable-Vectorof (invariant)
- Mutable-VectorTop (similar to Mutable-HashTableTop)

The old vector types are now unions of these new types.

Also:
- more specific _return_ types for some built-in functions (e.g. `vector-append`)
- `immutable?` distinguishes mutable & immutable vectors
- the contract for a `Vector` or `Vectorof` should be the same as before,
  even though these types are now unions

Current TODOs:
- [x] travis
- [x] think about adding more tests
- [x] clean up the "union deduplicating" in `type-contract.rkt`, it's right now an awkward implementation of a simple idea
- [X] "oops" below, revert the condition and make sure tests pass. EDIT: reverting the condition breaks some `typecheck-tests`, not sure why.
- [X] "oops 2", add a test for this bug & fix it
- [X] "oops 3" ... try simplifying the code & see if tests pass?
- [x] make sure HVecs are subtypes of VectorTop
- [x] make sure VectorTop generates a flat contract (try compiling some packages)
  - found errors in ~`plot`~ (fixed) and `scv`
- [x] sync with fixed https://github.com/racket/racket/issues/1745
- [x] should other `intersect`s in `tc-literal.rkt` be `resolve`s?
  - removed all these intersects; expected type isn't relevant when the type is immutable (because immutable types have subtyping)
- [X] change `type-contract-generalize` interface
  - incorrect to use just one `type->contract` continuation because ignores typed-side/untyped-side distinction
  - and try to make that function overall simpler
  - and should move this code to the static-contracts optimizer
- [x] fix #598 , move tests for `type-contract-generalize` to static-contracts-optimize
- [x] add history annotations
- [X] do we really need travis_wait? (why is typechecking slower) --- NO, do not accept the travis_wait, make this PR faster instead
- [x] resolve #743 (EDIT: resolved the part that matters for this PR)


typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-hetero.rkt
```diff
@@ -102,41 +124,78 @@
...
+
+;; The cases for `(vector-immutable e ...)` and `(vector e ...)` are similar
+;;  but use different match expanders, so we use a macro to abstract over the similar parts
+;; The `#:with` clause below contains the differences
+(define-values [tc-app-immutable-hetero-vector tc-app-mutable-hetero-vector]
+  (let ()
+    (define-syntax (make-tc-app-hetero-vector stx)
+      (syntax-parse stx
+       [(_ is-immutable?:boolean)
+        #:with (make-HV match-V match-HV maskV tc-expr/t/maybe-generalize)
```
bennn: Should I rename these to make them stand out more? Like `?makeHV`, etc.

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
```diff
@@ -387,22 +387,44 @@
...
-       [(HeterogeneousVector: ts)
-        (make-HeterogeneousVector
-         (for/list ([x (in-list xs)]
-                    [t (in-list/rest ts #f)])
-           (cond-check-below (find-stx-type x t) t)))]
-       [_ (make-HeterogeneousVector (for/list ([x (in-list xs)])
-                                      (generalize (find-stx-type x #f))))])]
+    [(and (vector xs ...) v)
+     (cond
+      [#true ;(immutable? v)
```
bennn: oops

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
```diff
@@ -426,7 +448,7 @@
         [(Mutable-HashTable: k v)
          (value->HT/find-stx-type h -Mutable-HT k v)]
         [_
-         (value->HT/find-stx-type h -HT)])])]
+         (value->HT/find-stx-type h -HT)])])] ;; TODO should be mutable
```
bennn: oops 2

typed-racket-lib/typed-racket/types/overlap.rkt
```diff
@@ -129,7 +129,7 @@
     #:no-order
     #f]
    [((Val-able: (not (? hash?)))
-     (or (? Mutable-HashTable?)
+     (or (? Mutable-HashTable?) ;; TODO simplify?
```
bennn: oops 3

typed-racket-lib/typed-racket/rep/free-variance.rkt
```diff
@@ -102,6 +102,12 @@
      (remove-frees (flip-variances inner) name)]))
 
 
+(define (make-covariant frees)
+  (combined-frees
```
pnwamk: (Unless I'm missing something) I'm pretty sure this function `make-covariant` should be deleted and anywhere that uses it should instead _just_ be calculating the free variables of the respective fields w/o modifying the variance.

typed-racket-lib/typed-racket/types/abbrev.rkt
```diff
@@ -34,7 +34,14 @@
...
 (define -set make-Set)
+(define -Mutable-vec make-Mutable-Vector)
+(define -Immutable-vec make-Immutable-Vector)
+(define (make-Vector a) (Un (-Mutable-vec a) (-Immutable-vec a)))
 (define -vec make-Vector)
+(define (make-HeterogeneousVector ts) (Un (make-Immutable-HeterogeneousVector ts)
+                                          (make-Mutable-HeterogeneousVector ts)))
+(define (-Immutable-vec* . ts) (make-Immutable-HeterogeneousVector ts))
+(define (-Mutable-vec* . ts) (make-Mutable-HeterogeneousVector ts))
 (define (-vec* . ts) (make-HeterogeneousVector ts))
```
pnwamk: Minor nitpick (sorry) - could the names for these abbreviated constructors (e.g. `-Immutable-vec`) be a little more consistent with the other abbreviated constructors? It looks like previously we've used either all lower case or camel case in the past (i.e. `-immutable-vec` or `-Immutable-Vec`), but I don't see any mixed-fix camel case for these sorts of abbreviations. Again, sorry for the minor nitpick =\

typed-racket-lib/typed-racket/types/overlap.rkt
```diff
@@ -113,7 +113,7 @@
     #f]
    ;; these are single-valued sequences
    [((Sequence: (list _ _ _ ...))
-     (or (? Pair?) (? Vector?) (? VectorTop?)))
+     (or (? Pair?) (Vector: _) (VectorTop:)))
     #:no-order
```
pnwamk: Is there a reason to favor match patters w/ predicates vs trivial match expanders? just curious ;-) (I'm probably not consistent myself with this choice, lol)

pnwamk:
Looks good at a glance! (I'll give it a more thorough look soon)

typed-racket-lib/typed-racket/rep/free-variance.rkt
bennn: The one use is [here](https://github.com/racket/typed-racket/pull/575/files#diff-78b0804be48e78b95cf9c2b2327596c0R437).

I don't know how to change it.

pnwamk: We just want delete the application of `make-covariant` so it looks like this:

```racket
[#:frees (f) (combine-frees (map f elems))]
```

similar to how the frees for [Sequence](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/rep/type-rep.rkt#L1144) are calculated.

bennn:
Problem: I tried giving the result of `for/vector` the type `(Mutable-Vector ....)`, but the implementation of `for/vector` uses a `set!` and so generalizes my `Mutable-Vector` into a `Vector`.

Is there a way to keep the `set!`, but prevent it from generalizing? ([Here](https://github.com/racket/typed-racket/pull/575/files#diff-72cf4e4479067e30d7f7287eeb580300R821) is the `set!`)

samth:
To fix the issue with `set!`, you need to add an annotation to those variables at the point of definition.

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
```diff
@@ -387,22 +387,44 @@
...
-                   (find-stx-type x t)))
-          t))]
-       [(HeterogeneousVector: ts)
-        (make-HeterogeneousVector
-         (for/list ([x (in-list xs)]
-                    [t (in-list/rest ts #f)])
-           (cond-check-below (find-stx-type x t) t)))]
-       [_ (make-HeterogeneousVector (for/list ([x (in-list xs)])
-                                      (generalize (find-stx-type x #f))))])]
+    [(and (vector xs ...) v)
```
bennn: Something strange is happening
- if I have phase 0 code that puts a _mutable_ vector into a syntax object, `syntax-e` returns an _immutable_ vector
- if I have phase 1 code that puts an _immutable_ vector into a syntax object, `syntax-e` returns a _mutable_  vector

Example:

```
#lang racket/base
(require rackunit (for-syntax rackunit racket/base))

(check-false ;; FAILS
 (immutable? (syntax-e #`#,(vector)))
 "(syntax-e (syntax (vector))) made immutable vector")

(check-true ;; PASSES
 (immutable? (syntax-e #`#,(vector-immutable)))
 "(syntax-e (syntax (vector-immutable))) made mutable vector")

(define-syntax (mvec stx)
  (define v (vector))
  (check-false ;; PASSES
   (immutable? v)
   "(vector) made immutable vector")
  #`#'#,v)

(check-false ;; PASSES
 (immutable? (syntax-e (mvec)))
 "(syntax-e (syntax (vector))) made immutable vector")

(define-syntax (ivec stx)
  (define v (vector-immutable))
  (check-true ;; PASSES
   (immutable? v)
   "(vector-immutable) made mutable vector")
  #`#'#,v)

(check-true ;; FAILS
 (immutable? (syntax-e (ivec)))
 "(syntax-e (syntax (vector-immutable))) made mutable vector")
```

Output:
```
--------------------
FAILURE
name:       check-false
location:   pr575-variation-3.rkt:4:0
params:     '(#t)
message:    "(syntax-e (syntax (vector))) made immutable vector"
--------------------
--------------------
FAILURE
name:       check-true
location:   pr575-variation-3.rkt:30:0
params:     '(#f)
message:    "(syntax-e (syntax (vector-immutable))) made mutable vector"
--------------------
```

I think this is a problem with Racket, so I've changed a few `typecheck-tests` to match the current (unexpected)  behavior to see if anything else goes wrong on Travis.

samth:
I recommend not using quasiquote and unquote if your goal is to test
something simpler than that.

bennn:
I don't know a simpler way to put a mutable vector into a syntax object.

For immutable vectors:
```
(immutable? (syntax-e #'#(1 2)))
;; ==> #f, seems wrong
```

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
bennn: Sam points out: `datum->syntax` is documented to turn mutable vectors into immutable vectors. So that explains the behavior in `typecheck-tests.rkt`

bennn:
@pnwamk  the current travis failure is from `with-integer-linear-arith2.rkt`, basically:

```
(define-type (V8 A) (Refine [v : (Vectorof A)] (= 8 (vector-length v))))
(ann (vector 0 1 2 3 4 5 6 7) (V8 Any))
```

should I look at `types/subtype.rkt` to fix this, or do you think the problem might be somewhere else?


I think it's probably subtyping because this code works:
```
(define-type (V8 A) (Refine [v : (Vectorof A)] (= 8 (vector-length v))))
(define-type (MV8 A) (Refine [v : (Mutable-Vectorof A)] (= 8 (vector-length v))))
(ann (ann (vector 0 1 2 3 4 5 6 7) (MV8 Any)) (V8 Any)) 
```
but I figure I might as well ask, because I'm not planning to work on this until later tonight.

pnwamk:
@bennn Played around a little with the issue, this seemed to fix it:

https://github.com/pnwamk/typed-racket/commit/d1f671588e48a6da5ddaff56196768441289bbc6

I haven't tested it more broadly, though.

pnwamk:
It's too late over here for me to think this through all the way at the moment, but I wonder if we need a similar match expander (that matches on the unions of the various kinds) for hash tables as well.

bennn:
Thanks for looking.

I don't know about the match expander. I think the I/M distinction is useful and don't want to start hiding it internally.

I'll look to see where the match expander helped fix this problem. (Hoping its just 1 or 2 places that need to special-case.)

bennn:
Ok fixed that test. There was a bug in `tc-literal.rkt`, I was intersecting the expected type with the wrong vector type.

(Next up: the optimizer, which I think definitely needs to recognize `(U IVec MVec)`)

pnwamk:
One small (possibly entirely unnecessary) note -- if we're going to bother having match expanders like `Vector:` (and I think they are occasionally useful), I think they need to match on the union case. i.e. if `Vector:` intuitively means as long as its "either kind of vector", I think we'll be accidentally introducing bugs if it omits the union case.

typed-racket-test/succeed/with-integer-linear-arith2.rkt
```diff
@@ -16,7 +16,7 @@
 (ann (vector 0 1 2 3 4 5 6 7) (V8 Any))
 (ann (vector 0 1 2 3 4 5 6 7) (V8 Byte))
 (ann '#(0 1 2 3 4 5 6 7) (V8 Any))
-(ann '#(0 1 2 3 4 5 6 7) (V8 Byte))
+(ann '#(0 1 2 3 4 5 6 7) (Refine [v : (Immutable-Vector Byte Byte Byte Byte Byte Byte Byte Byte)] (= 8 (vector-length v))))
```
bennn: @pnwamk without this change, I get an error like this:

```
foo.rkt:7:5: Type Checker: type mismatch
  expected: (Refine
             (d : (U (Immutable-Vectorof Byte) (Mutable-Vectorof Byte)))
             (= 8 (vector-length d)))
  given: (Refine
          (e
           :
           (Immutable-Vector
            Integer
            Integer
            Integer
            Integer
            Integer
            Integer
            Integer
            Integer))
          (= (vector-length e) 8))
  in: (quote #(0 1 2 3 4 5 6 7))
```

Any idea how I could fix the error without changing this test?

IMO the "given" should be a subtype of the "expected" ... but I don't know how to change Typed Racket's opinion.

bennn: and with this change, the travis tests all pass (1/3 travis checks error for "no output received in last 10 minutes")

pnwamk: Maybe I'm reading this wrong... but for the "given" to be a subtype of the "expected", wouldn't `Integer` have to be a subtype of `Byte`?

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-hetero.rkt
```diff
@@ -102,41 +124,82 @@
...
+                (match u-ts
+                  [(list t0) (tc/app #`(#%plain-app . #,form) (ret t0))]
+                  [_ (continue)])]
+               ;; if mutable, generalize element types
+               [(or #f (tc-any-results: _) (tc-result1: _))
+                (define tc-hv-elem ?tc-expr/t/maybe-generalize)
+                (ret (?make-HV
+                       (for/list ((e (in-syntax #'args)))
+                         (tc-hv-elem e))))]
+               [_ (ret Err)])])))]))
```
pnwamk: Why do we return an Error type here?? (I know you didn't add that, it was there before -- but usually we only introduce Err if we've registered at least one type checker error -- am I missing something here?)

typed-racket-test/succeed/with-integer-linear-arith2.rkt
bennn: ooops! you are reading it right, thank you for the hint

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-hetero.rkt
bennn: Maybe this case is impossible, or only happens when there's been an error? I don't know ... what does it mean if an `expected` type doesn't match the pattern `(or #f (tc-any-results: _) (tc-result1: _))` ?

pnwamk: My guess is that would be an internal error... as in `(int-err "unreachable code! aaah!")`

typed-racket-test/succeed/with-integer-linear-arith2.rkt
pnwamk: https://github.com/pnwamk/typed-racket/commit/fe0704967866148e22b4be3a5385f72a7585d630?w=1

pnwamk: I _think_ that fixes it -- it just uses some of the logic from tc-app-hetero for vector literals (and it gets rid of an unnecessary continuation usage in the match! yay!)

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-hetero.rkt
pnwamk: in fact we should just remove that last clause -- it would be better to let match error, or just assume it was `(or #f (tc-any-results: _) (tc-result1: _))` and type check the vector than to return the type `Error` -- I have wondered if we shouldn't hide the constructor for `Error` inside of a function that makes sure there has been at least one type error registered to ensure we're not dangerously letting it leak around and screwing things up.

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-hetero.rkt
```diff
@@ -102,41 +125,84 @@
...
+                (match u-ts
+                  [(list t0) (tc/app #`(#%plain-app . #,form) (ret t0))]
+                  [_ (continue)])]
+               ;; if mutable, generalize element types
+               [(or #f (tc-any-results: _) (tc-result1: _))
+                (define tc-hv-elem ?tc-expr/t/maybe-generalize)
+                (ret (?make-HV
+                       (for/list ((e (in-syntax #'args)))
+                         (tc-hv-elem e))))]
+               [_ (int-err "tc-app-hetero:~a: no clause matched for type ~a" '?fn expected)])])))]))
```
bennn: @pnwamk I like this way of removing `Err`, but if you have a preference between:

- using `int-err`
- removing this clause
- removing this clause & removing the changing the previous pattern from `(or ....)` to `_`

Let me know & I'll make the change (you know the type checker much better than I do!)

pnwamk: the third option is probably the simplest -- thanks for removing `Err` while you're in there!

bennn:
short question: `(∩ (Immutable-Vectorof Any) (Sequenceof Real))` should be `(Immutable-Vectorof Real)`, right?

- - - 

story: `(ann #(1/2 1/2) (Sequenceof Real))` no longer typechecks, because `tc-literal.rkt`:
- intersects the expected type with `VectorTop`
- giving the expected type `(U (∩ (IVof Any) (Seqof R)) (∩ MVTop (Seqof R))`
- which is a union where exactly 1 type `t` is such that `(mask t) = mask:immutable-vector`
- so `tc-literal` loops with the new expected type `(∩ (IVof Any) (Seqof R))`
- and intersects that type with `VectorTop`
- giving the expected type `(Immutable-Vectorof Any)`
- so the program no longer type checks

I think either the intersection or mask-checking is wrong.

samth:
That mask check is definitely right.

pnwamk:
would it be better to use the non-additive `restrict` instead of `intersect` there? (i.e. `restrict` with `VectorTop` instead of `intersect`)

pnwamk:
what is `raw-mask`? I don't see it on quick repo searching and C-f of the PR diff.

bennn:
I got `raw-mask` from the [definition of `mask`](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/rep/rep-utils.rkt#L195). But it's not in this PR.

Using `restrict` is a good idea --- after that change, the `(ann #(1/2 ....))` and all typecheck-tests pass.

pnwamk:
if that is the right solution, it might mean the other uses of intersect in that file should be restrict as well... ?

bennn:
Maybe ... I'll try to adapt the Vector programs into something that points out a flaw in other intersects. But if I don't find any programs that fail, I'll leave the `intersect` alone in this PR.

bennn:
(dear diary) the optimizer tests are failing because the vector in `(vector-length '#(1 2 3))` has type `(Immutable-Vectorof Any)`

this happens because `vector-length` is expecting a `VectorTop`, and `tc-literal` uses the expected type to infer `(Immutable-Vectorof Any)`

if I change the type of `vector-length` to be a `case->`, then the expected type in `tc-literal` is `#f` and the vector `'#(1 2 3)` gets a more specific type (and the optimizer tests should pass)

... I guess the next thing to do is figure out why the optimizer tests pass on `master`, then figure out how to fix `tc-literal` in this PR

bennn:
Ok, it worked before because:
- if you intersect with `VectorTop` and have `(Vector t)` left, then the programmer's expecting a maybe-mutable vector

So my match pattern was wrong --- if you're expecting mutable then keep the programmer's type, otherwise infer a heterogeneous vector type and let subtyping do its job.

bennn:
Status update:

this program type checks

```
#lang typed/racket/base

(define-new-subtype VB (make-VB (Mutable-Vectorof Boolean)))

(define x : VB (make-VB (vector #f #f #t)))

(for ([b : Boolean (in-vector x)]) b)
```

this program does not type check

```
#lang typed/racket/base

(define-new-subtype VB (make-VB (Vectorof Boolean)))

(define x : VB (make-VB (vector #f #f #t)))

(for ([b : Boolean (in-vector x)]) b)

;;; TYPE ERROR
;;;in-vector-subtype.rkt:7:0: Type Checker: Polymorphic function `unsafe-vector-ref' could not be applied to arguments:
;;;Types: VectorTop Integer #f * -> Any
;;;Arguments: VB Fixnum #f *
;;;Expected result: Boolean
;;;
;;;  in: (for ((b : Boolean (in-vector x))) b)
```


I'm not sure why not; trying to find out.

bennn:
Ok I think this is in good shape.


Note, I didn't change the pair and hashtable cases in `tc-literal.rkt`, for two reasons:
  1. I couldn't find a problem with the uses of `intersect` for pairs and hashtables (so no need to change to `restrict`)
  2. I thought it'd be safe to ignore the expected type in those cases, since without annotation TR will infer a covariant type. But when I made this change, a few tests related to `for/hash` in `typecheck-tests.rkt` failed, so I just reverted.

samth:
What's currently failing in the test suite here?

bennn:
Not sure ... I'm glad the current error isn't a timeout, but I'm able to run what I think is the failing command on my machine with no problems:

```
$ PLT_TR_CONTRACTS=1 raco setup typed typed-racket typed-scheme
$ PLT_TR_CONTRACTS=1 racket -l typed-racket-test -- --unit
```

I'll look again tonight

bennn:
Still not sure, I guess it's something at the end of `raco setup`.

It seemed like travis was getting killed while setting up `typed-racket-more/**/web-server`, but I deleted that folder, re-ran travis, and had the same trouble at the end of the setup.

typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt
```diff
@@ -156,8 +157,12 @@
   ((listof/sc (#:covariant)) listof #:flat)
   ((cons/sc (#:covariant) (#:covariant)) cons/c #:flat)
   ((set/sc (#:covariant #:chaperone)) set/c #:flat)
-  ((vector/sc . (#:invariant)) vector/c #:chaperone)
+  ((immutable-vectorof/sc (#:covariant)) immutable-vectorof/c #:flat)
+  ((mutable-vectorof/sc (#:invariant)) mutable-vectorof/c #:chaperone)
   ((vectorof/sc (#:invariant)) vectorof #:chaperone)
+  ((immutable-vector/sc . (#:covariant)) immutable-vector/c #:chaperone) ;; immutable-vector/c is not flat, see `utils/vector-contract.rkt`
```
pnwamk: but it could be flat, right? if all its elements are flat?

pnwamk: (maybe there's not a good way to say 'maybe-flat' currently though...?

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-hetero.rkt
```diff
@@ -90,6 +108,11 @@
     (match (single-value #'vec)
       [(tc-result1: (and vec-t (app resolve (Is-a: (HeterogeneousVector: es)))))
        (tc/hetero-ref #'index es vec-t "vector")]
+      [(tc-result1: (and vec-t (app resolve (Union: _ ts)))) (=> continue)
+       (define es (hetero-vecs->elems ts))
```
pnwamk: `(tc-result1: (and vec-t (app resolve (Union: _ (app hetero-vecs->elems (? list? es)))))`

would get rid of the need for the `(=> continue)`

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-hetero.rkt
```diff
@@ -102,41 +125,81 @@
     (match (single-value #'v)
       [(tc-result1: (and vec-t (app resolve (Is-a: (HeterogeneousVector: es)))))
        (tc/hetero-set! #'index es #'val vec-t "vector")]
+      [(tc-result1: (and vec-t (app resolve (Union: _ ts)))) (=> continue)
+       (define es (hetero-vecs->elems ts))
```
pnwamk: Same as above w.r.t. the continue

typed-racket-lib/typed-racket/utils/vector-contract.rkt
```diff
@@ -0,0 +1,12 @@
+#lang racket/base
+
+(require racket/contract)
+(provide immutable-vector/c immutable-vectorof/c mutable-vector/c mutable-vectorof/c)
+
+;; TODO if the element contracts are flat, we can pass the `#:flat` option
+;;  to `vector/c` and `vectorof` to make these combinators flat.
```
pnwamk: would it not work to just do this TODO now -- i.e. throw in an `ormap` with `flat-contract?` and and use `#:flat? #t` if it's true?

typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt
bennn: Right, I didn't see a good way.

bennn:
Squashed the commits.
Tests pass after changing a `raco setup` in the travis script to:

```
- travis_wait raco setup typed typed-racket typed-racket-test typed-scheme
```

I guess we could also split the `raco setup` into 2 calls.

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -954,17 +950,98 @@
...
+    (define (types->contracts q ts t->sc)
+      (match ts
+       ['()
+        '()]
+       [(cons t ts)
+        (define (=t t+)
+          (q t t+))
+        (define-values [qts other] (partition =t ts))
+        (cons ((q t->sc) (cons t qts)) (types->contracts q other t->sc))]))
+    (λ (ts t->sc)
```
bennn: Oops, having just one `t->sc`continuation is a design bug.
If you wanted to merge two arrow contracts, `(-> A B)` and `(-> C D)`, you'd want something like `(-> (t->sc/neg (merge A C)) (t->sc (merge B D)))`.

I'll change / specialize `type-contract-generalize` later tonight.

bennn:
Synced with master & tests pass.
Ok to merge?

bennn:
Synced again, after #612

bennn:
May I merge this?

samth:
Does the Travis change need to be merged? Is there a reason this makes things worse?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -978,17 +969,113 @@
...
+;;   (define ctc
+;;     (parametric->/c [A]
+;;       (let ([ctc0 (vectorof A #:immutable #false)]
+;;             [ctc1 (vectorof A #:immutable #true)])
+;;         (-> (or/c ctc0 ctc1) any))))
+;;   (define/contract (f x) ctc (void))
+;;   (f (vector 1)) ;; error: none of the branches of the or/c matched
+;; ```
+;;
+;; 2017-12-02: It may be possible to remove this function in the future
```
samth: Can you say when it would be possible?

bennn:
Without the travis change, the build fails.

That's all I know, but I'll try looking again this weekend.

samth:
Does the build take longer on Travis with this change?

bennn:
Yes the build seems to be taking longer.

On my laptop, running `gtime raco setup -v typed typed-racket typed-racket-test typed-scheme` with PLT_TR_CONTRACTS set (in `utils.rkt`) takes:
- 11 min 35 seconds on master
- 18 min 20 seconds on this branch

I'm not sure how to debug this :/

samth:
Can you try with contracts off?

bennn:
Without contracts: 8min 49sec on master, 9min 16sec on this branch.

samth:
That's not nearly as bad, but I'd still like more data. @pnwamk, what is the best way for Ben to run your benchmarks?

bennn:
If these are the benchmarks in https://github.com/pnwamk/tr-performance , I know how to run those & can post results tonight.

samth:
Yes, that's what I mean.

bennn:
Here's the results. I'll try to speed up `fsm` and `dungeon` this week.

(Key: `I` = this pull request, `M` = master, `Y` = with contracts, `N` = without contracts)

 program                     | I Y    | M Y    | I N    | M N    |
------------------------- | -------- | -------- | -------- | --------
 schml-specify-rep       | 20.80s | 19.12s | 14.98s | 14.10s 
 schml-interp-casts-help | 45.01s | 42.79s | 30.14s | 31.25s 
 parser                  |  4.32s |  4.30s |  3.60s |  3.55s 
 old-metrics             |  5.23s |  5.24s |  4.25s |  4.16s 
 new-metrics             |  8.50s |  8.38s |  5.65s |  5.63s 
 math-flonum             |  7.94s |  7.91s |  6.07s |  5.82s 
 fsm                     | 11.17s |  9.73s |  8.16s |  7.99s 
 forth                   | 10.18s |  9.49s |  7.86s |  7.83s 
 dungeon                 | 36.69s | 17.37s | 26.79s | 14.09s 
 acquire                 | 36.24s | 35.40s | 26.52s | 27.21s

pnwamk:
@bennn just to clarify, in your table the "with contracts" label is referring to the internal dev contracts that document TR. And we're just timing that as well so our internal dev "with contracts" builds are still reasonable? Is that right?

bennn:
Yes that's right.  Here's the change I made to `typed-racket/utils/utils.rkt` in the "with contracts" versions:

```
(define-for-syntax enable-contracts? (and #true #;(getenv "PLT_TR_CONTRACTS") #t))
```

ps. so far today I haven't been able to reproduce the 10-second slowdown on dungeon. On my laptop it's a little slower (1-5 sec.), but with high variance.

bennn:
I think this is finally ready

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -227,12 +230,36 @@
...
+(def-structural Mutable-Vector ([elem #:invariant])
+  [#:mask mask:mutable-vector])
+
+(define-match-expander Vector:
+  (lambda (stx)
+    (syntax-parse stx
+     [(_ elem-pat)
+      #'(or (Immutable-Vector: elem-pat)
+            (Mutable-Vector: elem-pat)
+            ;; Cannot use `list-no-order`
```
samth: You could use `list-no-order` with an explicit equality check.

typed-racket-lib/typed-racket/types/base-abbrev.rkt
```diff
@@ -126,6 +126,19 @@
...
+                            -Mutable-VectorTop))
+
+(define-match-expander VectorTop:
+  (lambda (stx)
+    (syntax-parse stx
+     [(_)
+      #'(? VectorTop?)])))
+
+(define (VectorTop? x)
+  (or (Immutable-Vector? x)
```
samth: This doesn't seem like the right predicate.

samth:
Just a couple small comments, then I think we're good to go!

typed-racket-lib/typed-racket/rep/type-rep.rkt
bennn: Like this, or is there a simpler way?
```
(and (Union-all: (list-no-order (.... elem-pat) (.... elem-pat2)))
     (? (lambda (_) (equal? elem-pat elem-pat2))))
```

samth:
The simpler way would use #:when.

pnwamk:
@samth aren't those match-expander definitions? They can't use `#:when`, they have to just expand into match patterns, right? I think the `or` method we had is right.

samth:
Yes, I agree that the current code is clearer.

bennn:
Okay, so I'll: keep the current code, re-word the comment to mention equality checks, and squash+merge?

samth:
Sounds good.

PR #577, opened 2017-07-09
attempt to fix float-complex magnitude optimization.

florence:
The current optimization missed two things that the runtime implementation does.

1. It does not check if the imag-part is zero, causing this program to behave differently in TR vs Racket:

```
#lang racket
(module untyped racket
  (printf "untyped: ~a\n"
          (magnitude 0.0+0.0i)))


(module typed typed/racket
  (printf "typed: ~a\n"
          (magnitude 0.0+0.0i)))
```

produces :
```
untyped: 0.0
typed: +nan.0
```

2. It does not switch the order of `l` and `r` based on which one is bigger. this program to behave differently in TR vs Racket:

```
#lang racket

(module untyped racket
  (printf "untyped: ~a\n"
          (magnitude 1.6496437860480294e+295+1.7259537815112588e-137i)))


(module typed typed/racket
  (printf "typed: ~a\n"
          (magnitude 1.6496437860480294e+295+1.7259537815112588e-137i)))

(require 'untyped)
(require 'typed)
```

produces:

```
untyped: 1.6496437860480294e+295
typed: +inf.0
```

This PR fixes these programs. However I'm not sure how to add them as tests cases....

### DIFF
--- a/typed-racket-lib/typed-racket/optimizer/float-complex.rkt
+++ b/typed-racket-lib/typed-racket/optimizer/float-complex.rkt
@@ -570,11 +570,18 @@
            #`(let*-values (c.bindings ...)
                ;; reuses the algorithm used by the Racket runtime
                (let*-values ([(r) (unsafe-flabs c.real-binding)]
-                             [(i) (unsafe-flabs c.imag-binding)]
-                             [(q) (unsafe-fl/ r i)])
-                 (unsafe-fl* i
-                             (unsafe-flsqrt (unsafe-fl+ 1.0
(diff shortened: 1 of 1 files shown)


florence:
The relevant part of the runtime to compare to is: https://github.com/racket/racket/blob/master/racket/src/racket/src/number.c#L3852

samth:
I think @stamourv probably knows where to put the tests.

samth:
We should merge this to the release.

stamourv:
*Just* sent you an email about this. ;)

stamourv:
Merged.

PR #582, opened 2017-07-13
backwards-compatibility: make HashTableTop generate a flat contract

### LINKED ISSUE #581: Changes to hash table types break `wrap`
```
raco setup:     compiling: <pkgs>/wrap/wrap/aws/dynamodb/parse.rkt
raco setup: error: during making for <pkgs>/wrap/wrap/aws/swf
raco setup:   /home/samth/tmp/wrap/wrap/aws/dynamodb/parse.rkt:78:10: Type Checker: Type HashTableTop could not be converted to a contract: required a flat contract but generated a chaperone contract
raco setup:     in: json
```

See error here: https://plt.eecs.northwestern.edu/pkg-build/server/built/fail/wrap.txt

bennn:
Changing `HashTableTop` from a singleton to the union:

```
  (U (Immutable-HashTable Any Any) MutableHashTable WeakHashTable)
```

is a backwards compatibility issue because the type `Any` requires a chaperone,
therefore `HashTableTop` requires a chaperone.

This commit adds a case to make sure `HashTableTop` generates a flat contract.

Closes #581

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -464,6 +464,13 @@
             (apply or/sc (append other-scs (map t->sc (nbits->base-types nbits)))))]
        [(? Union? t)
         (match (normalize-type t)
+          [(HashTableTop:)
+           ;; NOTE: this is a special case to make `HashTableTop` produce a flat contract.
+           ;; Without this case:
+           ;; - `HashTableTop` would make a chaperone contract
+           ;; - because `HashTableTop` is a union containing `(Immutable-HashTable Any Any)`
+           ;; - and `Any` makes a chaperone contract
+           hash?/sc]
           [(Union-all: elems)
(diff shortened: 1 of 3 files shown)


typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -464,6 +464,13 @@
             (apply or/sc (append other-scs (map t->sc (nbits->base-types nbits)))))]
        [(? Union? t)
         (match (normalize-type t)
+          [(HashTableTop:)
+           ;; NOTE: this is a special case to make `HashTableTop` produce a flat contract.
```
samth: This is just a strictly better contract, right?

bennn: Strictly better for the type HashTableTop, yes.

I don't think it's better for something that could be an `(ImmutableHashTable Any Any)`, but then again it seems like the contract for the `Any` type tends to get in people's way.

PR #584, opened 2017-07-16
WIP - dependent function types

pnwamk:
This PR adds dependent function types to Typed Racket. Specifically, it adds the following:

```
* new type DFun (see DFun def in type-rep.rkt)
  + represents a dependent function (i.e. arguments
    depend on one another)
  + syntax is like ->i, but w/o the 'i', e.g.
    (-> ([v : (Vec A)] 
         [i : (v) Nat #:where (< i (vec-len v))])
         A)
    order of dependencies doesn't matter, as long as there's
    no cycle in the dependencies
  + contracts for DFun generate ->i contracts (obviously)

* tc-expr/check can assign existential objects
  + sometimes we want a local/existential object for a value
    (like when applying a dependent function), tc-expr/check
    can assign one (notably this is used by tc-dep-fun-arg)

* #:with-logical-refinements flag turns on more reasoning
  namely existential objects for functions w/ dependent types
  and linear integer information for integer primitives (see below)


* integer arith prims get propositions w/ #:with-logical-refinements
  I ran into a bit of chicken and egg problem since our theory knows
  about certain integer operations (e.g. *, +, -, <, <=, =, >, <=). I
  decided to go ahead and hard-code their logical info so it is
  explicitly added when we observe their application while type checking.
  This means at the moment you couldn't express the same thing with the
  types we allow users to write... trade offs =\

* unsafe-require/typed/provide added
  see prims-contract.rkt

* misc
  + vector-length has a latent object now
  + tc/dep-lambda checks dependent function bodies (tc-lambda-unit.rkt)
  + logical propositions are extracted when recording initial types
    for identifiers (see tc-let-unit.rkt and perhaps other areas)
  + extract-props returns a simplified type if props were extracted
    in addition to any extracted props to avoid infinite loops
    e.g. old behavior: (extract-props obj {x : Int | x < 42})
                       ==> (< obj 42)
         new behavior: (extract-props obj {x : Int | x < 42})
                       ==> (values Int (< obj 42))

* VecLenPE treatment (uninterpreted-PE?)
  + there is special treatment of the `VecLenPE` path element in several places
     (grep for `uninterpreted-PE?`), this is because `VecLenPE` is the first path element
     that does not refer to some sub-portion of a type. If we learn something about
     the VecLenPE of an object, we need to remember it... and so currently we explicitly
     test for this case and store the info. In the future, if there are more such path elements,
     we would need to add these to the `uninterpreted-PE?` predicate as well (or distinguish
     them from interpreted path elements in some other way)
```

TODO
- [x] dependent function documentation
- [x] better documentation for linear integer stuff
- [x] add docs for unsafe-require/typed[/provide]
- [x] add SPECIFIC documentation of which functions get more logical info when the `#:with-logical-refinements` keyword is provided on the `#lang` line


pnwamk:
The math tests are failing due to some exponential proposition growth. It appears bubbling up the types of existential variables in the then- and else-propositions of type checking results is not a feasible technique (which isn't that surprising... it's pretty hacky). Will have to look at more efficient ways to propagate that information up (mutation? another field in tc-results? ?)

pnwamk:
we're passing tests and have some preliminary performance stats.

I think I need to write up a summary still and maybe play around with the perf degradation and see if I can narrow down what's causing it.

```
performance results:
schml-specify-rep: 2.2% faster (6.36s to 6.22s)
schml-interp-casts-help: 3.4% slower (15.86s to 16.4s)
parser: 7.04% slower (1.42s to 1.52s)
old-metrics: 4.02% slower (1.74s to 1.81s)
new-metrics: 4.8% slower (2.29s to 2.4s)
math-flonum: 10.11% slower (2.77s to 3.05s)
fsm: 4.29% slower (5.6s to 5.84s)
forth: 5.56% slower (9.72s to 10.26s)
dungeon: 4.9% slower (10.61s to 11.13s)
acquire: 1.48% slower (28.37s to 28.79s)
```

(perf results calculated by [this](https://github.com/pnwamk/tr-performance) code)

pnwamk:
w/ commit b8d6e05 we get the following timing results (notably the math related file is now more or less unaffected):

```
6.10 vs PR
Compile time ratio (old time / new time):
schml-specify-rep: 1.02 (i.e. faster, 6.33s (σ 0.01) to 6.18s (σ 0.01))
schml-interp-casts-help: 0.99 (i.e. slower, 15.84s (σ 0.02) to 16.0s (σ 0.05))
parser: 0.93 (i.e. slower, 1.42s (σ 0.0) to 1.52s (σ 0.0))
old-metrics: 0.96 (i.e. slower, 1.74s (σ 0.0) to 1.81s (σ 0.0))
new-metrics: 1.0 (i.e. slower, 2.3s (σ 0.0) to 2.3s (σ 0.0))
math-flonum: 0.99 (i.e. slower, 2.77s (σ 0.01) to 2.81s (σ 0.0))
fsm: 0.95 (i.e. slower, 5.58s (σ 0.0) to 5.85s (σ 0.01))
forth: 0.95 (i.e. slower, 9.73s (σ 0.01) to 10.28s (σ 0.02))
dungeon: 0.96 (i.e. slower, 10.56s (σ 0.01) to 10.96s (σ 0.02))
acquire: 0.98 (i.e. slower, 28.33s (σ 0.07) to 28.83s (σ 0.03))


;; 6.2 vs PR
Compile time ratio (old time / new time):
schml-specify-rep: 2.17 (i.e. faster, 13.43s (σ 0.01) to 6.2s (σ 0.01))
schml-interp-casts-help: 9.54 (i.e. faster, 152.69s (σ 1.34) to 16.0s (σ 0.05))
parser: 0.91 (i.e. slower, 1.38s (σ 0.0) to 1.52s (σ 0.0))
old-metrics: 0.99 (i.e. slower, 1.79s (σ 0.0) to 1.81s (σ 0.01))
new-metrics: 1.41 (i.e. faster, 3.25s (σ 0.01) to 2.3s (σ 0.0))
math-flonum: 1.12 (i.e. faster, 3.14s (σ 0.01) to 2.81s (σ 0.0))
fsm: 0.95 (i.e. slower, 5.58s (σ 0.0) to 5.85s (σ 0.02))
forth: 0.97 (i.e. slower, 9.93s (σ 0.03) to 10.23s (σ 0.02))
dungeon: 0.95 (i.e. slower, 10.39s (σ 0.01) to 10.98s (σ 0.04))
acquire: 1.11 (i.e. faster, 31.98s (σ 0.03) to 28.85s (σ 0.04))
```

typed-racket-lib/typed-racket/private/parse-type.rkt
```diff
@@ -223,6 +237,36 @@
...
+  #:description "dependent function argument"
+  #:attributes (name local-name deps type-stx refinement-prop-stx)
+  (pattern [name:id
+            :colon^
+            (~optional (dep:id ...)
+                       #:defaults ([(dep 1) null]))
+            type-stx:expr
+            (~optional (~seq #:where refinement-prop-stx:expr)
+                       #:defaults ([refinement-prop-stx #'Top]))]
+           #:fail-when (member #'name (syntax->list #'(dep ...)))
```
bennn: I think this error-checking should be with the checks on 762. So the syntax class just does parsing, and all the checking happens in the same place.

typed-racket-lib/typed-racket/rep/free-ids.rkt
```diff
@@ -0,0 +1,82 @@
...
+      [_ (Rep-for-each rep loop!)]))
+  ids)
+
+
+;; - - - - - - - - - - - - - - - - - - - - - - - - - -
+;; helper function for dependent arg id calculations
+;; - - - - - - - - - - - - - - - - - - - - - - - - - -
+;; note: in my testing (up to 12 args or so) lists
+;; seemed to be the clear winner over free-id-tables
+;; in terms of speed/memory/etc, so lists it is!
```
bennn: Can you put the test code in a gist or something, and link to it?

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -558,6 +559,43 @@
...
+(def-type DFun ([dom (listof Type?)]
+                [rng SomeValues?])
+  [#:mask mask:procedure]
+  [#:frees (f) (combine-frees (cons (f rng)
+                                    (for/list ([d (in-list dom)])
+                                      (flip-variances (f d)))))]
+  [#:fmap (f) (make-DFun (map f dom) (f rng))]
+  [#:for-each (f) (for-each f dom) (f rng)])
+
+(define/provide (SomeDFun? t)
```
bennn: Is this unused?

typed-racket-lib/typed-racket/rep/free-ids.rkt
```diff
@@ -0,0 +1,82 @@
+#lang racket/base
+
```
samth: Let's have a comment here about exactly what free ids are being discussed here.

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -1344,20 +1363,23 @@
 (define (term-var-transform initial transform)
   (let rec/lvl ([rep initial] [lvl 0])
     (define (rec rep) (rec/lvl rep lvl))
-    (define (rec/inc rep) (rec/lvl rep (add1 lvl)))
```
samth: Why delete this?

typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt
```diff
@@ -56,15 +58,54 @@
               #:when (not (Empty? o)))
      (list* idx o t))))
 
+(define (dfunapp-check! args-stx argress dom)
```
samth: Say why this function is needed.

typed-racket-lib/typed-racket/types/numeric-tower.rkt
```diff
@@ -102,3 +111,117 @@
...
+     (hash-ref all-bounded-int-types nbits #f)]
+    [(BaseUnion: 0 nbits)
+     (hash-ref all-bounded-int-types nbits #f)]
+    [_ #f]))
+
+
+;; given the fact that 'obj' is of type 'type',
+;; look inside of type trying to learn
+;; more info about obj
+(define (extract-props obj type)
```
samth: Maybe a comment about why this is here.

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -363,6 +388,7 @@
     (parameterize ([temp-objs seq])
       . body)))
 
+
 ;; the algorithm for recursive types transcribed directly from TAPL, pg 305
```
samth: Tragically no longer true.

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -397,15 +423,54 @@
...
+                              (cons lower-bound upper-bound)))
+        #:when (and (with-logical-refinements?)
+                    (subtype* A t1 -Int obj)
+                    (provable-int-subtype? A t1 lower-bound upper-bound obj))
+        A]
+       [(_ _) ;; otherwise we case on t1
         (subtype-cases A t1 t2 obj)])]))
 
+
+(define (provable-int-subtype? A t1 lower-bound upper-bound obj)
```
samth: Add a comment about this function.

typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -4021,6 +4028,152 @@
...
+               (: sum-ref (-> ([v : (Vectorof Number)]
+                               [i : (v) Natural #:where (< i (vector-length v))]
+                               [j : (i v) Natural #:where (< i j (vector-length v))])
+                              Any))
+               (define (sum-ref v a b)
+                 (+ (vector-ref v a)
+                    (vector-ref v b)))
+               (void))
+             -Void]
+       ;; ensure currently unsupported functions error for DFun types
```
samth: Add error tests for varargs.

pnwamk:
I'm going to add 'fixed' after in-line comments when I address them so I can keep track of what I've done... if there's a better way let me know! =)

typed-racket-doc/typed-racket/scribblings/reference/experimental.scrbl
```diff
@@ -13,11 +13,12 @@
 These features are currently experimental and subject to change.
 
 @defform[(declare-refinement id)]{Declares @racket[id] to be usable in
-refinement types.}
+a @racket[Refinement] types.}
```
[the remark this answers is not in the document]
pnwamk: fixed

typed-racket-doc/typed-racket/scribblings/reference/experimental.scrbl
```diff
@@ -47,13 +48,12 @@ you want, so you shouldn't use @racket[make-predicate] with these types.
...
 
-Refinement types have been added to Typed Racket's core, but
-Typed Racket does not yet have function types which allow
-dependencies between argument types, limiting how useful
-refinements are for now. Allowing argument dependencies
-is on our `to do' list.
+Typed Racket allows types to be `refined' or `constrained'
+by logical propositions. These propositions can mention
+certain program terms, allowing a program's types to depend
+on function values.
```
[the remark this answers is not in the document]
pnwamk: fixed

typed-racket-doc/typed-racket/scribblings/reference/experimental.scrbl
```diff
@@ -113,13 +114,17 @@ depending on which of their subcomponents hold hold:
 
 @racket[(unless p q)] is equivalent to @racket[(or p q)].
 
-Typed Racket's linear integer reasoning is turned off by
-default. If you want to activate it, you must add the
-@racket[#:with-linear-integer-arithmetic] keyword when
-specifying the language of your program:
+In addition to reasoning about propositions regarding types
+(i.e. something is or is not of some particular type), Typed
+Racket is equipped with a simple, linear integer arithmetic
```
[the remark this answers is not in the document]
pnwamk: fixed

typed-racket-doc/typed-racket/scribblings/reference/experimental.scrbl
```diff
@@ -134,3 +139,57 @@ detects that the comparison is guaranteed to produce
 @racket[#f], and thus the clearly ill-typed `then'-branch is
 ignored by the type checker since it is guaranteed to be
 dead code.
+
+
+@section{Dependent Function Types}
+
+Typed Racket also supports explicitly dependent function types:
```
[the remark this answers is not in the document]
pnwamk: fixed

typed-racket-doc/typed-racket/scribblings/reference/experimental.scrbl
```diff
@@ -134,3 +139,57 @@ detects that the comparison is guaranteed to produce
...
+  
+@defform*/subs[#:link-target? #f #:id -> #:literals (:)
+               [(-> ([id : opt-deps type opt-where] ...) type opt-props)]
```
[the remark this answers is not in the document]
pnwamk: I talked with Sam about this and  punted the decision making aspect to him =)

I personally prefer re-using `->` since the presence of `:` in the domain clauses allows us to unambiguously distinguish dependent and non-dependent functions (TR reserves `:`, so users cannot meaningfully use it except in a very limited number of places). YMMV.

typed-racket-doc/typed-racket/scribblings/reference/experimental.scrbl
```diff
@@ -134,3 +139,57 @@ detects that the comparison is guaranteed to produce
...
+
+
+The syntax is partially borrowed from Racket's dependent
+contracts (i.e. @racket[->i]).
+
+Each function argument has a name, an optional list of
+identifiers it depends on, an argument type, and an optional
+@racket[#:where] clause that can refine the type. An
+argument's type can mention (i.e. depend on) other arguments
+by name if they appear in its list of dependencies.
```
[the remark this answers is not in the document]
pnwamk: yep - fixed

typed-racket-lib/typed-racket/private/parse-type.rkt
pnwamk: good point - fixed

typed-racket-lib/typed-racket/rep/free-ids.rkt
pnwamk: I don't have the code any more =(

Are we interested in having some on hand?

Is it unwise to assume just sticking with the list representation for these dependencies (i.e. what we parse from the syntax) is going to be faster than bothering to build free-id-tables and other data structures for this calculation? I'm just trying to imagine in practice how many args are there going to be, how large will these dependency lists actually be, etc.

typed-racket-lib/typed-racket/rep/type-rep.rkt
pnwamk: looks like it -- fixed!

typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt
pnwamk: fixed

typed-racket-lib/typed-racket/types/numeric-tower.rkt
pnwamk: fixed

typed-racket-lib/typed-racket/types/subtype.rkt
pnwamk: comment updated

typed-racket-lib/typed-racket/types/subtype.rkt
pnwamk: fixed

pnwamk:
I've addressed just about every PR comment. There's just a slight bug I need to fix when I get home (it's the failing integration test).

pnwamk:
@samth I believe the only thing left before this is ready to merge is whether we allow `->` with dependent functions (as far as the user level syntax), or should we require some other arrow (e.g. `->i`, etc), or should we support `->` _and_ other arrows, etc.

samth:
@rfindler I'm interested in your thoughts on the naming question for writing dependent function types in Typed Racket. There are a few options:

1. Re-use `->` for dependent functions. Since types aren't just expressions, it's possible to have this work fine. This is what @pnwamk has already implemented. Cons: not like contracts.
2. Use `->i` for dependent functions. Pros: aligns with contracts. Cons: the `i` refers to an aspect of the semantics that isn't relevant for type checking, and is thus confusing.
3. Allow `->d` and `->i`, meaning the same thing.
4. Allow `->`, `->d`, and `->i`, all meaning the same thing. 

I'm somewhat leaning toward 4, but I'd like to know what you think as well.

typed-racket-lib/typed-racket/static-contracts/structures.rkt
```diff
@@ -9,6 +9,10 @@
 
 (provide prop:combinator-name gen:sc)
 
+(provide static-contract-may-contain-free-ids?)
+
+(define static-contract-may-contain-free-ids? (make-parameter #f))
```
bennn: Could this be a free id set (of the ids that can appear "free") instead of a boolean?

typed-racket-lib/typed-racket/static-contracts/combinators/dep-function.rkt
```diff
@@ -0,0 +1,79 @@
...
+                                      [(not (null? dep-ids))
+                                       (parameterize ([static-contract-may-contain-free-ids? #t])
+                                         (rec d/sc))]
+                                      [else (rec d/sc)]))]
+                                   [else (make-list (length dom/scs) #'any/c)])]
+                      [(dep ...) deps]
+                      [(r-deps ...) rng-deps])
+          #`(->i ([id dep c] ...)
+                 #,(cond
+                     [typed-side? #'any]
```
bennn: I'm worried about this because of #598 .
Is `typed-side?` definitely `#false` when the return type is a higher-order value?

typed-racket-lib/typed-racket/typecheck/tc-app-helper.rkt
```diff
@@ -124,12 +128,17 @@
...
+                          [dom-t (in-list (list-extend arg-tys (car doms) #f))]
                           [arg-t (in-list (list-extend (car doms) arg-tys #f))]
-                          [i (in-naturals 1)])
+                          [arg-id (in-list/rest arg-names #f)])
                          (let ([dom-t (or dom-t "-none-")]
                                [arg-t (or arg-t "-none-")])
-                           (format "Argument ~a:\n  Expected: ~a\n  Given:    ~a\n" i (make-printable dom-t) (make-printable arg-t))))
+                           (format "Argument ~a~a:\n  Expected: ~a\n  Given:    ~a\n"
+                                   i
+                                   (if arg-id (format " (~a)" (syntax-e arg-id)) "")
```
bennn: minor: if `arg-id`, I think it'd be better to print "Argument x (position i): ....."

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
```diff
@@ -39,6 +40,9 @@
             [(cons p ps)
              (match p
                [(TypeProp: (Path: pes (? identifier? x)) pt)
+                #:when (match pes
```
bennn: This (and line 79) should be a helper function or match expander --- there's going to be more cases like `VecLenPE` in the future, right?

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -269,8 +269,8 @@
                         body))])]))
 
 
-;; positional: natural? - the number of positional arguments
-;; rest: boolean? - if there is a positional argument
+;; positional: (listof identifier?)
+;; rest: id or #f
 ;; syntax: syntax? - the improper syntax list of identifiers
```
bennn: Is the "improper syntax list" the same as the `(listof identifier?)` in the positional field?

(The comment makes it sound like it is. I think `syntax` must be something different, but not sure what that could be.)

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -406,12 +406,45 @@
...
+        (define instantiated-deps
+          (for/list ([d (in-list dom)])
+            (instantiate-obj d arg-names)))
+        (with-extended-lexical-env
+            [#:identifiers arg-names
+             #:types instantiated-deps]
+          (tc-body/check body (values->tc-results rng (map -id-path arg-names))))
+        dep-fun-ty])]
+    [(fs bs)
+     (tc-error/expr #:return dep-fun-ty
```
bennn: This code is unreachable at the moment right? Since there's no `case->d`.

typed-racket-lib/typed-racket/static-contracts/combinators/dep-function.rkt
bennn: Ok, I think you should definitely delete this line.

Instead, let the optimizer reduce the range to `any/sc` when that makes sense, and add a case to [`static-contracts/optimize.rkt`](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/static-contracts/optimize.rkt#L115) to let the function here (I mean `sc->contract`) know when it should replace the range with `any`.

typed-racket-lib/typed-racket/static-contracts/combinators/dep-function.rkt
```diff
@@ -0,0 +1,79 @@
...
+(define depth 0)
+
+(struct dep-function/sc static-contract (typed-side?
+                                         ids
+                                         dom
+                                         dom-deps
+                                         rng
+                                         rng-deps)
+  #:transparent
+  #:property prop:combinator-name "dep->/sc"
```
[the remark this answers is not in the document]
pnwamk: After seeing how `any/sc` and `any/c` don't correspond as directly with each other as their name (and the comments where they are defined) might suggest, I worry such direct punning may do more harm than good. Perhaps I'm being paranoid.

typed-racket-lib/typed-racket/static-contracts/combinators/dep-function.rkt
pnwamk: I dropped the `typed-side?` optimizations -- glancing at all that code, it's not obvious to me what optimizations for dependent functions are worth adding to optimize.rkt... so perhaps I'll leave that for another day ;-)

typed-racket-lib/typed-racket/static-contracts/structures.rkt
pnwamk: What would be ideal is if static-contracts knew what their set of free ids were -- then we wouldn't need a parameter. Without that info... I think this is the best we can do.

I added a comment describing why we're using the parameter and the note about if static-contracts knew their free ids this would be unnecessary.

typed-racket-lib/typed-racket/static-contracts/combinators/dep-function.rkt
pnwamk: Okay, I thought about it a little more and came up w/ a simple case we can safely replace the range with `any` for after the optimization pass (as you suggested):

```racket
[(and typed-side? (andmap any/sc? rng-deps)) #'any]
```

typed-racket-lib/typed-racket/typecheck/tc-app-helper.rkt
pnwamk: agreed - done

typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
pnwamk: Good point -- I just refactored this part quite a bit and created an `uninterpreted-PE?` predicate that can later be extended to include similar path elements w/o modifying this code.

Now to test the refactor...

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
pnwamk: yes and no... its more or less `(append positional (or id '()))` -- i.e. what `read` produces when reading the function's definition (right?).

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
pnwamk: Yah -- it's more of an artifact of how "listy" all of our lambda/function related code is, I think (i.e. all of our functions are actually case->, etc)

pnwamk:
Changes from comments have been made, commits are squashed and ready for merge, just waiting for word on additional/different `->` syntax for dependent functions (if any is desired in this PR).

rfindler:
I think the right approach (which is actually what seems to be being followed, but perhaps not articulated, so maybe not explicitly followed?) is that whenever there is an arbitrary choice in the syntax to be made, follow the contract library (or talk to me to get the contract library extended/changed). So, writing the dependencies explicitly in parens the way it happens above matches, good. The parens are in the same places, good. The `#:where` keyword isn't something that has a direct correspondence in the contract system; well perhaps via `#:pre`: possibly good, possibly bad. I guess the right thing is to think about whether or not I can add something simple to the contract system to match what you have to do, or if you can use `#:pre` instead. Or maybe I add something *and* you support `#:pre`. Etc.

I'm happy to discuss specifics more. 

Meanwhile, I agree that the `i` isn't suitable because there is no such thing as "indy". I'd probably just use `->` only; or I'd separate them and use `i` only as an homage to the contract system. I wouldn't have multiple different ways to spell the same thing and I wouldn't use `d` because that is a different notation in the contract system.

samth:
Ok, we'll go with plain `->`. @pnwamk what do you think about `#:pre` (it makes sense as a precondition)?

pnwamk:
A #:pre could definitely express anything that could be written in a
#:where clause.

Having #:pre would probably make the most sense as an _additional_ field on
our representation of dependent function types. This has the downside of
making our dependent function types a little more complicated, but would
probably provide more useful error messages more often for users (i.e. we
could much more easily report when the domain types were satisfied but the
precondition was unprovable at a given application site using a #:pre than
we could if we were merely using #:where clauses as syntactic sugar for a
Refinement type).

rfindler:
I don't know what a where is, but my initial reaction is to ask why pre and
where aren't interchangeable (is the answer only that pre allows Racket
expressions and pre is more restricted? If so then I suggest changing the
notation to write "pre" instead of "where" to make things more similar)

Also: is there a grammar somewhere?

samth:
What would be the distinction between `#:pre` and `#:where` that you're thinking of? And would this distinction make sense for contracts too?

pnwamk:
The grammar is in the updated TR docs that are a part of the pull request.

Basically

[x : T #:where P(x)]

as a domain clause is shorthand for

[x : (Refine [x : T] P(x))]

where that Refine means values x of type T where P(x) holds, i.e. where
just lets you refine a type without having to introduce another binder in
the syntax.

Any such refinement proposition P(x), however, could just as easily be
moved into a precondition clause (like #:pre for ->i) without any loss of
expressiveness.

(Tried to clarify from my phone -- hope it wasn't all for not!)

rfindler:
Thanks Andrew that helps!

So: it seems to me that the syntax differences "where" vs "pre" are just
arbitrary choices and so should be made the same.

I don't have a strong opinion about adding where to the contract system
(although I confess to not being super excited about trying to find time to
work on it), but I do have a strong option that we don't remove pre from
the contract system.

Robby

samth:
Just to be clear, the type for `checked-vector-ref` on integer vectors is currently:

```
(-> [v (Vectorof Integer)]
     [i (v) Integer #:where (< i (vector-length v))]
     Integer)
```

and the contract would be:

```
(->i [v () (vectorof integer?)]
      [i () integer?]
      #:pre (< i (vector-length v))
     [_ integer?] )
```

Here, the distinction between `#:pre` and `#:where` feels significant, especially since the encoding of `#:pre` into refinement types would be not totally obvious. My feeling is that both features make sense in the other system, of course with the same keywords.

@rfindler Is what you're suggesting that TR use `#:pre` in the current place in the syntax that uses `#:where`, or that it switch to the syntax more like `->i` with one big precondition for everything?

rfindler:
That's not a syntactically valid `->i`. More important than the missing parens, perhaps, is that `#:pre` needs more stuff:

This one is valid:

```
(->i ([v () (vectorof integer?)]
      [i () integer?])
      #:pre (i v) (< i (vector-length v))
      [_ integer?])
```

but I don't think it would be advised to write it like that. Write it like this instead:

```
(->i ([v (vectorof integer?)]
      [i integer?])
      #:pre (i v) (< i (vector-length v))
      [_ integer?])
```

I can believe that the `#:where` clauses give more information that's useful to the type checker than `#:pre` does.

I can try to add `#:where` to `->i` (but if someone else wants to, that'd be great).

I'll point out that there is a proper `->i` parser in the contract system that TR could, I believe, meaningfully use.

rfindler:
Looking at the docs, I'm not sure what `#:+`, `#:-`, and `#:obj` do. And in addition to the things we've explicitly discussed already, there is an extra `:` in the TR version.

(I'll also note that the contract system's `->i` parser does the acyclicity checking.)

mfelleisen:
Sam wrote "My feeling is that both features make sense in the other system, of course with the same keywords.” 

In a sense, the concept of where is available in the contract system via and/c. Indeed, is is true that 

 Integer #:where (< i (vector-length v))

is the moral equivalent of 

 (and/c integer (</c (vector-length v)). 

In the same spirit could we just write 

  (and/type Integer (< i (vector-length v)))

samth:
We already have `and/type`, that's intersections. What we don't have is `</type`, which we could add. 

More generally, this feature is already available for types too -- `#:where` is just syntactic sugar, but it's useful sugar that we've found we wanted in practice.

rfindler:
@mfelleisen I think the issue is that we don't want slightly different notations for what are things that are supposed to be similar. So while we could, in a contract, use `and/c` and in a dependent type use `#:where`, we should probably change either the dependent type or the contract system so they both either allow or disallow `#:where`. Ditto for `#:pre`. 

More generally, trying to get similar notations for things that are similar when we can seems helpful to our users. (I think that's all that's going on here.)

pnwamk:
I added `#:pre` to dependent function types. It works quite nicely, allowing us to easily generate more detailed error messages than we otherwise would normally (i.e. we can highlight when the issue was that we couldn't prove the precondition, instead of just saying we couldn't apply the function for some reason).

With `#:pre` added, there's really no reason to want a `#:where` clause anymore (i.e. in this PR). Perhaps at a later date there will be reasons to consider them (or other syntactic options), but I don't think it's worth worrying about right now.

Here's the dependent function syntax with `#:pre` added and `#:where` removed:

```
(-> ([id : opt-deps arg-type] ...)
    opt-pre
    rang-type
    opt-props)

 
opt-deps = 
  | (id ...)
     
opt-pre = 
  | #:pre opt-deps proposition
     
opt-props = 
  | opt-pos-prop opt-neg-prop opt-obj
     
opt-pos-prop = 
  | #:+ proposition
     
opt-neg-prop = 
  | #:- proposition
     
opt-obj = 
  | #:object symbolic-object


proposition = Top
  | Bot
  | (: symbolic-object type)
  | (! symbolic-object type)
  | (and proposition ...)
  | (or proposition ...)
  | (when proposition proposition)
  | (unless proposition proposition)
  | (if proposition proposition proposition)
  | (linear-comp symbolic-object symbolic-object)
     
linear-comp = <
  | <=
  | =
  | >=
  | >
     
symbolic-object = exact-integer
  | linear-term
  | (+ linear-term linear-term ...)
  | (- linear-term linear-term ...)
     
linear-term = symbolic-path
  | (* exact-integer symbolic-path)
     
symbolic-path = id
  | (path-elem symbolic-path)
     
path-elem = car
  | cdr
  | vector-length
```

rfindler:
That is really great!

Small comments:

in `->i`, #:pre requires the dependencies list (it isn't optional)

Can we drop the `:` in `([id : opt-deps arg-type] ...)` (first line of the grammar)?

The `when` and `unless` aren't the same as Racket's. I don't know what they do, but perhaps they should have different names or fewer subexpressions?

samth:
@rfindler Here, `id : type` is not an arbitrary choice, it's consistency with everywhere else in TR, which also uses `id : type` (such as `define`, `let`, and `lambda`).

Is `#:pre` more efficient if you don't have any dependencies? 

I think `when` and `unless` are like Racket, although perhaps they should have arbitrarily-many subexpressions. `(when A B)` is like `B` if `A` is true (and the reverse for `unless`).

rfindler:
On Thu, Sep 21, 2017 at 5:09 PM Sam Tobin-Hochstadt <

> @rfindler <https://github.com/rfindler> Here, id : type is not an
> arbitrary choice, it's consistency with everywhere else in TR, which also
> uses id : type (such as define, let, and lambda).
>

Ah of course! Silly me.


Is #:pre more efficient if you don't have any dependencies?
>

An empty sequence? Nope.

I think when and unless are like Racket, although perhaps they should have
> arbitrarily-many subexpressions. (when A B) is like B if A is true (and
> the reverse for unless).
>


What does implicit begin mean here?

pnwamk:
Oops, sorry. the grammar was incorrect (the parser for `#:pre` clauses did require the dep list).

`when` and `unless` are described in the documentation (i.e. near their grammar) as follows:

`(when p q)` is equivalent to `(or (not p) (and p q))`

`(unless p q)` is equivalent to `(or p q)`

i.e. they behave as Racket's `when` and `unless` w.r.t. how they behave as boolean combinators. I omitted the "implicit begin" and fixed their arity since there are not "effects" or other sensible things that would go there in a type.

PR #589, opened 2017-07-20
Fix cast under multiple cases of case->

### LINKED ISSUE #587: Cast produces wrong contract for multiple cases of a case->
##### What version of Racket are you using?
Version 6.10.0.1 from the 2017-07-15 snapshot

##### What program did you run?
```racket
#lang typed/racket

(: f : (case-> (-> (Boxof String) Any)
               (-> (Boxof Integer) Any)))
(define (f x)
  (cast x Any))

(f (box "a"))
```

##### What should have happened?
I expected it to pass and print out `'#&"a"`, since a box containing a string should be a valid argument to `f`.

(btw, it does the correct thing when passed `(box 1)`)

##### What was the error message?
```racket
. . broke its own contract
  promised: Integer
  produced: "a"
  in: the content of
      (box/c Integer)
  contract from: typed-world
  blaming: typed-world
  at: /Users/Alex/Desktop/case-arrow-cast-contract.rkt:6.2
```

The error message says the contract protecting the typed value was `(box/c Integer)`, where it should be `(or/c (box/c String) (box/c Integer))`. 

#### The cause, I think...
When typed racket typechecks the body, it typechecks it multiple times. Once with `(Boxof String)` and once with `(Boxof Integer)`. Both times it's [_mutating the cast table_][mutating-cast-table], overriding the old type and completely replacing it with the new type. This means it overrides the `(box/c String)` to replace it with `(box/c Integer)`.

When actually both are incomplete. The contract should include all of them together in an `or/c`.

  [mutating-cast-table]: https://github.com/racket/typed-racket/blob/6e3c509a86c6ee7706e272a6c55930ee95bbfb3f/typed-racket-lib/typed-racket/base-env/prims-contract.rkt#L353-L355

AlexKnauth:
Fixes #587.


bennn:
Looks good to me.

I was wondering if the `or/c` would let me do something weird later on, like writing an integer to `(f (box "a"))`, but the `Any` prevents all writes so I think it's safe.

AlexKnauth:
Even without the `Any`, the `or/c` wouldn't allow writes unless the value passes both the `integer?` and `string?` contracts. If it doesn't pass both interior contracts it's a violation:
```racket
#lang typed/racket
(: f : (case-> (-> (Boxof (U #f String)) (Boxof Any))
               (-> (Boxof (U #f Integer)) (Boxof Any))))
(define (f x)
  (cast x (Boxof Any)))

(: b : (Boxof (U #f String)))
(define b (box #f))
(define b* (f b))
(set-box! b* 1) ;contract violation
```
This keeps it sound. Without the change in this PR, that `set-box!` call would pass, and then `(unbox b)` would have type `(U #f String)` but value `1`.

AlexKnauth:
@bennn approves, should I merge?

typed-racket-lib/typed-racket/private/cast-table.rkt
```diff
@@ -16,10 +16,12 @@
 
 (define cast-table (make-free-id-table))
 
-;; cast-table-set! : Id Type-Stx -> Void
-(define (cast-table-set! id type-stx)
-  (free-id-table-set! cast-table id type-stx))
+;; cast-table-add! : Id Type-Stx -> Void
+(define (cast-table-add! id type-stx)
```
pnwamk: Could we add a brief purpose statement to cast-table-add! -- i.e. why is there a list of casts that we're adding to? When would there be more than one vs just one? (so the function is a little more self documenting in what it does in the grand scheme of things)

AlexKnauth: Okay. Are the new comments clear enough?

PR #591, opened 2017-07-26
doc: scribblings for 'syntax/source-syntax'

bennn:
Add documentation for `syntax/source-syntax` module.
(This fixes a TODO on the package server.)

The example is boring because it returns different results inside `examples` than in a top-level module.

I thought about using a `racketblock` instead, but I think a boring `examples` is better.

### DIFF
--- a/source-syntax/info.rkt
+++ b/source-syntax/info.rkt
@@ -9,3 +9,5 @@
 (define pkg-authors '(samth stamourv eli))
 
 (define version "1.1")
+
+(define scribblings '(("scribblings/source-syntax.scrbl" () (tool-library))))
--- a/source-syntax/scribblings/source-syntax.scrbl
+++ b/source-syntax/scribblings/source-syntax.scrbl
@@ -1,0 +1,24 @@
+#lang scribble/manual
+@require[scribble/example (for-label racket/base racket/contract)]
+
+@title{Source Syntax}
+
+@defmodule[syntax/source-syntax]{}
+
+@defproc[(recover-source-syntax [orig syntax?]
+                                [expanded syntax?]
+                                [#:traverse-now? now? boolean? #f])
+         (-> syntax? (or/c syntax? #f))]{
+  Return a procedure that accepts a syntax object from @racket[expanded]
+  and returns the outermost syntax object in @racket[orig] that has the same
+  location as the given syntax object.
+  If no syntax object in @racket[orig] has the same location as the given syntax
+  object, the procedure repeats with the parent of the given syntax object.
+}
+
+@examples[#:eval (make-base-eval '(require syntax/source-syntax))
+  (let* ([orig #'(λ (x [y 0]) (+ x y))]
+         [expanded (expand orig)]
+         [recovered ((recover-source-syntax orig expanded) expanded)])
+    (syntax? recovered))
+]


bennn:
Preview:
screen shot 2017-07-26 at 18 22 55

source-syntax/info.rkt
```diff
@@ -8,4 +8,6 @@
 
 (define pkg-authors '(samth stamourv eli))
 
-(define version "1.1")
\ No newline at end of file
+(define version "1.1")
+
+(define scribblings '(("scribblings/source-syntax.scrbl" () (legacy))))
```
stamourv: Why `legacy`?

source-syntax/scribblings/source-syntax.scrbl
```diff
@@ -0,0 +1,23 @@
...
+
+@defmodule[syntax/source-syntax]{}
+
+@defproc[(recover-source-syntax [orig syntax?]
+                                [expanded syntax?]
+                                [#:traverse-now? now? boolean? #f])
+         (-> syntax? (or/c syntax? #f))]{
+  Return a procedure that accepts a syntax object from @racket[expanded]
+  and returns the outermost syntax object in @racket[orig] that has the same
+  location as the given syntax object.
```
stamourv: I forget the exact details (@samth would know), but I think this is only the best-case scenario.
I think it's sometimes necessary to fall back to an enclosing syntax object.

source-syntax/info.rkt
bennn: I guess `tool-library` would be better?

stamourv: I think so.

source-syntax/scribblings/source-syntax.scrbl
bennn: What do you mean by "enclosing"?

(If `A` and `B` are syntax objects in `orig` with the same location `loc` and `A` contains `B`, then a lookup for `loc` should return `A` (aka the enclosing object))

stamourv: I think the tricky case arises when no original syntax object has location `loc`. In which case (IIRC) the library tries to find the tightest original syntax object whose location surrounds `loc`. But I forget the details.

bennn: Ok, looks like:
- if you search with a syntax object `e` with location `loc`
- and no original syntax object has location `loc`
- then you get the result for the nearest ancestor of `e` that has a location in the original syntax

stamourv: That sounds right.
It's a bit ad-hoc, but I think it's helped get better error messages for TR.
In any case, worth documenting, I think.
Thanks for digging it up!

source-syntax/scribblings/source-syntax.scrbl
```diff
@@ -12,7 +12,8 @@
   Return a procedure that accepts a syntax object from @racket[expanded]
   and returns the outermost syntax object in @racket[orig] that has the same
   location as the given syntax object.
-
+  If no syntax object in @racket[orig] has the same location as the given
+  syntax object, the procedure repeats with parent of the given syntax object.
```
stamourv: Nit: "the parent"?
Otherwise the whole thing LGTM.

bennn: oh wow, thank you

samth:
This breaks the dependency checking in the build: http://drdr.racket-lang.org/42252/pkg-src/build/make#output1

More generally, `typed-racket-lib`, which depends on `source-syntax`, should not depend on `scribble` etc. So this documentation needs to go somewhere else, perhaps the `typed-racket-doc` package.

bennn:
#597

PR #596, opened 2017-09-03
Remove typed rackunit code and depend on other package

jackfirth:
Rackunit now has a `rackunit-typed` package that contains this code. The package catalog is down at the moment and `rackunit-typed` has not been added to it yet, so this won't compile.


samth:
Looks good, but we shouldn't merge until `rackunit-typed` is available.

Also `rackunit-typed` should be added as a dependency of `main-distribution`.

jackfirth:
The `rackunit-typed` package is now in the package catalog, and I opened racket/main-distribution#1 to add it to the main distribution.

samth:
I restarted the build.

jackfirth:
The integration tests timed out, are they supposed to take that long?

bennn:
All checks passed!

jackfirth:
I don't have write access, so merge whenever

PR #599, opened 2017-09-17
static-contracts: less or/sc optimization

### LINKED ISSUE #598: Unsound static contracts optimization

bennn:
Closes #598 . 

Update: generalized the optimizer's notion of a `side` so that `or/sc` can say "I'm on the positive side, but don't optimize my immediate children, but do optimize my children's children (if any)"

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/optimize.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/optimize.rkt
@@ -117,22 +117,105 @@
     [(arr/sc: args rest (list (any/sc:) ...))
      (arr/sc args rest #f)]
     [(none/sc:) any/sc]
-    [(app sc-terminal-kind 'flat) any/sc]
+    [(or/sc: (? flat-terminal-kind?) ...) any/sc]
+    [(? flat-terminal-kind?) any/sc]
     [else sc]))
 
+(define (flat-terminal-kind? sc)
+  (eq? 'flat (sc-terminal-kind sc)))
 
+;; The side of a static contract describes the source of the values that
+;;  the contract needs to check.
+;; - 'positive : values exported by the server module
+;; - 'negative : values imported from a client module
+;; - 'both     : values from both server & client
+(define (side? v)
+  (memq v '(positive negative both)))
+
+;; A _weak side_ is a side that is currently unsafe to optimize
+;; Example:
+;;  when optimizing an `(or/sc scs ...)` on the 'positive side,
+;;  each of the `scs` should be optimized on the '(weak positive) side,
+;;  and their sub-contracts --- if any --- may be optimized on the 'positive side
+(define (weak-side? x)
+  (match x
+   [(list 'weak (? side?))
+    #true]
+   [_
+    #false]))
+
+(define (strengthen-side side)
+  (if (weak-side? side)
+    (second side)
+    side))
+
+(define (weaken-side side)
+  (if (weak-side? side)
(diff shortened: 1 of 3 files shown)


bennn:
I think this is ok now.
(I guess I can add more tests for other type constructors under an `or/sc`)

bennn:
I just noticed https://github.com/racket/racket/pull/712 , for the same issue.

That PR is trying to change `(or/sc flat/sc ... chap/sc ...)` into a contract that either:
1. uses one of the `chap/sc`, if the chaperone's first-order check works for the value
2. falls back to `any/sc`

(I'm not sure if that's always safe; depends on how the first order checks for all Typed Racket's types work.)

This PR is more conservative, it just turns the optimizer off for `or/sc` unless all its components are flat.

bennn:
https://github.com/racket/typed-racket/commit/9df037b0f6e6b2df3cc0993631bb2870367e291a

typed-racket-lib/typed-racket/static-contracts/optimize.rkt
```diff
@@ -117,23 +117,106 @@
...
+    ;; Keep same side by default.
+    ;; This is precisely safe for "unguarded" static contracts like and/sc
+    ;;  and conservatively safe for everything else.
+    side]))
+
+;; guarded-sc? : sc? -> boolean?
+;; Returns #true if the given static contract represents a type with a "real"
+;;  type constructor. E.g. list/sc is "real" and or/sc is not.
+(define (guarded-sc? sc)
+  (match sc
```
pnwamk: I was swimming through this code and had the thought -- would it be easier to recognize the type constructors that are _not_ real (i.e. or/sc, and/sc... is that it? maybe 1 or two more?) instead of all the ones that are real?

Maybe there's a reason why that would be problematic, but I figured I would ask just in case.

pnwamk: (also -- I know this PR was merged a while ago, but I wasn't sure where to ask the question about this line... ¯\\\_(ツ)\_/¯)

bennn: That would be shorter, but I picked this because its safer. If `guarded-sc?` leaves something out, then its only a missed optimization and not a soundness hole.

pnwamk:
Maybe as a “To Do” for me later then, it would be nice—for our type and
static contract representations—if they were aware if they are a “true”
type constructor or not. We have lots of code which is essentially
reinventing this wheel. It would be cool if we had a way to reduce that
boiler plate that we were confident was sound as well.

PR #600, opened 2017-09-19
static-contracts: fix list-length/sc

bennn:
Fix a bug in `list-length/sc`.

I think it's impossible to write a `#lang typed/racket` program that illustrates the bug (and doesn't use `require`). But still.

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/combinators/lengths.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/combinators/lengths.rkt
@@ -27,8 +27,7 @@
 (define (list-length/sc* n)
   (if (zero? n)
       empty-list/sc
-      empty-vector/sc))
-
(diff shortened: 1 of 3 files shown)


samth:
Can you write a program that does use `require` that illustrates the bug?

bennn:
The program I have in mind with `require` would look like the unit test, so I don't think there's any point. The program would need to make a `list-length/sc` contract directly.

It's hard to make the program because:
- TR only uses `list-length/sc` in the static contract optimizer, to turn a `(list/sc any/sc ...)` into a `list-length/sc`
- and the only way to get an `any/sc` is to have a flat type in a trusted positive position
- so that means `(list/sc any/sc ...)` is in a trusted positive position, and so is the `list-length/sc`
- and since `list-length/sc` is flat and in a positive position, it gets optimized to `any/sc`

PR #611, opened 2017-09-26
fix domain printing, especially rest args

pnwamk:
Currently the following program:

```racket
#lang typed/racket/base
(define x : Number 42)
(apply + x)
```

has this error message:

```
Type Checker: Bad arguments to function in `apply':
Domains: Number *
Arguments: Number *
 in: (apply + x)
```

Which is awful, since the expected domains line and argument type line are _identical_.

This PR changes it to the following (i.e. it drops the `*` from the `Arguments` line):

```
temp.rkt:7:0: Type Checker: Bad arguments to function in `apply':
Domains: Number *
Arguments: Number

  in: (apply + x)
```

Perhaps there is an even better way to display this? (not sure -- comments welcome)

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-app-helper.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-app-helper.rkt
@@ -78,15 +78,43 @@
     [(tc-any-results: _) (-AnyValues -tt)]
     [_ t]))
 
-(define (stringify-domain dom rst [rng #f])
-  (let ([doms-string (if (null? dom) "" (string-append (stringify (map make-printable dom)) " "))]
-        [rng-string (if rng (format " -> ~a" rng) "")])
+;; creates a "pretty-printed" version of the domain
+;; (and optionally also followed by "-> rng")
+;; of a function, so we can easily print things like:
+;; Domain was          Number Number *
+;; but arguments were  String
+(define/cond-contract (stringify-domain dom rst [rng #f])
+  (->* ((listof (or/c Type? tc-results/c))
+        (or/c #f Type? RestDots?))
+       ((or/c Type? SomeValues? tc-results/c))
+       string?)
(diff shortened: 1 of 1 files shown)


pnwamk:
Also, this program:

```racket
(: my-car (All (A) (case-> (-> ([p : (Pairof A A)]) A #:object (car p))
                           (-> String String))))
(define (my-car p)
  (if (pair? p)
      (car p)
      (error 'foo)))

(string (my-car (cons 0 1)))
```

produced this error message:

```
Type Checker: Polymorphic function `my-car' could not be applied to arguments:
Types: (Pairof A A) #f * -> (A : (Top | Top) : (car (0 0)))
       String #f * -> String
Arguments: (Pairof Zero One) 
Expected result: Char

  in: (my-car (cons 0 1))
```
and now produces this error message with the PR changes in 176190b:

```
Type Checker: Polymorphic function `my-car' could not be applied to arguments:
Types: (Pairof A A) -> (A : (Top | Top) : (car (0 0)))
       String -> String
Arguments: (Pairof Zero One) 
Expected result: Char

  in: (my-car (cons 0 1))
```

typed-racket-lib/typed-racket/typecheck/tc-app-helper.rkt
```diff
@@ -79,14 +79,26 @@
...
+       (format "~a ~a ... ~a~a" doms-string dty dbound rng-string)]
       [rst
-       (format "~a~a *~a" doms-string rst rng-string)]
-      [else (string-append (stringify (map make-printable dom)) rng-string)])))
+       (format "~a~a~a"
+               doms-string
+               (if rst (format "~a *" rst) "")
+               rng-string)])))
+
+(define (stringify-args dom rst)
```
samth: Signature + purpose statement.

PR #612, opened 2017-09-28
tc-app: add missing 'function' annotations

bennn:
Update the type-table with annotations for some identifiers that are handled specially by `tc-app`.

- - -

Are there any tests for this kind of thing --- for checking whether the `type-table` is updated correctly?
If not, I'll try calling `type-check` directly on some syntax objects, then checking that their components all have annotations.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-list.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-list.rkt
@@ -6,7 +6,7 @@
          "utils.rkt"
          syntax/parse syntax/stx racket/match racket/sequence
          (typecheck signatures tc-funapp error-message)
-         (types abbrev utils substitute)
+         (types abbrev type-table utils substitute)
          (rep type-rep)
          (env tvar-env)
          (prefix-in i: (infer infer))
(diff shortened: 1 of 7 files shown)


typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-list.rkt
```diff
@@ -108,29 +109,46 @@
...
+                (add-typeof-expr #'op-name (ret (->* argtys (-Tuple argtys))))
                 (ret (-Tuple argtys))))
             (or result
                 (begin (expected-but-got t (-Tuple (map tc-expr/t args-list)))
                        (fix-results expected)))]
-           [else (ret (-Tuple (map tc-expr/t args-list)))])])]
-     [_ (ret (-Tuple (map tc-expr/t args-list)))])))
+           [else
+            (define argtys (map tc-expr/t args-list))
+            (add-typeof-expr #'op-name (ret (->* argtys (-Tuple argtys))))
```
samth: This allocates `(-Tuple argtys)` twice.

typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -4426,4 +4470,104 @@
...
+
+    ;; tc-app-values
+    (tc/type-table (values 0)
+     ([values (t:-> (-val 0) (-values (list (-val 0))))]))
+    ;; TODO
+    ;(tc/type-table (values 0 0)
+    ; ([values (t:-> (-val 0) (-val 0) (-values (list (-val 0 ) (-val 0))))]))
+    ;; TODO why are tests failing? because props are different? what to do?
+    ;(tc/type-table (call-with-values (λ () 'A) symbol->string)
+    ; ([call-with-values (->* (list (t:-> (-val 'A)) (t:-> -Symbol -String)) -String)]))
```
bennn: Error message:

```
--------------------
type-table tests > Typechecker tests > 4560 (call-with-values (λ () (quote A)) symbol->string)
FAILURE
location:   ("<pkgs>/typed-racket-test/unit-tests/typecheck-tests.rkt" 4560 4 193592 146)
actual:     #(struct:tc-results (#(struct:tc-result (-> (-> 'A) (-> Symbol String) String) (Top | Top) -)) #f)
expected:   #(struct:tc-results (#(struct:tc-result (-> (-> 'A) (-> Symbol String) String) (Top | Top) -)) #f)

(type-of call-with-values) did not return the expected value. The types don't match.
--------------------
```

I don't know what's wrong here.

typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -122,6 +125,38 @@
...
+  ;; test-type-table : syntax? (listof (cons/c identifier? tc-results?)) -> void?
+  ;; Typecheck the given expression
+  ;;  and assert that the type-table has an entry for every identifier
+  ;;  in the given association list.
+  ;;
+  ;; Example: `(test-type-table expr (list (cons f t)))` will
+  ;; 1. type-check `expr`
+  ;; 2. search `expr` for occurrences of `f`
+  ;; 3. check that the type-table entry for `f` matches `t`
+  (define (test-type-table expr assoc)
```
bennn: @samth @pnwamk any comments about this new testing function?

The goal is to make sure the `type-table` has the right mappings after TR type-checks a syntax object.

samth:
Is this ready to merge?

bennn:
No, because there are 3 places where I'm calling `add-typeof-expr` and not able to write a test (listed in my last comment).

I'll take another look, and if I can't write a test then I'll comment out those three places & leave a note.

samth:
I think we should leave the code in the right state, not the wrong one, even if we can't write a test.

bennn:
Ok ... not being able to test makes me worry the annotations are wrong, but the code does look right to me.

Issue #605, opened 2017-09-26
hash


schackbrian2012:
### What version of Racket are you using?
6.10.1

### What program did you run?
```
#lang typed/racket
(hash 0 1 2 3 4 5 6 7 8 9)
```

### What should have happened?
`'#hash((2 . 3) (6 . 7) (0 . 1) (4 . 5) (8 . 9))`

### If you got an error message, please include it here.
```
Type Checker: Polymorphic function `hash' could not be applied to arguments:
Types: a b a b a b a b #f * -> (Immutable-HashTable a b)
       a b a b a b #f * -> (Immutable-HashTable a b)
       a b a b #f * -> (Immutable-HashTable a b)
       a b #f * -> (Immutable-HashTable a b)
       #f * -> (Immutable-HashTable a b)
Arguments: Zero One Positive-Byte Positive-Byte Positive-Byte Positive-Byte Positive-Byte Positive-Byte Positive-Byte Positive-Byte #f *
Expected result: AnyValues
 in: (hash 0 1 2 3 4 5 6 7 8 9)
```

pnwamk:
There is not a type in Typed Racket's vocabulary currently that can describe the type of `hash` (it would require describing rest arguments with a non-uniform type (e.g. the key type followed by the value type followed by the key type etc...).

You can, however, write out the sequence of key/value pairs as a list of cons cells and use `make-hash`.

pnwamk:
or `make-immutable-hash` (depending on whether you wanted a mutable or imutable hash)

schackbrian2012:
@pnwamk Thank you, in this situation, my keys and values were constants, so I used #hash. But this is still a missing and undocumented functionality in typed/racket.

pnwamk:
The type that was printed out documents what `hash` supports (perhaps there are better ways to inform users of the supported type?). In general though, there will be functions where the supported type is not as expressive as the original function, and in those cases you'll have to either decipher what the type means, or ask ;-)

pnwamk:
BTW, [:print-type](https://docs.racket-lang.org/ts-reference/Exploring_Types.html?q=%3Atype#%28form._%28%28lib._typed-racket%2Fbase-env%2Fprims..rkt%29._~3aprint-type%29%29) may be the easiest way to see what Typed Racket thinks the type of an identifier is (if a type exists)

samth:
I think we should just add a special case to do better here.

pnwamk:
[...]

PR #613, opened 2017-09-28
tc-app: add special case for 'hash' functions

bennn:
For #605. Recognize `hash` `hasheq` `hasheqv` and give precise types.

This PR works for the 1 example.

TODO
- [ ] more tests
- [ ] better error message for odd number of arguments
- [ ] really want `restrict` and not `intersect` ?

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-hetero.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-hetero.rkt
@@ -1,6 +1,7 @@
 #lang racket/unit
 
 (require "../../utils/utils.rkt"
+         racket/list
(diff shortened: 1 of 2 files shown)


pnwamk:
I'm most of the way through (I think) adding support for richer rest argument specifications:

https://github.com/pnwamk/typed-racket/tree/hrest

In particular it will let you specify a heterogeneous, repeating pattern such as those used by `hash` and friends.

(Sorry for not communicating better -- didn't expect anyone else to jump on this so quick! =)

bennn:
Sounds good.

Having this special case for `hash` didn't really make sense to me because the return type is so simple.

schackbrian2012:
@bennn @pnwamk Thank you for working on this issue!

PR #617, opened 2017-09-30
static-contract: temporarily fix for incoming syntax

### LINKED ISSUE #616: new type error in old test case
6.10.1.2 (nightly build)

### What program did you run?

https://github.com/florence/cover/blob/master/cover/tests/multibyte-coverage/arrow.rkt

No type error. 

Until the most recent build (today) this file ran fine. I would guess it's related to https://github.com/racket/typed-racket/commit/9df037b0f6e6b2df3cc0993631bb2870367e291a since the cover builds run daily.

```
cover/tests/multibyte-coverage/arrow.rkt:8:32: Type Checker: Type (-> Syntax Any) could not be converted to a contract: required a flat contract but generated a chaperone contract
  in: (→ Syntax Any)
  compilation context...:
   /home/travis/build/florence/cover/cover/tests/multibyte-coverage/arrow.rkt
  context...:
   /home/travis/racket/share/pkgs/typed-racket-lib/typed-racket/private/type-contract.rkt:289:4: unpack268
   /home/travis/racket/collects/racket/private/more-scheme.rkt:163:2: select-handler/no-breaks
   /home/travis/racket/share/pkgs/typed-racket-lib/typed-racket/private/type-contract.rkt:83:0: generate-contract-def
   /home/travis/racket/share/pkgs/typed-racket-lib/typed-racket/private/type-contract.rkt:195:0: change-contract-fixups
   /home/travis/racket/share/pkgs/typed-racket-lib/typed-racket/core.rkt:35:10
   /home/travis/racket/share/pkgs/typed-racket-lib/typed-racket/typed-racket.rkt:23:4
   /home/travis/racket/collects/compiler/cm.rkt:363:0: compile-zo*
   /home/travis/racket/collects/compiler/cm.rkt:572:26
   /home/travis/racket/collects/compiler/cm.rkt:564:42
   /home/travis/racket/collects/compiler/cm.rkt:635:0: compile-root
   /home/travis/racket/collects/compiler/cm.rkt:737:4
   /home/travis/racket/collects/setup/../racket/private/more-scheme.rkt:261:28
   [repeats 1 more time]
   /home/travis/racket/collects/setup/parallel-do.rkt:435:20: loop
```

bennn:
Closes #616 

Add a case to the static contract optimizer so that contracts
 generated from the `Syntax` type disappear in trusted positions.

This restores the "old" behavior for the `Syntax` type --- the behavior
 before fixing the 'or/c' unsoundness in issue 598.

<https://github.com/racket/typed-racket/issues/598>

Note: this fix is temporary, until the definition of `Syntax` can
 be in terms of _immutable_ vectors & boxes.

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/optimize.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/optimize.rkt
@@ -109,7 +109,7 @@
     [else sc]))
 
 
-;; Reduce a static contract assuming that we trusted the current positive side
+;; Reduce a static contract assuming that we trusted the current side
 (define (trusted-side-reduce sc)
   (match sc
     [(->/sc: mand-args opt-args mand-kw-args opt-kw-args rest-arg (list (any/sc:) ...))
@@ -119,6 +119,11 @@
     [(none/sc:) any/sc]
     [(or/sc: (? flat-terminal-kind?) ...) any/sc]
     [(? flat-terminal-kind?) any/sc]
+    [(syntax/sc: (? recursive-sc?))
+     ;;bg; _temporary_ case to allow contracts from the `Syntax` type.
+     ;;    This is temporary until TR has types for immutable-vector
+     ;;    and box-immutable & changes the definition of the `Syntax` type.
+     any/sc]
     [else sc]))
 
 (define (flat-terminal-kind? sc)
--- a/typed-racket-test/unit-tests/contract-tests.rkt
+++ b/typed-racket-test/unit-tests/contract-tests.rkt
@@ -228,6 +228,8 @@
    (t-sc (-lst Univ) (listof/sc any-wrap/sc))
    (t-sc (Un (-lst Univ) -Number) (or/sc number/sc (listof/sc any-wrap/sc)))
 
+   (t-int Any-Syntax syntax? #'#'A #:typed) ;; GitHub issue #616
+
    ;; Github pull request #226
    (let ([ctc (-> Univ -Boolean)])
      ;; Ordinary functions should have a contract


pnwamk:
Can we have a unit test that checks for this in `unit-tests/contract-tests`?

https://github.com/racket/typed-racket/blob/master/typed-racket-test/unit-tests/contract-tests.rkt

And if the included tests are more value then we should keep them. But if they just check the thing a one or two liner unit test could check we could probably do without.

pnwamk:
LGTM

typed-racket-lib/typed-racket/static-contracts/optimize.rkt
```diff
@@ -119,6 +119,11 @@
     [(none/sc:) any/sc]
     [(or/sc: (? flat-terminal-kind?) ...) any/sc]
     [(? flat-terminal-kind?) any/sc]
+    [(syntax/sc: (? recursive-sc?))
+     ;;bg; _temporary_ case to allow contracts from the `Syntax` type.
```
pnwamk: Do we need the `(? recursive-sc?)`? Doesn't anything there _have_ to be something we can check with a flat-contract and so we can optimize any `syntax/sc:` to `any/sc`?

bennn: I want to keep the `(? recursive-sc?)`.

If we optimize any `syntax/sc:` to `any/sc`, then a type like `(Syntaxof (-> Integer Integer))` would get optimized to `any/sc` instead of raising an exception.

- - -

(while I'm in the neighborhood...)

Why `recursive-sc?` and not anything stronger? Checking `recursive-sc?` is a coarse way of looking for the contract I actually care about (the static contract from `Syntax`), but:
- I don't think I can match against the actual recursive contract
- it seems unlikely that someone has code with a non-flat recursive type inside a `Syntaxof` (even less likely than having a function inside)
- the only other strategy I can think of to fix the issue here is changing `type->contract` (which can definitely recognize the `Syntax` type), but then `type->contract` would need to know whether the current position is one that's safe to optimize

Issue #614, opened 2017-09-28
->* w/ opt and rest args is broken


pnwamk:
### What version of Racket are you using?
6.10.1.2
### What program did you run?
```racket
(: foo (->* () (String) #:rest Number Any))
(define foo
  (case-lambda [(str . nums)
                (string->number str) ;; TR error, expected String, given Number in str
                (+ 42 str)] ;; TR error, expected Number, given String in str
               [l 42]))
```
### What should have happened?
Optional arguments come _before_ rest arguments... right? I'm not sure exactly what we're doing in Typed Racket yet... but it seems wrong.

Here's a similar case-lambda in regular Racket with contracts that seems to work fine:

```racket
#lang racket/base
(require rackunit)


(module foo-module racket
  (provide/contract [foo (->* () (string?) #:rest (listof number?) any/c)])
  (define foo
    (case-lambda [(str . nums)
                  (apply + (string->number str) nums)]
                 [l 42])))

(require 'foo-module)

(check-equal? (foo) 42)
(check-equal? (foo "0") 0)
(check-equal? (foo "0" 1 2 3) 6)
(check-exn exn:fail? (λ () (apply foo "0"))) ;; expected a list? given "0"
(check-equal? (apply foo "0" (list 1 2 3)) 6)
```

This is likely related to previously reported issues involving optional and rest arguments (https://github.com/racket/typed-racket/issues/413 https://github.com/racket/typed-racket/issues/432).

Will need to dig deeper to see what the root cause is =\

pnwamk:
working on a fix for this -- shouldn't be too complex I don't think...

pnwamk:
Fixed by ffcf5afe921742abd768ce53ae80d3273dbf93e4.

Issue #432, opened 2016-09-22
Functions with optional arguments and rest argument are unsound


SuzanneSoy:
### What version of Racket are you using?

6.6.0.5
### What program did you run?

```
#lang typed/racket
(ann ((λ ([y : String "default"] . rest)
        y)
      42)
     String)
```
### What should have happened?

The typechecker should reject passing `42` as the value for `y`, but it does not.

Issue #413, opened 2016-08-02
Overlapping case-lambda cases are unsound


jsmaniac:
### What version of Racket are you using?

6.1.0.1
### What program did you run?

``` racket
#lang typed/racket

(define fn (case-lambda
         #:∀ (A)
         [([a : A])
          (ann a A)]
         [[rest : A *]
          (ann rest (Listof A))]))

(ann (fn 1) (Listof One)) ; => 1 (just the number, not in a list)
```
### What should have happened?

The `case-lambda` tries the first case, and if the call does not match the signature, it tries the second, third etc. However a `case→` in Typed/racket is unordered (and can be narrowed down to any case using `(ann fn the-case)`).

The problem arises when two cases `c1` and `c2` have input types with a non-empty intersection. I'm not sure about the conditions required for the type to be safe.

I think it is enough if when two cases of the `case-lambda` have intersecting inputs, they must have equal output types in the `case→`, but maybe this rule is too strong even.

Saying that "when two cases of the case→ intersect, they must have equal output types" would be a too strong condition, as the following example should typecheck but does not respect that stronger rule:

``` racket
#lang typed/racket
(ann (λ (v) (if (eq? v 'a) 'x 'y))
     (case→ (→ (U 'a) (U 'x))
            (→ (U 'a 'b) (U 'x 'y))))
```
### If you got an error message, please include it here.

Running the following variation segfaults:

``` racket
#lang typed/racket

(define fn (case-lambda
         #:∀ (A B ...)
         [([a : A])
          (ann a A)]
         [[rest : B ... B]
          (ann rest (List B ...))]))

(car (ann (fn 1) (List One))) ; => segfault
```

PR #618, opened 2017-09-30
fix #:rest and optional args

pnwamk:
Since we don't explicitly represent optional arguments in TR (i.e. we just use case-> for everything) we have converted this function type:

```racket
(->* (A) (B) #:rest C D)
```

into this function type:

```racket
(case-> (-> A C * D)
        (-> A B C * D))
```

which is incorrect and unsound (see https://github.com/racket/typed-racket/issues/614). This combined with some other subtle (probably related) decisions in our handling of type checking `case-lambda` forms has been needing some work for a while.

This PR aims to fix these issues. Now the function type

```racket
(->* (A) (B) #:rest C D)
```

will be converted internally into

```racket
(case-> (-> A D)
        (-> A B C * D))
```

I've been working to touch up all of the code affected by this so we no longer have issues with optional and rest args and case-lambdas. Things seem to be going well: I have a bunch of new test cases for what should and should not work, and all of these new tests are now passing. I just need to do more investigation to see if anything else is broken w/ these changes before this is ready to merge.

As always, feedback and comments welcome =)

### DIFF
--- a/typed-racket-lib/typed-racket/env/init-envs.rkt
+++ b/typed-racket-lib/typed-racket/env/init-envs.rkt
@@ -194,15 +194,15 @@
              ,(type->sexp t)
              (list ,@(map path-elem->sexp pth)))]
     [(Fun: (? has-optional-args? arrs))
-     (match-define (Arrow: fdoms rest *kws rng) (first arrs))
-     (match-define (Arrow: ldoms _ _ _) (last arrs))
+     (match-define (Arrow: fdoms _ kws rng) (first arrs))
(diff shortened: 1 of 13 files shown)


pnwamk:
@samth thoughts? I think it's a relatively straightforward fix. It doesn't change our arrows fundamentally, just fixes how we treat optional and rest args.

In other words, this PR fixes https://github.com/racket/typed-racket/issues/614 and https://github.com/racket/typed-racket/issues/432 (and probably a few others), but it does not fix all the case-lambda issues (e.g. https://github.com/racket/typed-racket/issues/413). After this gets merged I can make a PR for those.

PR #620, opened 2017-10-01
fix case lambda

pnwamk:
## Summary

This adds to (i.e. __depends on__) the changes in https://github.com/racket/typed-racket/pull/618 by:

+ cleaning up our lambda checking code more
+ fixing some case-lambda issues we've had (in particular how we handle giving types to case lambdas that have no expected type (i.e. that we are checking entirely in type synthesis mode))

This fixes https://github.com/racket/typed-racket/issues/413 (possibly more).

## Detailed description of general approach

In general, whenever we have a `case->` expected type we ensure that the body of the function checks successfully __at each type__ in the `case->`.

This also needs to happen when we are checking a `case-lambda`, however as we analyze each case, we may see that certain arrow types in the `case->` type are "unreachable" for a particular case-lambda clause (i.e. the arrow's domain would have been handled by a previous `case-lambda` clause already), and we thus need not verify that clause at that arrow type.

## Detailed description of the problem

Previously (i.e. until this PR), when we saw a `case-lambda` _without_ an annotation we would just check each `case-lambda` clause (each producing an arrow type) and then we would throw all the resulting arrows into a `case->` and be done with it.

That approach, however, does not do the aforementioned checking that is required. i.e. we __must__ check that all of the arrows in the resulting `case->` are valid arrow types for this `case-lambda`.

This PR fixes this, by going back and checking clauses at any arrow types they might be used at. It is careful not to re-check clauses at types they are already known to be (i.e. we make sure not to do the work twice for a given clause).

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
@@ -2,7 +2,7 @@
 
 (require "../utils/utils.rkt"
          racket/list syntax/parse syntax/stx
-         racket/match syntax/private/id-table racket/set
+         racket/match syntax/private/id-table
          racket/sequence
          (contract-req)
          (rep type-rep object-rep rep-utils)
@@ -41,6 +41,26 @@
            #:attr mapping (make-immutable-free-id-table)
            #:attr flag-mapping (make-immutable-free-id-table)))
 
+;; positional: (listof identifier?)
+;; rest: id or #f
+;; syntax: syntax? - the improper syntax list of identifiers
+;; (i.e. (append positional (or id '())) but syntax)
+(struct formals (positional rest syntax) #:transparent)
+
+(define (make-formals stx)
+  (let loop ([s stx] [acc null])
+    (cond
+      [(pair? s) (loop (cdr s) (cons (car s) acc))]
+      [(null? s) (formals (reverse acc) #f stx)]
+      [(pair? (syntax-e s)) (loop (stx-cdr s) (cons (stx-car s) acc))]
+      [(null? (syntax-e s)) (formals (reverse acc) #f stx)]
+      [else (formals (reverse acc) s stx)])))
+
+;; Currently no support for objects representing the rest argument
+(define (formals->objects f)
+  (for/list ([i (in-list (formals-positional f))])
+    (make-Path null i)))
+
 (define (expected-str tys-len rst arg-len rest-id)
   (format "Expected function with ~a argument~a~a, but got function with ~a argument~a~a"
           tys-len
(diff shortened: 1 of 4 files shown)


pnwamk:
This solution is more or less what @AlexKnauth described in https://github.com/stchang/typed-rosette/issues/5 as well, for the long term reader familiar with past issues in this space.

samth:
Does this fix #432 as well?

pnwamk:
#432 is fixed by https://github.com/racket/typed-racket/pull/618 (i.e. the previous function fixing PR that this one "depends on")

pnwamk:
#432 was an issue with out handling of optional and rest args, but was not affected by our generating of types for case-lambdas with no expected type

typed-racket-test/fail/apply-dots.rkt
```diff
@@ -7,7 +7,7 @@
                  w))
 
 (plambda: (a ...) ([z : String] . [w : Number *])
-          (apply (case-lambda: (([x : Number] . [y : Number ... a]) x)
```
samth: This case seems fine to me.

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -271,146 +296,222 @@
...
+          ;; If the empty function is expected, then don't error out
+          (not (null? expected-arrows)))
+     ;; TODO improve error message.
+     (tc-error/expr #:return (list (-Arrow null -Bottom #:rest Univ))
+                    "Expected a function of type ~a, but got a function with the wrong arity"
+                    (make-Fun expected-arrows))]
+    [else
+     (apply
+      append
+      (for*/list ([(f+b arrows) (in-assoc to-check-list)]
```
samth: maybe name `arrows` `expected-arrows`?

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -271,146 +296,222 @@
...
+                  [clause-formals (in-value (car f+b))]
+                  [clause-body (in-value (cdr f+b))]
+                  [orig-arrows (in-value (cond
+                                           [(assoc f+b orig-arrows) => cdr]
+                                           [else '()]))]
+                  [arrow (in-list arrows)])
```
samth: Note that checking each one separately is less complete than intersecting everything in `arrows`, but is sound. Maybe put a comment here to that effect.

typed-racket-test/succeed/apply-dots.rkt
```diff
@@ -12,12 +12,6 @@
           (apply (lambda: ([x : Number] . [y : Number *]) x)
                  1 w))
 
-(plambda: (a ...) ([z : String] . [w : Number *])
```
samth: This no longer works due to the incompleteness mentioned above, but we're not worrying about that right now. Maybe it should go in `xfail` or in a comment here.

PR #622, opened 2017-10-02
Handle subtyping with all-optional kw args better.

### LINKED ISSUE #608: map string-trim
6.10.1

### What program did you run?
`(map string-trim null)`

`'()`

```
Type Checker: Polymorphic function `map' could not be applied to arguments:
Types: (-> a b ... b c) (Listof a) (Listof b) ... b -> (Listof c)
       (-> a c) (Pairof a (Listof a)) #f * -> (Pairof c (Listof c))
Arguments: (->* (String) ((U Regexp String) #:left? Any #:repeat? Any #:right? Any) String) Null #f *
Expected result: AnyValues
 in: (map string-trim null)
```

samth:
Problem reported by @schackbrian2012. Closes #608.

### DIFF
--- a/typed-racket-lib/typed-racket/infer/infer-unit.rkt
+++ b/typed-racket-lib/typed-racket/infer/infer-unit.rkt
@@ -381,11 +381,15 @@
 
      (define s-seq (seq ss (rest->end s-rest)))
      (define t-seq (seq ts (rest->end t-rest)))
-     (and (null? s-kws)
-          (null? t-kws)
-          (% cset-meet
-           (cgen context s t)
-           (cgen/seq context t-seq s-seq)))]))
+     (and
+      ;; if all keywords are optional, then we can just treat
+      ;; them like they aren't there (or if there are none)
+      (for/and ([s-kw (in-list s-kws)])
+        (not (Keyword-required? s-kw)))
+      (null? t-kws)
+      (% cset-meet
+         (cgen context s t)
+         (cgen/seq context t-seq s-seq)))]))
 
 (define/cond-contract (cgen/flds context flds-s flds-t)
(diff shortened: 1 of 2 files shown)


typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -4204,6 +4204,7 @@
                    [else 42]))
                (void))
              -Void]
+       [tc-e (map string-trim (ann '("a" "b") (Listof String))) (-lst -String)]
        )
```
pnwamk: should we add a negative test (i.e. that it doesn't work when mandatory keywords are present?)

typed-racket-lib/typed-racket/infer/infer-unit.rkt
```diff
@@ -381,7 +381,11 @@
 
      (define s-seq (seq ss (rest->end s-rest)))
      (define t-seq (seq ts (rest->end t-rest)))
-     (and (null? s-kws)
+     (and (or (null? s-kws)
+              ;; if all keywords are optional, then we can just
```
pnwamk: Is the `(null? s-kws)` unnecessary? (i.e. the `for/and` returns true whenever `(null? s-kws)` is true).

typed-racket-test/unit-tests/typecheck-tests.rkt
samth: I don't think we should add negative tests for incompleteness.

pnwamk: isn't it possible to write a test that _should fail_ here because of a non-provided mandatory keyword argument? (sorry if I'm missing something)

samth: Ah, I understand now. Added.

PR #623, opened 2017-10-03
initial #:rest-star support

pnwamk:
This PR adds support for rest specifications that are the Kleene closure of some sequence.

e.g. functions like `hash` or `hash-set*` which require a cyclic rest argument with alternating key and value types will be able to be specified with something like the following:

```racket
(: hash-set* (All (K V) (->* ((HashTable K V)) () #:rest-star (K V)
                             (HashTable K V))))
(define (hash-set* h . k+vs)
  ...)
```

Thanks for comments/suggestions from @AlexKnauth, @bennn, and @samth; also thanks to @takikawa for previous work exploring some of these features.

If we later at more general rest argument support, this will simply be subsumed by that and hopefully add some tests and use cases that can help prove a more general implementation is correct and behaves as expected.

### DIFF
--- a/typed-racket-lib/typed-racket/env/init-envs.rkt
+++ b/typed-racket-lib/typed-racket/env/init-envs.rkt
@@ -264,7 +264,9 @@
      `(make-Opaque (quote-syntax ,pred))]
     [(Refinement: parent pred)
      `(make-Refinement ,(type->sexp parent) (quote-syntax ,pred))]
-    [(Mu-name: n b)
+    [(Mu-maybe-name: n (? Type? b))
+     `(make-Mu (quote ,n) ,(type->sexp b))]
+    [(Mu: n b)
      `(make-Mu (quote ,n) ,(type->sexp b))]
     [(Poly-names: ns b)
      `(make-Poly (list ,@(for/list ([n (in-list ns)])
@@ -324,6 +326,8 @@
        ,(and rest (type->sexp rest))
        (list ,@(map type->sexp kws))
        ,(type->sexp rng))]
+    [(Rest: tys )
+     `(make-Rest (list ,@(map type->sexp tys)))]
     [(RestDots: ty db)
      `(make-RestDots ,(type->sexp ty)
                      (quote ,db))]
--- a/typed-racket-lib/typed-racket/infer/infer-unit.rkt
+++ b/typed-racket-lib/typed-racket/infer/infer-unit.rkt
@@ -22,6 +22,7 @@
          "signatures.rkt" "fail.rkt"
          "promote-demote.rkt"
          racket/match
+         ;racket/trace
          (contract-req)
          (for-syntax
            racket/base
@@ -175,20 +176,25 @@
 ;; of them.
 (struct seq (types end) #:transparent)
 (struct null-end () #:transparent)
-(struct uniform-end (type) #:transparent)
+(define -null-end (null-end))
+;; ts is the pattern of the rest of the seq that can
+;; occur 0 or more times
+;; e.g. a rest argument of Num would just be (list Num)
+;;      a rest arg of (Num Str) would be (list Num Str)
+(struct star-end (ts) #:transparent)
 (struct dotted-end (type bound) #:transparent)
 
 (define (Values->seq v)
   (match v
-    [(Values: ts) (seq ts (null-end))]
+    [(Values: ts) (seq ts -null-end)]
     [(ValuesDots: ts dty dbound) (seq ts (dotted-end (-result dty) dbound))]
     [_ #f]))
 
 
 (define (List->end v)
   (match v
-    [(== -Null) (null-end)]
-    [(Listof: t) (uniform-end t)]
+    [(== -Null) -null-end]
+    [(Listof: t) (star-end (list t))]
     [(ListDots: t dbound) (dotted-end t dbound)]
     [_ #f]))
 
@@ -256,19 +262,27 @@
     [((seq ss (null-end))
       (seq ts (null-end)))
       (cgen/list context ss ts)]
-    ;; One is null-end the other is uniform-end
+    ;; One is null-end the other is star-end
     [((seq ss (null-end))
-      (seq ts (uniform-end t-rest)))
-     (cgen/list context ss (list-extend ss ts t-rest))]
-    [((seq ss (uniform-end s-rest))
+      (seq ts (star-end t-rest)))
+     (define ss-len (length ss))
+     (define ts-len (length ts))
+     #:return-unless (<= ts-len ss-len) #f
+     (define fewer-args (- ss-len ts-len))
+     (define cycle-len (length t-rest))
+     #:return-unless (zero? (remainder fewer-args cycle-len)) #f
+     (define repetitions (quotient fewer-args cycle-len))
+     (define new-ts (append ts (repeat-list t-rest repetitions)))
+     (cgen/list context ss new-ts)]
+    [((seq ss (star-end _))
       (seq ts (null-end)))
      #f]
-    ;; Both are uniform-end
-    [((seq ss (uniform-end s-rest))
-      (seq ts (uniform-end t-rest)))
-     (cgen/list context
-                (cons s-rest ss)
-                (cons t-rest (list-extend ss ts t-rest)))]
+    ;; Both are star-end
+    [((seq ss (star-end s-rest))
+      (seq ts (and t-end (star-end t-rest))))
+     (cgen/seq context
+               (seq (append s-rest ss) -null-end)
+               (seq (append t-rest ts) t-end))]
     ;; dotted below, nothing above
     [((seq ss (dotted-end dty dbound))
       (seq ts (null-end)))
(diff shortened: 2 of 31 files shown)


pnwamk:
Needs more tests, and probably some code cleanup now that things appear to be working.

AlexKnauth:
I remember a previous pull request by @takikawa about this used a `#:rest@` syntax or something like that, but generalized it to work with all list types, not just simple fixed cycles. It did this by making it more like the original rest arguments for the `->*` contract combinator. You would use it like:
```racket
(All (K V) (->* [(HashTable K V)] [] #:rest@ (EvenList K V) (HashTable K V)))
```
Where `EvenList` is defined as
```racket
(define-type (EvenList X Y)
  (List* X Y (EvenList X Y)))
```
This is more flexible because it's not limited to simple repeating cycles, as long as you could write the type of the list with some kind of recursive type you could do it.

I found the old pull request on the racket repository, it was made before the packages were split into separate repositories:
https://github.com/racket/racket/pull/564

pnwamk:
Yes, thanks for linking to that PR: there are definitely more general approaches than what is contained herein. 

This PR is ready today (i.e. for the next release likely, if we want). A more general approach could likely borrow from what @takikawa put together and---if it is useful---the contents of this PR, but wouldn't be fully ready for a while (I assume).

The more general approach still looks like it has some important unknowns:
1. there are no tests in that PR actually stressing the feature -- we don't know how feasible it actually is in general (although this PR still needs a few more tests at the moment, TBH)
2. in its present state, it seems to rely on match expanders for types in several places -- that's just not going to work in general, since you can write any type that is a subtype of `(Listof Any)` and we're going to have to be able to traverse arbitrarily deep into it to grab the _nth type_ inside of it in certain cases.

Out of curiosity, can we think of any rest argument patterns in general use that aren't a simple cycle? This PR requires the cycle be complete (i.e. the provided args modulo the cycle length must equal 0) -- it would be easy to also allow incomplete cycles.

Anyway -- I don't have strong feelings either way. If we want to add this (and maybe incomplete cycles) now I'm happy to do that. If we want to do something more general, I'm happy to scrap this as well.

The _real_ benefit of this PR was the bug fixes and cleaning up of `tc-lambda-unit.rkt` and some discussions with Sam about what our `case->` actually means, anyway XD (that and forcing me to learn more of what's going on in that file).

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -117,71 +112,110 @@
...
+            [(andmap type-annotation arg-list)
+             (get-types arg-list #:default Univ)]
+            [(eqv? 0 extra-arg-count) arg-tys]
+            [(negative? extra-arg-count) (take arg-tys arg-len)]
+            [else
+             (define tail-tys (match rst
+                                [(Rest: rst-tys)
+                                 (define rst-len (length rst-tys))
+                                 (for/list ([idx (in-range extra-arg-count)])
+                                   (list-ref rst-tys (remainder idx rst-len)))]
```
bennn: why not:
```
(for/list ([_ (in-range extra-arg-count]
           [ty (in-cycle rst-tys)])
  ty)
```

typed-racket-lib/typed-racket/env/init-envs.rkt
```diff
@@ -324,6 +324,8 @@
        ,(and rest (type->sexp rest))
        (list ,@(map type->sexp kws))
        ,(type->sexp rng))]
+    [(Rest: tys )
+     `(make-Rest (list ,@(map type->sexp tys)))]
```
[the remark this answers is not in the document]
pnwamk: no, we need the explicit list there since this is hackily generating syntax (otherwise the first type sexp would be in operator position instead of `list` in output)

typed-racket-lib/typed-racket/infer/infer-unit.rkt
```diff
@@ -175,20 +176,24 @@
 ;; of them.
 (struct seq (types end) #:transparent)
 (struct null-end () #:transparent)
-(struct uniform-end (type) #:transparent)
+(define -null-end (null-end))
+;; cycle is the pattern of the rest of the seq,
```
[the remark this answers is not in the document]
pnwamk: `star-end` is nice - cool

typed-racket-lib/typed-racket/infer/infer-unit.rkt
```diff
@@ -947,9 +975,16 @@
...
 (define (infer/vararg X Y S T T-var R [expected #f]
                       #:objs [objs '()])
-  (define new-T (if T-var (list-extend S T T-var) T))
   (and ((length S) . >= . (length T))
-       (infer X Y S new-T R expected #:objs objs)))
+       (let* ([fewer-ts (- (length S) (length T))]
+              [new-T (match T-var
+                       [(? Type? var-t) (list-extend S T var-t)]
+                       [(Rest: rst-ts)
+                        #:when (eqv? 0 (remainder fewer-ts (length rst-ts)))
```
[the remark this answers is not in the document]
pnwamk: Is there something wrong with eqv? It just seemed like the sensible comparator here to me ¯\_(ツ)_/¯. I don't need all the expresiveness `zero?` gives by any stretch, so I just grabbed good ol faithful eqv?

I'd rather not use `unsafe-fx=` in code that could easily change.

pnwamk:
w.r.t. the rest arg potions you described for command-line flags, I think that should expressible with just a `case->`, right?

pnwamk:
Can you say more about what you mean by "I vote we re-use `#:rest` ... so `#:rest Integer` is now a synonym for `#:rest [Integer]`"

I don't understand how we could make that change and not break some parsing we're currently doing (i.e. we parse that expression as a type currently... how can we tell if we should parse it as a type (the old way) or parse it as a sequence of types (the suggested new way). Perhaps I'm missing something to what you meant, though.

bennn:
> w.r.t. the rest arg potions you described for command-line flags, I think that should expressible with just a case->, right?

I don't think a case-lambda could handle all these calls:

```
> (raco-test "file.rkt")
> (raco-test 'drdr "file.rkt")
> (raco-test 'drdr 's "test" "file.rkt")
> (raco-test 's "stress-test" 'drdr 's "test" "file.rkt")
```

> Can you say more about what you mean by "I vote we re-use #:rest ...

In terms of syntax-parse 😄  the old rest pattern (IIUC) is something like

```
(~seq #:rest t:type)
```

I'm suggesting the new way be:

```
(~or (~seq #:rest ty:type)
     (~seq #:rest [tys:type ...]))
```

typed-racket-lib/typed-racket/infer/infer-unit.rkt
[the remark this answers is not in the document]
bennn: There's nothing technically wrong as long as `remainder`s always returning exact integers.

pnwamk:
Just to be clear, I said a `case->` not a `case-lambda`, and the two are not necessarily related (a `case->` is just a type that is a simple intersection of function types, a `case-lambda` is a funny kind of run-time value).

_Sorry if I'm coming across as overly pedantic for the sake of pedantry -- that's not my intent. I worry (perhaps too much) in the past it has been unclear or we have confused details about our functions and their types, and so I'm just trying to be extra clear about what we mean._

Yes, I see now. If we wanted to have a rest type that is expressing some arbitrary ordering of sequences of arguments, a la:

```racket
(Rec T (U (List* 'noarg-flag T)
          (List* 'arg-flag Integer T)
          (Listof String)))
```

We could not do that with just a cycle. But certainly we could write a function that accepts the input and then does argument checking internally. I suspect that even with general rest types, some of the restrictions you would ideally want (i.e. probably no duplicates of certain flags, etc) will not be expressible (at least any time soon) with a richer rest argument type.

Anyway, ramble ramble ramble. Sorry for being long winded -- thanks for the real world example! =)

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
pnwamk: good point! But `in-cycle` super slow dynamic dispatch yuckiness... so I wrote an `in-list-cycle` that is __significantly__ faster and achieves the same purpose. Now it's efficient and simple =)

The only sequence we're ever working with is lists anyways, so that should be a potentially useful macro elsewhere as well.

pnwamk:
Okay -- I think I applied most of Ben's comments (thanks @bennn!).

Assuming I didn't add any new bugs... I just need to add more tests and get @samth to look at it.

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -508,6 +508,19 @@
       (= (length kws) 1)
       (equal? kws (sort kws Keyword<?))))
 
+;; a Rest argument description
+;; tys: the cycle describing the rest args
+;; e.g.
+;; tys = (list Number) means all provided rest args
```
samth: Does this have to be `(list T)` instead of `T`?

typed-racket-test/fail/apply-dots.rkt
```diff
@@ -1,13 +1,13 @@
 #;
-(exn-pred 3)
+(exn-pred 2)
```
samth: Why did this change?

typed-racket-lib/typed-racket/private/parse-type.rkt
```diff
@@ -958,19 +958,35 @@
               (parse-type #'rng)
               : (-PS (attribute latent.positive) (attribute latent.negative))
               : (attribute latent.object)))]
+      ;; like ->* below but w/ a #:rest-pat present
       [(:->*^ (~var mand (->*-args #t))
               (~optional (~var opt (->*-args #f))
                          #:defaults ([opt.doms null] [opt.kws null]))
-              rest:->*-rest
+              #:rest* (rest-types-stx:non-keyword-ty ...)
```
samth: Are there any tests/docs for this new form?

typed-racket-lib/typed-racket/rep/type-rep.rkt
pnwamk: that makes handling any rest args (i.e. uniform or more interesting cycles) all able to use the same code

typed-racket-lib/typed-racket/private/parse-type.rkt
pnwamk: no -- wanted feedback on what it should actually look like for user syntax first.

I don't want to just re-use `#:rest` because I worry that will lead to ambiguous parsing, but I don't have strong feelings on what to use instead. `rest*` as in Kleene-star seems sensible to me. I dunno.

typed-racket-lib/typed-racket/rep/type-rep.rkt
samth: Ah, I see now

typed-racket-lib/typed-racket/private/parse-type.rkt
samth: `#:rest*` seems fine, or `#:rest-sequence` or perhaps something else?

pnwamk:
Update! Things are working well it seems. I believe all that is left before possibly being ready to merge is the following:

1. add parsing tests for `#:rest-star`
2. add an integration test in `succeed/` that uses the features to implement a few useful functions (e.g. maybe show that you can use it to implement your own version of `hash` and `hash-set*` and it all works as expected)
3. add to the documentation of TR's `->*` to describe `#:rest-star`

typed-racket-test/fail/apply-dots.rkt
pnwamk: This was a silly mistake on my part. The file still errors w/ 3 type checker errors (I had changed the file, forgot I changed it, and mindlessly updated the exn-pred line (doh!))

It's fixed now. Same behavior as before

pnwamk:
I think it's ready(ish) for further review... possibly merging? Hard to know.

All the bullet points I had wanted to address have at least been addressed to some degree.

samth:
I think this is good to go.

samth:
Can we get this merged?

Issue #625, opened 2017-10-06
`msgpack` failure


bennn:
cc @samth

### Description

The `msgpack` package fails to build with the current TR release candidate. The source compiles, but test programs written in `#lang racket` do not. For example:

```
#lang racket/base
(require msgpack)
unpack
```

Error message:

```
foo.rkt:3:0: Type Checker: could not convert type to a contract;
 required a flat contract but generated a chaperone contract
  identifier: unpack
  type: (-> Input-Port Packable)
  in: unpack
```

The type for unpack is:

```
(: unpack (-> Input-Port
              (U Void Boolean Integer Real String Bytes (Vectorof Packable) (Listof Packable) (HashTable Packable Packable) Ext)))
```

where `Packable`  is:

```
(define-type Packable
  (U Void
     Boolean
     Integer
     Real
     String
     Bytes
     (Vectorof Packable)
     (Listof Packable)
     (HashTable Packable Packable)
     Ext))
```

### Minimal Example

The TR error message seems to be due to the `HashTable` type in `Packable`; here's a tiny program the raises a similar exception:

```
#lang racket/base

(module t typed/racket/base
  (provide f)

  (define-type T
    (HashTable (Vectorof Integer) Symbol))

  (: f (-> Symbol T))
  (define (f x)
    (hash '#(1) x)))

(require 't)

f
```

Error message (on Racket v6.10.1 or earlier):

```
bar.rkt:15:0: Type Checker: could not convert type to a contract;
 required a flat contract but generated a chaperone contract
  identifier: f
  type: (-> Symbol T)
  in: f
```

The program runs if the type for `f` is `(-> Symbol (U Integer T))` because that type gets converted to a contract like `(-> Symbol any/c)`

### Diagnosis

Since `(hash (vector 1) 'a)` is valid, the problem is likely the constraint in [`static-contracts/combinators/structural.rkt`](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt) that hashtable keys must be flat:

```
  ((hash/sc (#:invariant #:flat) (#:invariant)) hash/c #:chaperone)
```

pnwamk:
Was the motivation for requiring flat key contracts perhaps to not change the behavior of `eq?` based hash tables?

samth:
I think we can fix that by distinguishing `equal?` based tables. Will that fix `msgpack`?

bennn:
[...]

PR #626, opened 2017-10-06
relax constraint on hash contracts (was: fix msgpack failure)

bennn:
#625 

@samth I guess this is how/where to do the fix,
but I'm not sure what the error message should say.

(Writing some tests now.)


bennn:
cc @pnwamk

typed-racket-lib/typed-racket/utils/hash-contract.rkt
```diff
@@ -1,8 +1,18 @@
...
+(define (immutable-hash/c k v)
+  (tr-hash/c k v #:immutable #t))
+
+(define (weak-hash/c k v)
+  (and/c hash? hash-weak? (tr-hash/c k v #:immutable #f)))
+
+(define (tr-hash/c k v #:immutable [immutable 'dont-care])
+  (if (flat-contract? k)
+    (hash/c k v #:immutable immutable)
+    (and/c hash? hash-equal? (hash/c k v #:immutable immutable))))
```
samth: I think we need to do something to produce a good error message in this case if `hash-equal?` is `#f`, but I don't have a good suggestion.

Also, it would be even better if we could make this a contract that rejected the non-flat parts of `k` (probably we'd have to do that split in the static contract). But this fix, assuming it doesn't break things, is good for the release.

bennn: Can we just let `hash/c` raise the error message for the release?

The only improvement I can think of is printing the key, e.g. "expected a hash-equal? because of key (vectorof foo?)"

samth:
How about just, "expected an equal-based hash because the key contract is not a flat contract"?

bennn:
Example with the current message:

```
contract violation
  expected: "hash-equal? (because the key contract is not a flat contract)"
  given: '#hasheqv()
  in: the 1st conjunct of
      the 2nd conjunct of
      the 1st argument of
      (->
       (and/c
        hash?
        (and/c
         "hash-equal? (because the key contract is not a flat contract)"
         (hash/c (vectorof Symbol) Symbol)))
       any)
```

samth:
LGTM

Issue #628, opened 2017-10-06
sc optimizer fails to simplify (-> flat-recursive-contract)


bennn:
Similar to #625 , I just found an error in some gradually-typed versions of the [zordoz](https://github.com/nuprl/gradual-typing-performance/tree/master/benchmarks/zordoz.6.5/typed) benchmark.

### What version of Racket are you using?
6.10.1

### What program did you run?

```
#lang racket/base

(module t typed/racket/base
  (provide f)

  (define-type Spec
    (-> (U Spec String)))

  (: f (-> Spec))
  (define (f)
    (λ () "hello")))

(require 't)

f
```

### What should have happened?

No error

### If you got an error message, please include it here.

```
recursive-contract: contract violation
  expected: flat-contract?
  given: (-> (or/c String g5))
```

### Probable Cause

It's an issue instantiating recursive contracts, the type above makes a `(recursive-contract .... #:flat)`. In this case it shouldn't be flat. Probably a bug with `or/sc`.

bennn:
Small update: there's no problem if either:
- the contract contains a function with mandatory arguments
- the contract contains a function that returns a non-flat value

I still don't really understand the problem.
Somehow the static-contracts code thinks its making a flat recursive contract, but puts an `->` inside.

I guess the optimizer needs to do whatever `static-contracts/combinators/function.rkt` is doing to tell when a function doesn't need a contract.

```
#lang racket/base

(module t typed/racket/base
  (provide f)

  (define-type Spec
    ;; Now with non-zero arity
    (-> Symbol (U Spec String)))

  (: f (-> Spec))
  (define (f)
    (λ (s) "hello")))

(require 't)

f
;; No error
```

bennn:
Just posted 2 ideas for resolving this.

1. Improve the optimizer, so it can see that `(Rec Spec (-> (U String Spec)))` is a flat contract (need to solve recursive contract constraints in the optimizer)
2. Change the constraints on `function/sc`, because it's not currently true that a 0-arity function coming from the typed side that returns flat values will get a flat contract from the static contract instantiator

I'd rather merge Option 1 (#631). It'll do a little more work at compile-time, but can safely turn more contracts into `any/c`s

bennn:
Note this is not really a "new error". The example program fails on 6.10.1.

samth:
The example program fails on 6.10.1, but zordoz does not fail on 6.10.1, right?

bennn:
Right

samth:
[...]

PR #631, opened 2017-10-14
static-contracts: improve optimizer's test for flat scs

bennn:
This is one idea for fixing #628 .

I'm not happy about re-computing `recursive-kinds` in the optimizer,
but I don't see how to tell whether a `name/sc` is flat without it,
and I didn't see an easy way to share the optimizer's `recursive-kinds` with `instantiate.rkt`

- - -

Improve the static contract optimizer's ability to recognize flat contracts:

1. get the kind of all recursive, named contracts
2. any `name/sc` with `'flat` kind is a flat contract

Step 1 is repeated later in the `instantiate` pass.

PR #632, opened 2017-10-14
static-contracts: functions always make chaperone contracts

bennn:
This is one idea for fixing #628 

- - -

change `sc->constraints` for `function/sc` to always add a `chaperone` constraint,
because 0-arity functions with flat ranges do not always make flat contracts

(the contracts are only flat when the function contract is optimized to any/c)


bennn:
#631 is better

PR #637, opened 2017-10-22
Revert "static-contracts: less or/sc optimization"

samth:
This reverts commit 9df037b0f6e6b2df3cc0993631bb2870367e291a.

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/optimize.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/optimize.rkt
@@ -178,6 +178,7 @@
 ;;  for optimizing the sub-contracts of the given `sc`.
 (define (update-side sc side)
   (match sc
+   #;
    [(or/sc: scs ...)
     #:when (not (andmap flat-terminal-kind? scs))
     (weaken-side side)]
--- a/typed-racket-test/unit-tests/static-contract-optimizer-tests.rkt
+++ b/typed-racket-test/unit-tests/static-contract-optimizer-tests.rkt
@@ -178,14 +178,6 @@
     (check-optimize (or/sc none/sc none/sc)
       #:pos any/sc
       #:neg none/sc)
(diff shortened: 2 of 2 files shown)


bennn:
`trusted-side-reduce` should be:

```
(define (trusted-side-reduce sc)
  (match sc
    [(->/sc: mand-args opt-args mand-kw-args opt-kw-args rest-arg (list (any/sc:) ...))
     (function/sc #t mand-args opt-args mand-kw-args opt-kw-args rest-arg #f)]
    [(arr/sc: args rest (list (any/sc:) ...))
     (arr/sc args rest #f)]
    [(none/sc:) any/sc]
    [(app sc-terminal-kind 'flat) any/sc]
    [else sc]))
```

samth:
@bennn Can you have a look at this, and see if it fixed the zordoz issue?

samth:
@racket/release This is the change I'd like to make to Typed Racket for the release (and only for the release).

bennn:
Yes, this fixes the `zordoz` issue.
I'm able to run all gradually-typed versions of zordoz without error. (On master, e.g., `configuration10111` errors.)

Attached is all configurations --- to re-test, compile & run `main.rkt` in each directory matching `benchmark/configuration*/`.
[zordoz.tar.gz](https://github.com/racket/typed-racket/files/1415060/zordoz.tar.gz)

samth:
I've also tested that #625 is still fixed with this change.

samth:
@bennn Also, can you take a look at the code and confirm that you think this does the right thing?

bennn:
Besides that one "side channel" that we can now close for the release. I think this looks good and does the right thing.

samth:
I'd prefer to make the minimum changes needed to fix the issue. Will leaving that case in make anything worse than it was in 6.10.1?

bennn:
There's a small chance that if we leave it in, someone will write a program that works on 6.11 but fails on 6.12 and 6.10.

bennn:
^ I think that chance is very small, but figured it'd be good to remove it

PR #646, opened 2017-11-06
Avoid wrapping 5 when contracted as (Sequenceof Integer).

samth:
Closes racket/math#13

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -10,7 +10,7 @@
  (env type-name-env row-constraint-env)
  (rep core-rep rep-utils free-ids type-mask values-rep
       base-types numeric-base-types)
- (types resolve utils printer match-expanders union)
+ (types resolve utils printer match-expanders union subtype)
  (prefix-in t: (types abbrev numeric-tower subtype))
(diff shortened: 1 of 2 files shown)


typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -526,6 +526,11 @@
        [(? Fun? t) (t->sc/fun t)]
        [(? DepFun? t) (t->sc/fun t)]
        [(Set: t) (set/sc (t->sc t))]
+       [(Sequence: (list t))
+        #:when (subtype t:-Nat t)
+        ;; sequence/c is always a wrapper, so avoid it when we just have a number
+        (or/sc (flat/sc #'exact-nonnegative-integer?)
```
bennn: We should merge #631 first, because with this change and without that PR, every `(Sequence T)` type in a "trusted positive position" is going to optimize to an `any/c` contract.

```
(or/sc (flat/sc ....) (sequence/sc ....))
=>
(or/sc any/c (sequence/sc ....))
=>
any/c
```

bennn: (Even though sequences are read-only, this is a problem if `T` is something like `(Boxof Int)`)

PR #650, opened 2017-11-09
update check-duplicates with #:default kw arg

pnwamk:
=) Adds the new keyword are for `check-duplicates` (see https://github.com/racket/racket/pull/1840)

=( Will break any programs using `inst` and `check-duplicates`

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -3323,10 +3323,22 @@
    (->key (-lst a) (-> b b -Boolean) #:key (-> a b) #f #:cache-keys? -Boolean #f (-lst a)))))
 (check-duplicates
  (-poly
-  (a b)
+  (a b c)
   (cl->*
-   (->optkey (-lst a) ((-> a a Univ)) #:key (-> a a) #f (-opt a))
-   (->optkey (-lst a) ((-> b b Univ)) #:key (-> a b) #f (-opt a)))))
+   (->optkey (-lst a) ((-> a a Univ))
+             #:key (-> a a) #f
+             (-opt a))
+   (->optkey (-lst a) ((-> b b Univ))
+             #:key (-> a b) #f
+             (-opt a))
+   (->optkey (-lst a) ((-> a a Univ))
+             #:key (-> a a) #f
+             #:default (-> c) #f
+             (Un a c))
+   (->optkey (-lst a) ((-> b b Univ))
+             #:key (-> a b) #f
+             #:default (-> c) #f
+             (Un a c)))))
 (remove-duplicates
  (-poly
   (a b)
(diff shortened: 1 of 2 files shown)


pnwamk:
Backwards compatible alternatives?

1. have the default argument be a non-polymorphic type (e.g. `Any`)
2. ....?

stamourv:
Thanks for fixing this! And sorry I didn't see it coming.

pnwamk:
Okay -- I've been playing around with the type to see if it can work to just have two type variables. I've got this at the moment:

```racket
(All (a b)
     (case->
      (->* ((Listof a)) ((-> a a Any) #:key (-> a a)) (U False a))
      (->* ((Listof a)) ((-> a a Any) #:default b #:key (-> a a)) (U a b))
      (->* ((Listof a)) ((-> b b Any) #:key (-> a b)) (U False a))
      (->* ((Listof a)) ((-> b b Any) #:default a #:key (-> a b)) a)))
```

This should be backwards compatible and cover most use cases. The only downside I think is the fourth clause of the `case->`, it forces the list and default type to be the same `a`, and then the `#:key` function has to be able to make any of those `a`'s to a `b`... this may not actually be big issue in practice... I'm not sure.

pnwamk:
@samth when you have a chance if you want to spot check this to make sure I'm not crazy =)

samth:
This doesn't handle the procedure case, right? I think that's unsound.

pnwamk:
Ah, right. Thanks.

pnwamk:
Would another option be to use a type with 3 type variables (as would be natural now with the new interface) and make `inst` do "something sensible" when users provide fewer types to `inst` than a function expects? (maybe it chooses `Any` or `Bottom` for the types not provided?)

That might allow for evolution of interfaces like this a little better (although perhaps there are some complications I'm not thinking of at the moment)

pnwamk:
i.e. this type will work, but it does not seem ideal (you really want 3 type variables)... is there something we can do that doesn't bind us to an suboptimal type for the function?

samth:
I think `inst` using `Any` for extra type variables is probably the right thing.

pnwamk:
Okay -- I'll look into making that tweak and then giving a proper type to
check-duplicates (later today or this weekend).

samth:
Let's merge something for the moment to fix the tests.

pnwamk:
okay -- barring any typos, this will stop the drdr tests from failing by adding a third type variable and 2 more cases to the type of `check-duplicates`:

```racket
(All (a b c)
      (case->
       (->* ((Listof a)) ((-> a a Any) #:key (-> a a)) (U False a))
       (->* ((Listof a)) ((-> b b Any) #:key (-> a b)) (U False a))
       (->* ((Listof a))
            ((-> a a Any) #:default (-> c) #:key (-> a a))
            (U a c))
       (->* ((Listof a))
            ((-> b b Any) #:default (-> c) #:key (-> a b))
            (U a c))))
```

and another PR in a bit will make `inst` assign `Any` to unspecified type variables when too few are provided.

samth:
Great.

PR #651, opened 2017-11-10
partial inst

pnwamk:
In addition to potentially being convenient, this helps us not break backwards compatibility when APIs add features that require additional polymorphic variables to properly type.

### DIFF
--- a/typed-racket-lib/typed-racket/types/utils.rkt
+++ b/typed-racket-lib/typed-racket/types/utils.rkt
@@ -17,10 +17,12 @@
 (define (instantiate-poly t types)
   (match t
     [(Poly: ns body)
-     (unless (= (length types) (length ns))
+     (unless (<= (length types) (length ns))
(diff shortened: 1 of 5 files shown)


stamourv:
Very good idea.

typed-racket-lib/typed-racket/types/utils.rkt
```diff
@@ -17,10 +17,14 @@
 (define (instantiate-poly t types)
   (match t
     [(Poly: ns body)
-     (unless (= (length types) (length ns))
+     (unless (<= (length types) (length ns))
        (int-err "instantiate-poly: wrong number of types: expected ~a, got ~a"
                 (length ns) (length types)))
-     (subst-all (make-simple-substitution ns types) body)]
+     ;; if fewer types were provided than (length ns), we use Any as the default type
+     (let ([types (if (= (length types) (length ns))
```
samth: Can't we remove this `if`?

PR #652, opened 2017-11-10
Start work on optionally disabling extflonum tests.

samth:
cc @jackfirth


jackfirth:
When they're not available, logging a message (with a logger, not to stdout) stating that the tests were skipped might be nice. Maybe `rackunit/log` should supply first class support for counting skipped tests. But that can be another issue.

samth:
Closing in favor of #841

Issue #653, opened 2017-11-17
define/match ignores function annotations


bennn:
https://groups.google.com/d/msg/racket-users/vyXR5_aT2TY/rtrjUaX-CAAJ

### What version of Racket are you using?

6.11.0.1

### What program did you run?

```
#lang typed/racket


(define-type Reading (Vector Integer Integer))


(: close? (->* (Reading Reading) (Integer) Boolean))
(define/match (close? r1 r2 [err 1])
  [((vector a b) (vector c d) e)
    (and (= a (+ c e)) (= b (+ d e)))])
```

### What should have happened?

The program should typecheck.

Seems like TR is trying to infer a type for `close?` instead of checking that it matches the type annotation.

### If you got an error message, please include it here.

```
foo.rkt:10:17: Type Checker: type mismatch
  expected: Number
  given: Any
  in: c
  compilation context...:
   /home/ben/foo.rkt
  context...:
   /home/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-lib/typed-racket/utils/tc-utils.rkt:133:21: for-loop
   /home/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-lib/typed-racket/utils/tc-utils.rkt:123:0: report-all-errors
   /home/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:348:0: type-check
   /home/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:591:0: tc-module
   /home/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-lib/typed-racket/tc-setup.rkt:94:0: tc-module/full
   /home/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-lib/typed-racket/typed-racket.rkt:23:4
   /home/ben/code/racket/fork/racket/collects/compiler/cm.rkt:363:0: compile-zo*
   /home/ben/code/racket/fork/racket/collects/compiler/cm.rkt:572:26
   /home/ben/code/racket/fork/racket/collects/compiler/cm.rkt:564:42
   /home/ben/code/racket/fork/racket/collects/compiler/cm.rkt:635:0: compile-root
   /home/ben/code/racket/fork/racket/collects/compiler/cm.rkt:737:4
[...]

PR #654, opened 2017-11-23
add typed/racket/match

bennn:
Simple copy/paste fix for #653 . I didn't see a better way / how to use `opt-lambda-property`.

### DIFF
--- a/typed-racket-lib/typed/racket.rkt
+++ b/typed-racket-lib/typed/racket.rkt
@@ -3,13 +3,16 @@
 (require typed/racket/base racket/require
          (subtract-in racket typed/racket/base racket/contract
                       typed/racket/class
+                      typed/racket/match
                       typed/racket/unit)
          typed/racket/class
+         typed/racket/match
          typed/racket/unit
 	 (for-syntax racket/base))
 (provide (all-from-out racket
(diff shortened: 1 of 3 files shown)


samth:
I think this is a bad idea. We should avoid copy and paste like this unless
absolutely necessary.

Can we just add a simple syntax property to lambda in racket/base?

bennn:
I'm not sure how the property would work. Do you mean `opt-lambda-property` and `kw-lambda-property`? What values do those properties expect?


Another thing I thought about (but couldn't get to work) was having `typed/racket/match` expand to `define/match` with a property attached. Something like:

```
(define-syntax (-define/match stx)
  (syntax-property #`(define/match . #,(cdr (syntax-e stx))) ??? ???))
```

samth:
Yes, I'm suggesting that `typed/racket` should just recognize a property added by `lambda` and/or `define` itself. I think the TR properties don't have any interesting value; they're just set to `#t`.

jackfirth:
Question from the peanut gallery: would it solve this problem if `define` and/or `lambda` were syntax parameters?

samth:
A syntax parameter such as `current-lambda` that `match` could use would indeed be helpful here.

PR #657, opened 2018-01-21
unsafe-provide: unsafely provide other modules' exports

bennn:
Change `unsafe-provide` to un-rename any rename transformers that its asked to provide.

This means one typed module can safely provide some identifiers and, later, another typed module can import those identifier and UNSAFELY provide them.

### DIFF
--- a/typed-racket-compatibility/info.rkt
+++ b/typed-racket-compatibility/info.rkt
@@ -11,4 +11,4 @@
 
 (define pkg-authors '(samth stamourv))
 
-(define version "1.7")
+(define version "1.8")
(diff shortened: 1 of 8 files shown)


samth:
What's the status here? Also, this should probably have docs changes, and should definitely have a `history` note.

bennn:
This is in good shape, it implements what I intended.

Added a small change to the docs, a history note, and bumped the version (maybe the version bump isn't necessary?)

bennn:
ok to merge?

samth:
Yes.

PR #661, opened 2018-02-08
Add some missing numeric type names, and exn:fail:contract:variable-id

samth:
Thanks to @rfindler for noticing these.


samth:
@stamourv Is there a reason we didn't add these originally?

stamourv:
Not that I can think of.

PR #673, opened 2018-02-28
fix invalid reference to nonexisting nested attribute

rmculpepper:
This is an error in the new syntax implementation (racket/racket#1803)


samth:
This seems fine, but why doesn't it work anymore?

rmculpepper:
Because `syntax` will get `template`'s stricter error checking, intended to catch things like this.

More precisely, if `x.y` occurs in a template and `x` is bound as an attribute but `x.y` is not, then an error is raised, because the programmer probably intended to refer to a nested attribute of `x`.

PR #684, opened 2018-03-25
bugfix: don't let current-orig-stx be #false

### LINKED ISSUE #658: internal error (but still running 6.11.04)

 6.11.0.4--2017-12-18(-/f) [3m].

### What program did you run?

#lang typed/racket

(struct Node [{reachable : Edge}])
(struct Edge [{to : Symbol} {cost : Positive-Integer}])

(define-type Graph [Immutable-HashTable Symbol Node])

(: make-graph (Node *-> Graph))
(define (make-graph . lo-node)
  (make-immutable-hash))

TR should have signaled a parser error for the type. 

Internal Typechecker Error: erroneous syntax was not a syntax object: #f
 (error message: type cannot be applied
  type: Node
  arguments...: (Error Graph))
while typechecking:
#f
originally:
#f

bennn:
Make sure `types/resolve.rkt` does not set `current-orig-stx` to `#f`.

Closes #658

### DIFF
--- a/typed-racket-lib/typed-racket/types/resolve.rkt
+++ b/typed-racket-lib/typed-racket/types/resolve.rkt
@@ -146,9 +146,10 @@
 
 
 (define (resolve-app rator rands [stx #f])
-  (parameterize ([current-orig-stx (or stx (current-orig-stx))]
+  (define orig-stx (or stx (current-orig-stx)))
+  (parameterize ([current-orig-stx orig-stx]
                  [already-resolving? #t])
-    (resolve-app-check-error rator rands stx)
+    (resolve-app-check-error rator rands orig-stx)
     (match rator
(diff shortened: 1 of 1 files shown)


bennn:
With this PR, running the program in #658 gives this error:

```
foo.rkt:9:0: Type Checker: cannot apply a non-polymorphic type
  type: Node
  arguments: (Error Graph)
  in: (define (make-graph . lo-node) (make-immutable-hash))
  compilation context...:
   /Users/..../foo.rkt
  location...:
   foo.rkt:9:0
  context...:
```

Thank you to `racket/contract` for helping me debug this problem. Once I figured out the `current-orig-stx` parameter was getting a bad value, I recompiled the library with a `parameter/c`, re-ran, and looked at the blame error.

PR #690, opened 2018-04-09
add special types for racket/private/template

### LINKED ISSUE #688: missing type for identifier t-resyntax
The last [two](https://github.com/racket/typed-racket/commit/c526f222c270d04fc2ec88bdb0737dc6338b7871) [builds](https://github.com/racket/typed-racket/commit/d2771a0566e5f572c9ee9baf3622e503aab12318) on master show this error message:

```
$ if [ $TESTS = "unit" ]; then racket -l typed-racket-test -- --unit; fi
--------------------
tc-expr tests > Typechecker tests > Unit Tests > Typed Racket Tests > 3473 (syntax->datum (quasisyntax ((unsyntax (lambda (x) x)))))
ERROR
Type Checker: missing type for identifier;
  identifier: t-resyntax
  from module: template.rkt
  in: t-resyntax
--------------------
```
Looks like this is from merging `syntax/parse/experimental/template` [(commit)](https://github.com/racket/racket/commit/8d607b83f9966a4e3d3ae269dbc9960f3810d7ae) --- Typed Racket needs to add type annotations for the new implementation.

bennn:
Add type definitions for the new 'Guide' data structure.

@rmculpepper can you double-check these? 

Closes #688


samth:
Thanks for getting to this!

bennn:
@stamourv I think this test failure means there's something wrong with the optimizer's dead-code elimination.

```
The command "if [ $TESTS = "int"  ]; then racket -l typed-racket-test -- --int; fi" exited with 0.
334.30s$ if [ $TESTS = "int"  ]; then racket -l typed-racket-test -- --opt; fi
--------------------
Log Comparison for dead-case-lambda.rkt >  > Optimization Tests > Typed Racket Tests > Unnamed test
FAILURE
name:        check-equal?
location:    typed-racket-test/optimizer/run.rkt:28:6
expression:  (check-equal? log expected-log)
params:
  '(("TR opt: dead-case-lambda.rkt 3:7 (case-lambda (() (void)) ((d) (void)) ((d . rst) (void))) -- dead case-lambda branch"
     "TR opt: dead-case-lambda.rkt 6:10 (d . rst) -- dead case-lambda branch")
    ("TR opt: dead-case-lambda.rkt 4:10 () -- dead case-lambda branch"
     "TR opt: dead-case-lambda.rkt 6:10 (d . rst) -- dead case-lambda branch"))
actual:
  '("TR opt: dead-case-lambda.rkt 3:7 (case-lambda (() (void)) ((d) (void)) ((d . rst) (void))) -- dead case-lambda branch"
    "TR opt: dead-case-lambda.rkt 6:10 (d . rst) -- dead case-lambda branch")
expected:
  '("TR opt: dead-case-lambda.rkt 4:10 () -- dead case-lambda branch"
    "TR opt: dead-case-lambda.rkt 6:10 (d . rst) -- dead case-lambda branch")
--------------------
165 success(es) 1 failure(s) 0 error(s) 166 test(s) run
Typed Racket Tests did not pass.
```

Here's the function (from `optimizer/tests/dead-case-lambda.rkt`); it looks like the first case isn't getting added to the `lambda-dead-table`, and this seems to happen on the `master` branch too.

```
#lang typed/racket
(procedure-arity
  (ann (case-lambda
         [() (void)]
         [(d) (void)]
         [(d . rst) (void)])
          (Any -> Any)))
```

rmculpepper:
`t-const` and `t-var` are actually bound to `#%expression`, so they shouldn't need types. Similarly, `t-list` and `t-list*` are bound to `list` and `list*`.

`t-append` is essentially `append` except with a shortcut if the second list is empty, but beware that it allows improper lists. `t-orelse` is a macro; it's run-time helper `t-orelse*` takes two thunks and returns the result of one of them. I think in practice the return type is either `Syntax` or `(Listof Syntax)`.

Here are the types of functions that I think occur in expanded code:

    Stxish = '() | Syntax | (cons Syntax Stxish)
    t-append : (Listof Syntax) Stxish -> Stxish
    t-resyntax : (U Syntax #f) Syntax Any -> Syntax
    t-relocate : Syntax (U Syntax #f) -> Syntax
    t-orelse* : forall a, (-> a) (-> a) -> a
    t-struct : Any (Listof Syntax) -> Syntax
    h-splice : Any (U Syntax #f) (U Syntax #f) -> (Listof Syntax)
    t-subst : (U Syntax #f) Syntax X Y ... -> Syntax
      where X, Y are somewhat complicated
    check-same-length : (U Syntax #f) (U Syntax #f) (Listof Any) -> Void

AlexKnauth:
What does `make-template-identifier` do?

samth:
It solves the problem with non-provided identifiers that your other issue
mentioned.

samth:
Can we merge this soon? What do we still need?

stamourv:
@bennn: Re dead case-lambda branch: Looking at the test results, I think it's just a different source location; the test itself is still morally passing. So it should be ok to just fix the test to expect the new output, I think.

bennn:
Updated with Ryan's types and changed the failing `case-lambda` test.

(Vincent after looking again, I agree the test is morally passing. Probably we should file an issue for about the changed source location.)

PR #694, opened 2018-04-11
Fix type of parse-command-line

### LINKED ISSUE #692: Incorrect & incomplete type for `parse-command-line`
This is an issue to track the fact that the type of `parse-command-line` from [racket/cmdline](http://docs.racket-lang.org/reference/Command-Line_Parsing.html#%28def._%28%28lib._racket%2Fcmdline..rkt%29._parse-command-line%29%29) is both incorrect and incomplete, as noted on the [email list](https://groups.google.com/d/topic/racket-users/7hL-zpOdaT0/discussion). The problem also affects the expansion of the `command-line` macro.

The current type reported for `parse-command-line` is:
```racket
(All (b a ...)
      (-> Path-String
          (U (Listof String) (Vectorof String))
          (Listof
           (U (Pairof 'ps (Listof String))
              (Pairof
               (U 'final 'help-labels 'multi 'once-any 'once-each)
               (Listof (Listof Any)))))
          (-> Any a ... a b)
          (Listof String)
          b))
```

The problems are:

1. Most significantly, the type for the `'help-labels`/`#:help-labels` specification is wrong: the `cdr` of the pair for such entries in the table must be `(Listof String)` (like `'ps` currently), not `(Listof (Listof Any))`.
2. Similarly, the `'usage-help`/`#:usage-help` specification is not supported by the current type. It should be handled the same way as `'ps` and `'help-labels`.
3. The current type also does not support the optional "help-proc" or "unknown-proc" arguments accepted by `parse-command-line`.

@samth identified https://github.com/racket/typed-racket/blob/c526f222c270d04fc2ec88bdb0737dc6338b7871/typed-racket-lib/typed-racket/base-env/base-env.rkt#L3084 as the spot that needs to be corrected.

The following program, which fails to typecheck, illustrates the `'usage-help` problem:
```racket
#lang typed/racket

(parse-command-line
 "example"
 #("-h")
 '([usage-help "here is some usage help"])
 void
 null)
```

This program should fail to typecheck, but currently, because of the `'help-labels` problem, typechecking succeeds but `parse-command-line` raises a runtime error:
```racket
#lang typed/racket

(parse-command-line
 "example"
 #("-h")
 '([help-labels ("this is a help label")])
 void
 null)
```
Conversely, this program should typecheck and run successfully, but it currently fails to typecheck because of the `'help-labels` problem:
```racket
#lang typed/racket

(parse-command-line
 "example"
 #("-h")
 '([help-labels "this is a help label"])
 void
 null)
```

LiberalArtist:
1. Fix incorrect type for `'help-labels` specification

2. Add type for `'usage-help` specification

3. Support optional "help-proc" and "unknown-proc" arguments

4. Use more specific types for "finish-proc" and the flag specs

Closes #692

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -3082,15 +3082,34 @@
 
 ;; Section 15.9 (racket/cmdline)
 [parse-command-line
- (let ([mode-sym (one-of/c 'once-each 'once-any 'multi 'final 'help-labels)])
-   (-polydots (b a)
-              (cl->* (-Pathlike
-                      (Un (-lst -String) (-vec -String))
-                      (-lst (Un (-pair mode-sym (-lst (-lst Univ)))
-                                (-pair (-val 'ps) (-lst -String))))
-                      ((list Univ) [a a] . ->... . b)
-                      (-lst -String)
-                      . -> . b))))]
+ (let ([mode-sym (one-of/c 'once-each 'once-any 'multi 'final)]
+       [label-sym (one-of/c 'ps 'help-labels 'usage-help)])
+   (-polydots
+    (b a) 
+    (cl->* (->opt -Pathlike
+                  (Un (-lst -String) (-vec -String))
+                  (-lst (Un (-pair mode-sym
+                                   (-lst (-lst* (-lst -String)
+                                                ;; Accepts flags procedures that take 0-5 mandatory
+                                                ;; command-line arguments and (implicitly)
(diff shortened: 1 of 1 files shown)


samth:
This looks like it has some errors in the test suite, as you can see here: https://travis-ci.org/racket/typed-racket/jobs/365295658#L747

LiberalArtist:
Sorry it's taken me so long to get back to this. I believe this commit fixes the test failures, at the cost of using `Any` as the type of the flag specification list.

samth:
That seems like the right compromise to me. And I see the tests now pass.

PR #699, opened 2018-04-23
prefab fix RFC

pnwamk:
Trying to describe how prefabs _should_ behave.

Currently they act like standard structs, which is unsound (since we cannot guarantee what values will inhabit their fields since they can be created anywhere (i.e. they're prefabs -- anyone can construct them)).

[Rendered](https://github.com/pnwamk/typed-racket/blob/prefab-rfc/rfcs/text/0000-prefab-structs.md)

### DIFF
--- a/rfcs/text/0001-prefab-structs.md
+++ b/rfcs/text/0001-prefab-structs.md
@@ -1,0 +1,353 @@
+- Feature Name: prefab-structs
+- Start Date: 2018-04-24
+- RFC PR: (leave this empty)
+- Feature Commit(s): (leave this empty)
+
+# Summary
+
+Typed Racket's handling of #:prefab structs is currently (as of April 2018) unsound.
+Instead of treating them like standard structs where we can assume (and enforce)
+that fields always contain values of a certain type (i.e. the current behabior), 
+prefabs should more-or-less be treated exactly like tagged n-ary cons/mcons cells.
+
+# Motivation
+
+Prefabs are clearly useful and desirable (e.g. they're serializable) and our current
+lack of sound support for them should be addressed.
+
+# Guide-level explanation
+
+An __immutable prefab__ is in essence a n-ary tuple whose fields can contain
+any value. They are defined with the following syntax:
+
+```racket
+(struct point ([x : Integer] [y : Integer])
+  #:prefab)
+```
+
+This defines a predicate and accessors which work for _any_ compatible
+prefab (i.e. any prefab with the same [prefab key](https://docs.racket-lang.org/reference/structutils.html#%28def._%28%28quote._~23~25kernel%29._prefab-struct-key%29%29)), and a constructor which expects two `Integer`s.
+
+Examples:
+
+```racket
+(point? (point 1 2)) ; => #t
+(ann (point 1 2) point)
+(ann (point 1 2) (Prefab point Integer Integer))
+(point "1" "2") ; => fails w/ type error "Expected Integer, Given String"
+(point? #s(point "1" "2")) ; => #t
+(ann #s(point "1" "2") (Prefab point String String))
+(point-x (point 1 2)) ; => 1
+(point-x #s(point "1" "2")) ; => "1"
+(ann (lambda (x) 
+       (if (point? x)
+           ;; We don't know what the types for the fields
+           ;; of `x` are; they could be anything.
+           (ann x (Prefab point Any Any))
+           42)))
+     (-> Any Any))
+```
+
+A __mutable prefab__ (i.e. a prefab where each field is mutable) is 
+defined by adding the keyword `#:mutable`:
+
+```racket
+(struct initials ([first : Symbol] [last : Symbol])
+  #:prefab
+  #:mutable)
+```
+
+and is the same as an immutable prefab except that when the
+field types for the prefab are unknown, the type system will allow
+those fields ty be read from but not written to (i.e. in the same 
+way after testing `box?` tells the type system only that there is 
+a box we can read from but not what kind of values can be written).
+
+# Reference-level explanation
+
+## Prefab Top Types
+
+`(PrefabTop key field-count)` is the top type for the prefab described by the
+key `key` which has `field-count` fields. For immutable prefab structs, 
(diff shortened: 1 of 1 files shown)


rfcs/text/0000-prefab-structs.md
```diff
@@ -0,0 +1,128 @@
...
+`point ≈ (Prefab point Integer Integer)`
+
+and the following functions:
+
+- `point` of type `(-> Integer Integer point)`
+
+- `point?` of type `(-> Any Boolean : (Prefab point Any Any))` 
+(_NOTE: the type has `Any` not `Integer` in the field! This is because
+        a prefab could have been created anywhere with any value in
+        its fields_)
```
bennn: And `(make-predicate point)` will have type `(-> Any Boolean : point)`?

rfcs/text/0000-prefab-structs.md
```diff
@@ -0,0 +1,128 @@
...
+
+- `initials-first` with type `(All (F L) (-> (Prefab initials F L) F))` _and_
+type `(-> (PrefabTop initials) Any)` (i.e. a `case->`/intersection)
+
+- `initials-last` with type `(All (F L) (-> (Prefab initials F L) L))` _and_
+type `(-> (PrefabTop initials) Any)` (i.e. a `case->`/intersection)
+
+- `set-initials-first!` with type `(All (F L) (-> (Prefab initials F L) F Void))`
+
+- `set-initials-last!` with type `(All (F L) (-> (Prefab initials F L) L Void))`
```
bennn: The `set!` functions should be `case->`s too, right? To work for the top type.

rfcs/text/0000-prefab-structs.md
pnwamk: I would think so -- the generated predicate would of course need to inspect the fields. I'll add a note to the RFC.

rfcs/text/0000-prefab-structs.md
pnwamk: No, I don't believe so. If we don't know what the mutable field types are, there's no safe way to write to them. Right?

samth: I'm not sure what you mean -- the type of `set-box!` doesn't mention `BoxTop`, and similarly for vectors.

samth: I don't see how these functions could have different types than `vector-set!` -- mutable prefabs are basically named vectors.

pnwamk:
This RFC is ready for general comments/feedback/etc from the broader racket-dev/racket-users community.

pnwamk:
I updated the `PrefabTop` type to include a field count (i.e. a literal natural number in the type) since without that the key alone doesn't seem specific enough... it can describe any struct with that key (i.e. the number of fields is arbitrary).

samth:
We should at least list some alternatives to doing this (and their drawbacks).

1. Do nothing (always an option).
2. Various different typings for selectors etc. 
3. Fix other things but somehow leave the bad predicate behavior in place.

And any others we've thought of.

pnwamk:
**Final comment period:** if there are no further changes or unresolved discussion by COB Friday (May 4th) we'll begin considering merging a PR of Typed Racket changes consistent with this proposal.

Click [here](https://github.com/pnwamk/typed-racket/blob/f8e9fa42b0b5b747f08e0a8f40a179143604bafe/rfcs/text/0000-prefab-structs.md) to view the current RFC.

samth:
@rfindler might be interested in this since it relates to backwards compatibility.

rfindler:
I don't understand why this program (in the backwards compatibility section):

```
(ann (lambda (p) (if (point? p) (point-x p) p))
     (-> (U point Integer) Integer))
```

will still typecheck. How do you know the argument came from that prefab declaration and not another one with a different type annotation (or one from untyped code)?

As for backwards compatibility, I don't see any effort in the RFC to try to quantify the effect that this change will have or to consider other alternatives that might have less of an effect.

samth:
For the program you mention, the type `point` refers to the declaration, and so the type system ensures that if `p` is a `point?`, it must be the right kind of `point?`. This of course implies that the contract for `point` can't just be `point?`.  The RFC explains this below the example, but probably it can be made more clear.

As to enumerating options with less effect, that's what the "Drawbacks & Alternatives" section is for. If there are other alternatives we haven't considered, I'd love to hear about them. 

We hope to quantify possible issues once we have a working implementation (it's hard to quantify the inference issues without one, for example).

pnwamk:
I'll try and make the "Drawbacks & Alternatives more clear" by adding some examples which illustrate the alternatives.

rfindler:
Sounds like you guys are well ahead of me. Great!

pnwamk:
Drawbacks and alternatives section has been moved (it seems to make more sense after the backwards compatibility discussion) and updated with examples.

Link [here](https://github.com/pnwamk/typed-racket/blob/ce93d46dc0f83960f4f8ffc438d9b3ef8a849719/rfcs/text/0000-prefab-structs.md).

rfcs/text/0000-prefab-structs.md
```diff
@@ -0,0 +1,336 @@
...
+```racket
+(map point-x
+     (list (point 1 2) (point 3 4)))
+```
+
+but will break if we only use the polymorphic type for accessors 
+(since our limited inference cannot infer types for polymorphic 
+arguments to polymorphic functions)
+
+# Drawbacks and Alternatives
```
samth: This section should discuss whether Typed Racket could change the behavior of `struct` to produce functions that work with the types we've been giving them. 

I think the answer is: yes for immutable prefabs whose fields are all first-order data, no in other cases.

samth:
@rfindler Can you take a look at the revised alternative section and see what you think?

rfindler:
I don't feel like I have a full enough sense of what existing TR code looks like or how the TR type system works in practice to feel confident in any judgment between those alternatives.

pnwamk:
I think we're at a point we're we've listed how things should work, the risks of the changes, and a variety of alternatives, and we can now try implementing the suggested solution and decide if it should be tweaked further.

Merging this soon if there's no objection.

PR #700, opened 2018-04-25
Improve contract-stronger for object/c-opaque

dfeltey:
The previous implementation would never return #t since it would always be comparing two opaque structs. This version compares the names stored in the `restric-typed->/c` and `restrict-typed-field/c` structures and significantly improves performance in this program:

main.rkt

```
#lang typed/racket

(require/typed "automata.rkt"
  (make-random-automaton
    (-> (Instance Automaton))))

(define-type Automaton
  (Class
   [match-pair
    (-> (Instance Automaton) (Instance Automaton))]))

(for/fold ([a (make-random-automaton)])
          ([_ (in-range 7)])
  (send a match-pair a))
```

automata.rkt

```
#lang racket

(provide
 make-random-automaton)

(define (make-random-automaton)
  (new automaton%))

(define automaton%
  (class object%
    (super-new)

    (define/public (match-pair other) other)
      
    (define/public (guts) #f)
    ))
```

Before this commit running main.rkt takes  ~20 seconds, with this commit ~500ms

### DIFF
--- a/typed-racket-lib/typed-racket/utils/opaque-object.rkt
+++ b/typed-racket-lib/typed-racket/utils/opaque-object.rkt
@@ -39,6 +39,10 @@
 
 (provide object/c-opaque)
 
+(module+ for-testing
+  (provide restrict-typed->/c
+           restrict-typed-field/c))
+
 ;; projection for base-object/c-opaque
 (define ((object/c-opaque-late-neg-proj ctc) blame)
   (λ (obj neg-party)
@@ -172,7 +176,11 @@
         #:property prop:chaperone-contract
         (build-chaperone-contract-property
          #:name restrict-typed->-name
-         #:stronger equal?
+         #:stronger
+         (λ (this that)
+           (define this-name (restrict-typed->/c-name this))
+           (define that-name (restrict-typed->/c-name that))
+           (eq? this-name that-name))
          #:late-neg-projection restrict-typed->-late-neg-projection))
 
 (define (restrict-typed-field-late-neg-proj ctc)
@@ -198,5 +206,9 @@
         #:property prop:flat-contract
         (build-flat-contract-property
          #:name restrict-typed-field-name
-         #:stronger equal?
+         #:stronger
+         (λ (this that)
+           (define this-name (restrict-typed-field/c-name this))
+           (define that-name (restrict-typed-field/c-name that))
+           (eq? this-name that-name))
          #:late-neg-projection restrict-typed-field-late-neg-proj))
--- a/typed-racket-test/succeed/opaque-object-stronger.rkt
+++ b/typed-racket-test/succeed/opaque-object-stronger.rkt
@@ -1,5 +1,6 @@
 #lang racket
-(require typed-racket/utils/opaque-object)
+(require typed-racket/utils/opaque-object
+         (submod typed-racket/utils/opaque-object for-testing))
 
 ;; --------------------------------------------------------------------------------------------------
 ;; test helpers
@@ -183,4 +184,22 @@
       (field (x real?))
       (m1 (->m any/c any/c any/c))))
 
+)
+
+(let () ;; restrict-typed->/c and restrict-typed-field/c
+    (test-stronger?
+     (restrict-typed->/c 'foo)
+     (restrict-typed->/c 'foo))
+
+    (test-not-stronger?
+     (restrict-typed->/c 'foo)
+     (restrict-typed->/c 'bar))
+
+    (test-stronger?
+     (restrict-typed-field/c 'foo)
+     (restrict-typed-field/c 'foo))
+
+  (test-not-stronger?
+     (restrict-typed-field/c 'foo)
+     (restrict-typed-field/c 'bar))
 )
--- a/typed-racket-test/succeed/opaque-object-stronger.rkt
+++ b/typed-racket-test/succeed/opaque-object-stronger.rkt
@@ -203,3 +203,22 @@
      (restrict-typed-field/c 'foo)
      (restrict-typed-field/c 'bar))
 )
+
+(let ()
+  (define ctc
+    (object/c-opaque
+     (mtd (->m any/c any/c))))
+  (define (make-obj)
+    (new
+     (class object%
+       (super-new)
+       (define/public (mtd x) x)
+       (define/public (guts) #f))))
+  (define c1
+    (value-contract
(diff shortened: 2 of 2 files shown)


typed-racket-test/succeed/opaque-object-stronger.rkt
```diff
@@ -184,3 +185,21 @@
       (m1 (->m any/c any/c any/c))))
 
 )
+
+(let () ;; restrict-typed->/c and restrict-typed-field/c
```
samth: Can we add some tests with actual object contracts?

typed-racket-lib/typed-racket/utils/opaque-object.rkt
```diff
@@ -198,5 +206,9 @@
         #:property prop:flat-contract
         (build-flat-contract-property
          #:name restrict-typed-field-name
-         #:stronger equal?
+         #:stronger
+         (λ (this that)
+           (define this-name (restrict-typed-field/c-name this))
```
samth: What sort of data do these accessors produce?

typed-racket-test/succeed/opaque-object-stronger.rkt
dfeltey: I don't know how to write a test like that. You only get one of the restrictions when an opaque object contract is dynamically applied to an object with methods or fields not mentioned in the contract.

typed-racket-lib/typed-racket/utils/opaque-object.rkt
dfeltey: I think they should always be symbols, the code building them is at https://github.com/dfeltey/typed-racket/blob/2eeed5eb40ce76e0b97a74153d26efee57cd24af/typed-racket-lib/typed-racket/utils/opaque-object.rkt#L67

typed-racket-test/succeed/opaque-object-stronger.rkt
bennn: How about running the example program with a big time & memory limit?

typed-racket-lib/typed-racket/utils/opaque-object.rkt
samth: Can you give an example of one of these names? Is it just the field name (I assume it's something more complicated)?

typed-racket-test/succeed/opaque-object-stronger.rkt
samth: Can you then write a contract showing that something gets wrapped only once (perhaps using `eq?`)?

typed-racket-lib/typed-racket/utils/opaque-object.rkt
dfeltey: I think it's just the field or method name, for example `'guts ` in the example.

samth: But then won't this fire for two contracts that just agree on the method name? Maybe I'm missing something important about how this works.

typed-racket-test/succeed/opaque-object-stronger.rkt
dfeltey: I'm not sure what a test case like that would look like, doing something like this:
```
(let ()
  (define ctc
    (object/c-opaque
     (mtd (->m any/c any/c))))
  (define o
    (new
     (class object%
       (super-new)
       (define/public (mtd x) x)
       (define/public (guts) #f))))
  (define oc (contract ctc o 'p 'n))
  (define oc2 (contract ctc oc 'p 'n))
  (define oc3 (contract ctc oc2 'p 'n))
  (define oc4 (contract ctc oc3 'p 'n))
  (eq? oc3 oc4))
```
doesn't work.

@rfindler do you have any ideas on writing a better test case for this?

typed-racket-lib/typed-racket/utils/opaque-object.rkt
dfeltey: It would fire for 2 contracts that enforce there is a method of that name, but that it can never be called. There's no associated contract because there's no type given for the method and TR has to guard it to make sure it can't be called.

typed-racket-test/succeed/opaque-object-stronger.rkt
rfindler: nope!

typed-racket-lib/typed-racket/utils/opaque-object.rkt
samth: Ah, now I understand. Just to make sure:

This changes `contract-stronger?` to recognize that if two contracts both restrict the same particular field to not be callable, then they are equally-strong.

If that's right, can you put it in the commit message?

typed-racket-test/succeed/opaque-object-stronger.rkt
samth: What about using `contract-stronger?` on the results of `value-contract`?

Also, why does that test fail? Also, why do you need 4 uses of `contract` there -- shouldn't two be enough?

dfeltey: I pushed a better test case using `contract-stronger?` and `value-contract`. I don't fully understand object contracts well enough to understand how the contract elimination strategy affects things overall.

typed-racket-lib/typed-racket/utils/opaque-object.rkt
dfeltey: Yes that's right, if things look good with the new test case I'll fixup the commit message when I squash and push.

typed-racket-test/succeed/opaque-object-stronger.rkt
samth: Awesome. @rfindler, do you know why @dfeltey's initial test didn't work?

typed-racket-lib/typed-racket/utils/opaque-object.rkt
samth: Great. This all looks good -- let's make sure Travis CI is happy and then merge.

typed-racket-test/succeed/opaque-object-stronger.rkt
rfindler: I think a reasonable test case calls `contract-stronger?` on the actual contract that gets generated (since that's what this fix is about). As for the test case above, I don't think it should be `eq?`?

PR #701, opened 2018-04-26
lift some work out ahead of the blame application in the opaque-object

rfindler:
contract combinator

Combined with recent other improvements in racket/contract
(https://github.com/racket/racket/commit/bff0c4113d41ab35d8a7b5f7a4f19c3b453f79f7 and https://github.com/racket/racket/commit/2a1c8a78a59515d9f68bd1f98c1a26dec91ab6bf)
this change is about a 1.4x speedup on configuration 01000
of dungeon and about a 1.3x speedup on configuration 100000000
of acquire

### DIFF
--- a/typed-racket-lib/typed-racket/utils/opaque-object.rkt
+++ b/typed-racket-lib/typed-racket/utils/opaque-object.rkt
@@ -155,23 +155,17 @@
 ;; method is typed (assuming that the caller is untyped or the receiving
 ;; object went through untyped code)
 (define (((restrict-typed->-late-neg-projection ctc) blame) val neg-party)
-  (define blame+neg-party (cons blame neg-party))
-  (chaperone-procedure val
-                       (make-keyword-procedure
-                        (λ (_ kw-args . rst)
-                          (with-contract-continuation-mark
-                           blame+neg-party
-                           (when (typed-method? val)
+  (cond
+    [(typed-method? val)
+     (chaperone-procedure val
+                          (make-keyword-procedure
+                           (λ (_ kw-args . rst)
                              (raise-blame-error (blame-swap blame) val #:missing-party neg-party
                                                 "cannot call uncontracted typed method"))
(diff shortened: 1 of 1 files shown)


rfindler:
It makes sense for someone to actually fix `opaque-object/c` along the lines described in the comment that Asumu wrote 5 years ago. This commit is a step in that direction, but I didn't do the rest because it isn't necessary for this particular speed up. It probably does speed up things in a measurable way, however.

typed-racket-lib/typed-racket/utils/opaque-object.rkt
```diff
@@ -60,19 +62,19 @@
...
-      (dynamic-object/c (append methods remaining-methods)
-                        (append method-ctcs
-                                (for/list ([m remaining-methods])
-                                  (restrict-typed->/c m)))
-                        (append fields remaining-fields)
-                        (append field-ctcs
-                                (for/list ([m remaining-fields])
-                                  (restrict-typed-field/c m)))))
+    (define restrict-guard/c
+      (dynamic-object/c remaining-methods
```
bennn: If `remaining-methods` and `remaining-fields` are empty lists, then could we skip making & applying this `dynamic-object/c`?

rfindler: good idea, thanks.

Do you know how to tell if there is test coverage?

rfindler:
actually, looking more carefully at what's going on with the rest of the `opaque-object/c` code, it seems suspicious. Look here:

https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/utils/opaque-object.rkt#L151-L168

Why ever create the identity chaperone? Wouldn't it be better to lift the `typed-method?` check outside of the call to chaperone-procedure?

rfindler:
and why not inspect the arity of the method procedure and do something more efficient when it is a simple arity?

Issue #703, opened 2018-04-28
Quasisyntax: No export for check-splicing-list


MichaelBurge:
### What version of Racket are you using?

v6.12

### What program did you run?
```scheme
(: xs (Listof Syntax))
(define xs (list #'1 #'2))

(: id (-> Syntax Syntax))
(define (id x) x)

(: x (Syntaxof Any))
(define x #`(#,@(map id xs) #'foo))

x
```

### What should have happened?

The program should compile and run without error. Instead, the type checker does not have a type for `check-splicing-list`, which is emitted by the `quasisyntax` macro.

### If you got an error message, please include it here.

```scheme
a.rkt:10:10: Type Checker: missing type for identifier;
 consider using `require/typed' to import it
  identifier: check-splicing-list
  from module: qqstx.rkt
  in: (quasisyntax ((unsyntax-splicing (map id xs)) (syntax foo)))
```

PR #704, opened 2018-04-28
Types for quasisyntax list splicing

MichaelBurge:
Corrects issue #703.

* Adds a type for check-splicing-list.
* Corrects an erroneous type for h-splice. It's called in exactly one place, where its 2nd argument looks like '(~@! . uqs5).

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-special-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-special-env.rkt
@@ -290,12 +290,15 @@
   ;; check-in-port
   [(make-template-identifier 'check-in-port 'racket/private/for)
    (-> Univ Univ Univ)]
-  ;; from the expansion of `with-syntax'
+  ;; from the expansion of `with-syntax' or `quasisyntax'
   [(make-template-identifier 'apply-pattern-substitute 'racket/private/stxcase)
    (->* (list (-Syntax Univ) Univ) Univ (-Syntax Univ))]
   ;; same
   [(make-template-identifier 'with-syntax-fail 'racket/private/with-stx)
    (-> (-Syntax Univ) (Un))]
+  ; from `quasisyntax'
+  [(make-template-identifier 'check-splicing-list 'racket/private/qqstx)
+   (-> Univ (-Syntax Univ) (-Syntax Univ))]
   ;; more from with-syntax, a Guide is ...
   [(make-template-identifier 't-append 'racket/private/template)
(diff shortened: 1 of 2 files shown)


typed-racket-lib/typed-racket/base-env/base-special-env.rkt
```diff
@@ -313,7 +316,7 @@
    (-> Univ (-lst (-Syntax Univ)) (-Syntax Univ))]
   ;; ... or a Guide is ...
   [(make-template-identifier 'h-splice 'racket/private/template)
-   (-> Univ (Un (-val #f) (-Syntax Univ)) (Un (-val #f) (-Syntax Univ)) (-lst (-Syntax Univ)))]
+   (-> Univ (-pair -Symbol -Symbol) (Un (-val #f) (-Syntax Univ)) (-lst (-Syntax Univ)))]
```
bennn: @rmculpepper  can you double-check this type?

In #690 you'd suggested the type `h-splice : Any (U Syntax #f) (U Syntax #f) -> (Listof Syntax)`

Maybe the 2nd argument should be a `-Stxish` [(link)](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/types/abbrev.rkt#L100)?

rmculpepper: The second and third arguments can probably just be `Any`; they're just there for error reporting.

MichaelBurge: @bennn I've pushed a new commit that widens these types.

typed-racket-lib/typed-racket/base-env/base-special-env.rkt
```diff
@@ -290,12 +290,15 @@
...
-  ;; from the expansion of `with-syntax'
+  ;; from the expansion of `with-syntax' or `quasisyntax'
   [(make-template-identifier 'apply-pattern-substitute 'racket/private/stxcase)
    (->* (list (-Syntax Univ) Univ) Univ (-Syntax Univ))]
   ;; same
   [(make-template-identifier 'with-syntax-fail 'racket/private/with-stx)
    (-> (-Syntax Univ) (Un))]
+  ; from `quasisyntax'
+  [(make-template-identifier 'check-splicing-list 'racket/private/qqstx)
+   (-> Univ (-Syntax Univ) (-Syntax Univ))]
```
samth: This could be more precise, re-using the type for `datum->syntax`, but it's fine to leave it alone at the moment.

PR #705, opened 2018-05-01
Property for forgetting type information.

samth:
Contains RFC and initial implementation.

[Rendered](https://github.com/samth/typed-racket/blob/ignore-type-info/rfcs/0000-ignore-type-info.md)

### DIFF
--- a/rfcs/0000-ignore-type-info.md
+++ b/rfcs/0000-ignore-type-info.md
@@ -29,6 +29,8 @@ This was developed specifically to support this pattern in a revised
 version of the keyword argument protocol currently under development
 by @mflatt.
 
+Note that simply using `ann` is not possible here, both because this property is intended to be used from places (such as `racket/private/kw`) that do not depend on Typed Racket, and because `ann` does not have the needed behavior with respect to propositions.
+
 # Guide-level explanation
 
 Any expression with the shape `(#%expression sub)` that has a true
--- a/rfcs/0000-ignore-type-info.md
+++ b/rfcs/0000-ignore-type-info.md
@@ -52,14 +52,16 @@ has type `Any` along with `true-prop` for both propositions and
 # Drawbacks and Alternatives
 [drawbacks]: #drawbacks
 
+#### Drawbacks
+
 Adds complexity to the type checker.
 
 #### Alternative: use expected type
 
 If there's an expected type when checking `(#%expression sub)` with
-the relevant property, we could instead use it as both the expected
-type when checking `sub` as well as the result of type checking the
-entire expression.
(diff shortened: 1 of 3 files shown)


rfcs/0000-ignore-type-info.md
```diff
@@ -0,0 +1,82 @@
...
+`empty-obj` for the object.
+
+
+
+# Drawbacks and Alternatives
+[drawbacks]: #drawbacks
+
+Adds complexity to the type checker.
+
+#### Alternative: use expected type
```
pnwamk: I think it should be just a little more clear, e.g. "use the expected type (if one exists) to type check the marked expression instead of using `Any` for the expected type -- and in either case the overall result will still be `Any`".

pnwamk:
made one minor comment to clarify the suggested alternative -- other than making that slightly clearer, LGTM.

rfcs/0000-ignore-type-info.md
samth: No, that's not quite the intent of the alternative -- it's to use the expected type in both cases.

MichaelBurge:
Is the syntax property equivalent to a type annotation with `Any`?
```scheme
(ann (ann "hello" Any) String)
```
A syntax property might be preferable for storing type information because macros often ignore them when parsing.

If that's the reason, would it make sense for the syntax property to store any type rather than just `Any`?

samth:
There are two reasons why that's not quite the same thing (this should probably be clarified in the RFC). 

1. This property can be added by code that doesn't depend on Typed Racket and therefore cannot generate a properly scoped reference to `Any`.
2. Annotating with `(ann e Any)` does not drop propositions (intentionally), which would make the specific use case where we need this not work.

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
```diff
@@ -151,6 +151,9 @@
       [stx:exn-handlers^
        (register-ignored! form)
        (check-subforms/with-handlers form expected) ]
+      [(~and stx:typed-racket:ignore-type-information^ (#%expression inner))
+       (tc-expr/check/internal #'inner Univ)
+       (ret Univ -true-propset -empty-obj)]
```
samth: This line is wrong, it should be `(ret Univ (-PS -tt -tt) -empty-obj)`.

cc @mflatt

mflatt:
In case someone wants to see an example use, see the top of "racket/collects/racket/private/kw.rkt" in mflatt/racket@99d069df780df5fe9ed69b5e2136fff3a6a84217

samth:
I hope to merge this soon, in view of #706.

PR #707, opened 2018-05-02
order-of-evaluation props

bennn:
Use props from checking a function to check its first argument,
use props from checking the first argument to check the second,
and so on.

Makes things like this type check: `(lambda: ((x : (Un Symbol Natural))) (+ (assert x integer?) x))`


typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
```diff
@@ -352,6 +352,25 @@
           #:stx form
           "expected single value, got multiple (or zero) values")]))
 
+;; Apply `single-value` to a list of forms in order, accumulate the prop info
+(define (map-single-value forms [props '()])
+  (define any-res (-tc-any-results #f))
```
pnwamk: Would `check-in-order/single-value` (or something similar) be a clearer name? `map-single-value` seems too generic to me.

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
```diff
@@ -352,6 +352,25 @@
...
+  (let loop ((forms forms)
+             (props props))
+    (match forms
+     ['()
+      '()]
+     [(cons e rst)
+      (define tcr
+        (with-lexical-env+props
+          props
```
pnwamk: I wonder if `with-lexical-env+props` is right here and not `with-lexical-env+props-simple`? The latter seems to do a little less work... TBH I'm not sure which to favor here.

typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -4467,6 +4467,9 @@
         [tc-e/t (let: ([x : (Un Flonum Natural) 0.0])
                   (if (not (natural? x)) x 1.0))
                -Flonum]
+        ;; props + evaluation order
+        [tc-e ((lambda: ((x : (Un Symbol Natural))) (+ (assert x integer?) x)) 1) -Nat]
+        [tc-e ((lambda: ((x : (Un Symbol Natural))) ((begin (assert x integer?) +) x x)) 1) -Nat]
        )
```
pnwamk: I think we need more tests. What about

1. Tests where the combinations of the first several arguments then allows the final argument(s) to type check?
2. Tests were at some point the environment becomes absurd (and thus the final arguments are dead code and don't need to be type checked)?

Maybe there's other's that would be useful.

pnwamk:
Questions:

1. Are there any downsides to doing this? The only possible conflict I can think of is that dependent function types allow arguments to depend on each other in _any_ order (not in a fixed left to right order) and you may want to check those in that particular order... I'm not sure if/how these features would interact. Maybe we should add a check that the function is not a dependent function to be safe?

2. Does this affect performance at all? It seems like it probably wouldn't for most programs (most arguments don't have interesting propositions, right?) but it would be nice to double check w/ some numbers just so we know.

pnwamk:
I like this PR -- I'm a little nervous because we could clearly never "undo" it so we need to make sure we feel like we've considered all the possible repercussions.

samth:
I think this is something that would benefit from writing down the RFC as part of this process.

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
bennn: I used `with-lexical-env+props` because that's what `tc-body/check` uses for function bodies.

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
bennn: Renamed to `map-single-value/unconditional-prop` because:

- `map single-value` is what we were doing before
- `map` implies left-to-right order
- `/unconditional-prop` explains the 2nd argument
- and the prop that gets accumulated is the "unconditional prop" from the arguments

I'm happy with this name, but not opposed to changing it

typed-racket-test/unit-tests/typecheck-tests.rkt
bennn: update: testing has led to some internal typechecker errors wrt the optimizer, will fix

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
pnwamk: I've thought more about this and looked at the code more; here are a few comments:

1. If this is a good change and the right thing to do (and I think it is) we should be doing it any time we can, every time we check function arguments. In particular we should also be accumulating the propositions [here](https://github.com/racket/typed-racket/blob/18383003b43adf50aa5af6cb5c1fe92d286eae7e/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-main.rkt#L98), [here](https://github.com/racket/typed-racket/blob/18383003b43adf50aa5af6cb5c1fe92d286eae7e/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-main.rkt#L97)., [here](https://github.com/racket/typed-racket/blob/18383003b43adf50aa5af6cb5c1fe92d286eae7e/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-main.rkt#L91), etc, i.e. we should do it for both standard and dependent functions _whenever_ there could be propositions that help the remaining arguments type check.

2. Does it make sense to also do this on arguments were we have calculated an expected type (i.e. [here](https://github.com/racket/typed-racket/blob/18383003b43adf50aa5af6cb5c1fe92d286eae7e/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-main.rkt#L111))? I can't remember... is there a way we can have an expected type but still extract meaningful propositions from an expression? If so we should remember those props as well.

3. With these changes, we're now extracting the unconditional propositions from argument type results twice (see [here](https://github.com/racket/typed-racket/blob/18383003b43adf50aa5af6cb5c1fe92d286eae7e/typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt#L64) for the other time). I think we should probably just do this once (i.e. _while_ we're checking the argument types in order from left-to-right as you have done here) and then we should pass that prop along (maybe as an additional argument to [tc/funapp](https://github.com/racket/typed-racket/blob/18383003b43adf50aa5af6cb5c1fe92d286eae7e/typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt#L62)) so it can be propagated/used where needed. In other words, the logic from the code I linked above that is also extracting unconditional argument propositions should appear in whatever we call `map-single-value/unconditional-prop`.

4. I think `map` as a part of the name is a misnomer because it is accumulating something other than a list along the way (the unconditional props) that it should be returning in addition to the list of type results (i.e. we should pass along that accumulated logical info since it is used/needed elsewhere).

5. One other misc note: this behavior reminds me of `tc-body/check` which also accumulates unconditional info as It checks expressions guaranteed to be evaluated in order. This observation may not be useful... but I figured I'd mention it.

pnwamk:
I think it _might_ be useful to have the RFC be it's own pull request. That way we can agree on the idea itself separate from the work to implement, merge the RFC when it's ready, and then we can focus on working together to get this PR done and in sync with the RFC.

typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt
bennn: `2.` yes
`4.` "map" alone would be bad to me, but I like the current "map with prop". (I'm happy to take suggestions, but I'm personally not going to think about new names)

bennn:
This program came up last year (on Discord?) and has exactly the same problem: `((get-foo a1) a1)` is checking and using `a1` is the same expression.

```
#lang typed/racket

(provide num/foo foo-get num Num)



(: prop:foo (Struct-Property (-> Self Number)))

(: foo-pred (-> Any Boolean : (Has-Struct-Property prop:foo)))

(: foo-get (-> (Has-Struct-Property prop:foo)

               (Some (X)

                     (-> X Number)

                     : #:+ X)))

(define-values (prop:foo foo-pred foo-get)

  (make-struct-type-property 'foo))



(struct num ([n : Number])

  #:type-name Num

  #:property prop:foo

  (lambda ([me : Num])

          : Number

    (num-n me)))



(: num/foo Number)

(define num/foo

  (let ([a1 : Num

            (num 42)])

    ((foo-get a1) a1)))



(module+ test

  (require typed/rackunit)

  ;; succeeds

  (check-equal? num/foo 42))



;; However adding the following inline test for the body of num/foo as follows:

;; 
;; 
;; (module+ test
;; 
;;   (check-equal?
;; 
;;    (ann
;; 
;;     (let ([a1 : Num
;; 
;;               (num 42)])
;; 
;;       ((foo-get a1) a1))
;; 
;;     Number)
;; 
;;    42))
;; 
;; 
;; 
;; Results in the following type error
;; 
;; 
;; 
;; Type Checker: type mismatch
;; 
;;   expected: (B 0)
;; 
;;   given: Num
;; 
;;   in: a1
```

PR #708, opened 2018-05-04
doc: add 'TR-optimize' log topic to the documentation

bennn:


stamourv:
LGTM.
Out of curiosity, why are you interested in the raw logs?

bennn:
Mostly because I want to keep running programs on the command-line, rather than fire up Dr. Racket.

the longer story: I have a fork of Typed Racket that guarantees a different kind of soundness. Instead of being type-sound, it guarantees that values match their type constructor (just like Reticulated Python).  I want this fork to work with the optimizer. Right now it doesn't and its helpful to see the logs to figure out why.

(I think my current problem is that I'm adding an `if` statement and not calling `test-position-add-true` etc.)

PR #709, opened 2018-05-05
Evaluation order rfc

bennn:
RFC for #707 

Goals:

- [ ] identify all kinds of function applications in Typed Racket (applying a normal function, applying a dependent function, the special cases in `typecheck/tc-app/*`)
- [ ] have a complete RFC


MichaelBurge:
Should evaluation order affect typechecking?

In the expression
```scheme
(+ x (begin (assert x integer?) x))
```
the right-hand `x` is known to be an integer. But the left-hand `x` is the same `x`: Since all arguments are evaluated before the function is invoked, it shouldn't be unsound if earlier arguments are accepted on the basis of later arguments passing their runtime checks.

Do we want the union of all propositions to apply to each argument?

One case where flow-sensitive analysis might be important, is if a `set!` form is used to change `x` partway through evaluating the arguments. But `set!` does not allow the type of `x` to be changed, so perhaps this doesn't matter.

bennn:
> Do we want the union of all propositions to apply to each argument?

No, because that could be unsound. In this application, we don't want to check the 1st argument to `+` assuming that `x` is an integer:

```
(+ (add1 x) (begin (assert x integer?) x))
```

I think we need to conservatively reject your example.

pnwamk:
Would it be helpful to describe this as checking function application in terms of its equivalent expansion into [A-normal form](https://en.wikipedia.org/wiki/A-normal_form) to help with intuition as for why it works and is correct?

i.e. we're basically leveraging the fact that checking this program:

```racket
(fun arg1 arg2 ...)
```

is equivalent to checking this one:

```racket
(let* ([x1 arg1]
       [x2 arg2]
       ...)
  (fun x1 x2 ...))
```

(where the `x`'s are _fresh_) because of how function application is defined in Racket. (I know that the `let*` could have been a `let` but perhaps the `*` helps emphasize the _necessary_ ordering of argument evaluation for people who haven't realized `let` also guarantees in-order evaluation in Racket)

Issue #710, opened 2018-05-07
regression: optional arguments


bennn:
### What version of Racket are you using?

<https://github.com/racket/typed-racket/commit/4b1a1bfa9d59aa2ae469e1217dd16aea9c2f5aa6>

### What program did you run?

```
#lang typed/racket/base

(: f (-> String String))
(define (f a [b ""])
  (string-append a b))
```

### What should have happened?

The program should type-check, possibly via:
- assume `a : String`
- infer `b : String`
- infer `f : (->* (String) (String) String)`
- check `(->* (String) (String) String) <: (-> String String)`
- success

### If you got an error message, please include it here.

```
Type Checker: type mismatch
  expected: String
  given: Any
  in: a
.... (clipped) ....
```

AlexKnauth:
The type of `a` should be less connected to the "outside" type. It comes from either the "outside" or from the default-expr. Here are two programs that show what I mean.

###### Program 1:
```racket
(: f (-> Integer))
(define (f [x 0])
  x)
```
Here the type of `x` in the body doesn't come from the outside, it comes from typechecking the `default-expr`, which is `0` with type `Zero`. That's why the type of `x` in the body is `Zero`.

###### Program 2:
```racket
#lang typed/racket
(: f (->* [] [Integer] (U Integer False)))
(define (f [x #false])
  x)
```
Here the type of `x` in the body doesn't come directly from the outside either. It comes from either the outside or the `default-expr`, so it is the union of those two types, `Integer` and `False`.

###### The new expansion and how it relates to TR
According to [racket/private/kw.rkt](https://github.com/racket/racket/commit/662a9022c0858f1de69712c31ce520bc80930fa3#diff-36cf40d20fdf98bdddffc0e7dc0c01edR44), Program 1 will expand to something like this: 
(if it treats 0 as if it was an arbitrary expression instead of a literal)
```racket
(: f (-> Integer))
(define f
  (let ([core
         (lambda (_x)
           (let* ([x (if (eq? _x unsafe-undefined)
                         0
                         _x)])
             ; body
             x))])
    (case-lambda
      [() (core unsafe-undefined)]
      [(x) (core x)])))
```
Here we see where the `Any` might come from: the `_x`. Typed Racket needs to recognize that `_x` always comes from either the "outside" if there is one, or `unsafe-undefined`. 

[...]

PR #711, opened 2018-05-08
repair problems with optional-argument handling

mflatt:
Intended as a repair for #710.

The last definition in the updated "opt-arg-test.rkt" does not typecheck in v6.12, but I'm pretty sure it should (as it now does).

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/annotate-classes.rkt
+++ b/typed-racket-lib/typed-racket/base-env/annotate-classes.rkt
@@ -11,8 +11,8 @@
 ;; ----------------
 ;;
 ;; A LambdaKeywords is a
-;;   (lambda-kws (Listof Keyword) (Listof Keyword) (Listof Keyword) (listof Boolean))
-(struct lambda-kws (mand opt opt-supplied pos-opt-supplied?))
+;;   (lambda-kws (Listof Keyword) (Listof Keyword) (Listof Keyword) Integer (listof Boolean))
+(struct lambda-kws (mand opt opt-supplied pos-mand-count pos-opt-supplied?))
 
 ;; interp.
 ;;   - the first list contains the mandatory keywords
@@ -276,6 +276,12 @@
                      (values (cons (syntax-e kw) mand-kws)
                              opt-kws
                              opt-kws-supplied))))
+             (define pos-mand-count
+               (for/sum ([kw (in-list kws)]
+                         [default (in-list defaults)]
+                         #:unless default
+                         #:unless kw)
+                 1))
              (define pos-opt-supplied?s
                (for/list ([kw (in-list kws)]
                           [default (in-list defaults)]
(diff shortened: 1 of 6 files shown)


bennn:
the tests & contract tests pass on my machine

pnwamk:
I've moved the tests into the type checking unit tests file:

https://github.com/pnwamk/typed-racket/tree/kw2

I added a few more tests---2 failed that I think should work at a glance:

```racket

(: foo (-> Number Number Number))
(define (foo x
             [y1 0]
             [y2 (+ 1 0)]
             #:z1 [z1 0]
             #:z2 [z2 (+ 1 0)])
  (+ x y1 y2 z1 z2))

```

fails with error message:

```
Type Checker: Expected a function of type 
(-> Nothing Boolean Nothing Boolean Number Number Number), 
but got a function with the wrong arity in: 
(define (foo x (y1 0) (y2 (+ 1 0)) #:z1 (z1 0) #:z2 (z2 (+ 1 0))) (+ x y1 y2 z1 z2))
```

and this program:

```racket
(: bar (-> Number Number Number Number))
(define (bar x
             [y1 0]
             [y2 (+ 1 0)]
             #:z1 [z1 0]
             #:z2 [z2 (+ 1 0)])
  (+ x y1 y2 z1 z2))

```

fails with this error:

```
Type Checker: Expected a function of type 
(-> Nothing Boolean Nothing Boolean Number Number Number Number), 
but got a function with the wrong arity in: 
(define (bar x (y1 0) (y2 (+ 1 0)) #:z1 (z1 0) #:z2 (z2 (+ 1 0))) (+ x y1 y2 z1 z2))
```

However, these also don't work on 6.12... so it's possibly (probably?) just an issue we've had all along and not a new issue (unless I'm just not noticing some obvious reason these function's shouldn't type check). If that's the case I'd be fine merging this fix and just submitting an issue for these programs.

pnwamk:
The error messages I posted were 6.12 messages. After this PR they instead are as follows:

```
Type Checker: Expected a function of type 
(-> Nothing #<unsafe-undefined> Number Number Nothing #<unsafe-undefined> Number), 
but got a function with the wrong arity in: 
(define (foo x (y1 0) (y2 (+ 1 0)) #:z1 (z1 0) #:z2 (z2 (+ 1 0))) (+ x y1 y2 z1 z2))
```

and

```
Type Checker: Expected a function of type 
(-> Nothing #<unsafe-undefined> Number Number Number Nothing #<unsafe-undefined> Number), 
but got a function with the wrong arity in: (define (bar x (y1 0) (y2 (+ 1 0)) #:z1 (z1 0) #:z2 (z2 (+ 1 0))) (+ x y1 y2 z1 z2))
```

pnwamk:
I commented out the two failing tests in my [fork](https://github.com/pnwamk/typed-racket/commit/230285e29dc70297495c889e7458f8e7a16670b2) --- like I said I'm happy to merge this (with the relocating of the tests I performed in my fork) and just submit an issue for the two failing tests that seem unrelated to these changes.

mflatt:
Thanks for the extra tests! I think I see the problem, so let me try a little more.

mflatt:
Updated to pass the new tests.

PR #714, opened 2018-05-08
static-contract: improve `or/sc` optimization

### LINKED ISSUE #712: Contract optimization generates `any/c` instead of `any` for `(Any -> Boolean)`
This program:

```
#lang typed/racket
(: f : Any -> Boolean)
(define (f x) (string? x))
(provide f)
```

generates the contract `(or/c struct-predicate-procedure? (-> any/c any/c))`. However, the last `any/c` should be `any`, which would turn it into a flat contract.

Thanks to @tonyg for the example that indicated this.

bennn:
Improve the static contract optimizer so a contract like:

```
(or/c struct-predicate-procedure? (-> any/c boolean?))
```

is optimized to a flat contract.

```
(or/c struct-predicate-procedure? (-> any/c any))
```

Closes #712

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/optimize.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/optimize.rkt
@@ -110,16 +110,25 @@
 
 
 ;; Reduce a static contract assuming that we trusted the current side
-(define ((make-trusted-side-reduce flat-sc?) sc)
+;; If `is-weak-side?` is true, then preserve the "head constructor" of the
+;;  result --- if `((make-trusted-side-reduce ...) sc #true) = sc+`, then
+;;  both `sc` and `sc+` must give the same answer to `contract-first-order`
+;;  after instantiation.
+(define ((make-trusted-side-reduce flat-sc?) sc is-weak-side?)
   (match sc
     [(->/sc: mand-args opt-args mand-kw-args opt-kw-args rest-arg (list (any/sc:) ...))
      (function/sc #t mand-args opt-args mand-kw-args opt-kw-args rest-arg #f)]
     [(arr/sc: args rest (list (any/sc:) ...))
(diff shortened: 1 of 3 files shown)


samth:
Can you explain more about both what the overall implementation approach is here (why didn't this work before, what the invariants are now) and why this is needed to fix #712?

bennn:
Before, the contract `(or/c struct-predicate-procedure? (-> any/c boolean?))` would get optimized to `(or/c struct-predicate-procedure? (-> any/c any/c))` and stop.

This PR lets `(or/c struct-predicate-procedure? (-> any/c boolean?))` optimize to `(or/c struct-predicate-procedure? (-> any/c any))`. After that, the existing code optimizes the whole contract to `any/c` because all components of the `or/sc` are now flat contracts.

The high-level idea is that its safe to change the contracts under an `or/sc` if the outermost contract-constructor doesn't change. (This may be overly conservative.)

For example, if we're exporting a value from Typed to untyped:
- It's safe to change `(or/sc integer? (-> boolean? boolean?))` to `(or/sc integer? (-> boolean? any/c))`
- it's **not safe** to change `(or/sc integer? (-> boolean? boolean?))` to `(or/sc any/c (-> boolean? any/c))`, because that will further optimize to `any/c` on leave the function un-protected

[Here](https://github.com/racket/typed-racket/blob/master/typed-racket-test/unit-tests/static-contract-optimizer-tests.rkt#L353) are some related tests.

Before this PR, all optimizations done by `make-trusted-side-reduce` were disabled. After this PR, some of those optimizations are enabled --- the ones that preserve contract constructors.

- - -

I tried to keep this explanation very short. Let me know if it's not clear, or you want more explanations, or you want an RFC for the static contract optimizer.

samth:
This is very helpful. Can you put that in the commit message and/or a comment in the file?

Issue #713, opened 2018-05-08
Missing optimization of flat contract


samth:
This program should generate an `any` result contract for `f`, but doesn't:

```
#lang typed/racket
(struct: splice ())
(define-type T (U Bytes splice #f))
(: f :  -> T)
(define (f) #f)
(provide f)
```

Removing any components of the type `T` makes the problem go away. 

Thanks to @tonyg for the original example.

bennn:
fixed via #716

PR #716, opened 2018-05-09
flatten contract for union types (fix flat contract optimization)

bennn:
This is a fix for #713 

A similar program with the same issue as #713 is this:

```
#lang racket

(module t typed/racket
  (define-type T (U Bytes #f (Listof #f)))
  (: f :  -> T)
  (define (f) #f)
  (provide f))

(require 't)
(has-contract? f) ;; should be #false
```

The problem is that the static contract for `(U Bytes #f (Listof #f))` is basically:
- `(or/sc (or/sc Bytes #f) (Listof #f))`

and the static contract optimizer doesn't know to flatten the 2 layers of `or/sc`.

It might be ok to change the optimizer to flatten layers of `or/sc`, but this PR is simpler and seems like a good thing to do anyway.

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/optimize.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/optimize.rkt
@@ -21,7 +21,7 @@
                                . ->* . static-contract?)])
 
 ;; Reduce a static contract to a smaller simpler one that protects in the same way
-(define (reduce sc)
+(define (reduce sc flat-sc?)
   (match sc
     ;; none/sc cases
     [(listof/sc: (none/sc:)) empty-list/sc]
@@ -61,7 +61,10 @@
       [(? (λ (l) (member none/sc l)))
        (apply or/sc (remove* (list none/sc) scs))]
       [(? (λ (l) (ormap (match-lambda [(or/sc: _ ...) #true] [_ #false]) l)))
-       (apply or/sc (flatten-or/sc scs))]
+       (define new-scs (flatten-or/sc scs flat-sc?))
+       (if new-scs
+         (apply or/sc new-scs)
+         sc)]
       [else sc])]
 
     ;; and/sc cases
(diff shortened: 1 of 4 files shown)


typed-racket-test/unit-tests/contract-tests.rkt
```diff
@@ -226,7 +226,9 @@
    (t-sc -Number number/sc)
    (t-sc -Integer integer/sc)
    (t-sc (-lst Univ) (listof/sc any-wrap/sc))
-   (t-sc (Un (-lst Univ) -Number) (or/sc number/sc (listof/sc any-wrap/sc)))
+   (t-sc (Un (-lst Univ) (-val #t)) (or/sc (flat/sc #''#t) (listof/sc any-wrap/sc)))
```
bennn: I changed this test because the change to `type-contract.rkt` flattens `number/sc` in the result to all its cases.

typed-racket-test/unit-tests/static-contract-optimizer-tests.rkt
```diff
@@ -166,6 +166,9 @@
     (check-optimize (or/sc set?/sc)
       #:pos any/sc
       #:neg set?/sc)
+    (check-optimize (or/sc (or/sc set?/sc list?/sc))
+      #:pos any/sc
+      #:neg (or/sc set?/sc list?/sc))
```
bennn: This test was just a sanity-check for me. It's not directly related to the change here.

samth:
I agree that this seems good (and should be merged), but I also think the optimizer should probably be recursive in some way that would avoid this.

typed-racket-test/unit-tests/static-contract-optimizer-tests.rkt
```diff
@@ -352,28 +355,28 @@
...
-      #:pos (or/sc cons?/sc (or/sc cons?/sc (box/sc cons?/sc)))
-      #:neg (or/sc cons?/sc (or/sc cons?/sc (box/sc cons?/sc))))
-    (check-optimize
-      ;; (or (or ...)), only the outer "or" contains a non-flat contract --- still don't optimize
-      (or/sc (box/sc cons?/sc) (or/sc cons?/sc set?/sc))
-      #:pos (or/sc (box/sc cons?/sc) (or/sc cons?/sc set?/sc))
-      #:neg (or/sc (box/sc cons?/sc) (or/sc cons?/sc set?/sc)))
+      #:pos (or/sc (box/sc cons?/sc) cons?/sc)
+      #:neg (or/sc (box/sc cons?/sc) cons?/sc))
+    (let ([rec-chaperone (recursive-sc (list #'x) (list cons?/sc) (box/sc (recursive-sc-use #'x)))])
```
bennn: Now using `recursive-sc` instead of `or/sc` because a recursive contract doesn't correspond to a value constructor (like `or/sc`) and doesn't get collapsed by other optimizations (unlike `or/sc`)

typed-racket-lib/typed-racket/static-contracts/optimize.rkt
```diff
@@ -60,6 +60,13 @@
       [(? (λ (l) (member any/sc l))) any/sc]
       [(? (λ (l) (member none/sc l)))
        (apply or/sc (remove* (list none/sc) scs))]
+      [(list pre-scs ... (or/sc: mid-scs ...) post-scs ...)
+       ;; Assumes `sc->kind` is the same for all `mid-scs`.
+       ;; If this assumption is false, then the flattened contract might
+       ;;  accept a value that the original contract failed for ---
+       ;;  consider `(or/c (or/c procedure? (-> boolean?)) (-> integer?))`
+       ;;  and `(or/c procedure? (-> boolean?) (-> integer?))` any thunk
```
samth: Seems like this comment ends prematurely.

typed-racket-lib/typed-racket/static-contracts/optimize.rkt
```diff
@@ -60,6 +60,14 @@
       [(? (λ (l) (member any/sc l))) any/sc]
       [(? (λ (l) (member none/sc l)))
        (apply or/sc (remove* (list none/sc) scs))]
+      [(list pre-scs ... (or/sc: mid-scs ...) post-scs ...)
```
samth: Will this flatten multiple `or/sc` that appear inside the outer or? Also, doing this repeatedly and via complex `...` patterns seems likely to be less efficient than just writing a recursive function directly.

bennn:
This PR currently:
- changes `type-contract.rkt` to produces a flat `or/sc` for a union type
- removes nested `or/sc` in the static contracts optimizer (with a recursive function)

Ok to merge? I'm a little nervous about the `or/sc` flattening because I'm not sure its safe for all contracts TR generates --- I just have a feeling its probably safe.

samth:
Can you say something about how it could be not safe to flatten `or/sc`? Is this because of `or/c` associativity, or because of subsequent optimization, or something else?

samth:
Could we just do flattening in cases where everything is a flat contract, then?

rfindler:
or/c adjusts the order based on flat vs non flatness.

It may be easiest to use a combinator that doesn't do that. Then flattening
is going to be fine.

bennn:
@samth yes I can pass [this flat-ness predicate](https://github.com/bennn/typed-racket/blob/25047de3519c9b47ac9543f6e5660de15c5c2d7e/typed-racket-lib/typed-racket/static-contracts/optimize.rkt#L331) to the [function that flattens or/scs](https://github.com/racket/typed-racket/pull/716/files#diff-e0a72c7e1300dddbafd62b9ff2c9752aR124)

bennn:
Current status:
- changes `type-contract.rkt` to produces a flat `or/sc` for a union type
- removes nested `or/sc` in the static contracts optimizer _if_ the nested `or/sc` contains [all flat]  or [all non-flat] contracts

can I merge this for racket 7?

bennn:
e6463cdaea3e48c127572a2218864bb706859520

PR #717, opened 2018-05-09
tweak a few expected err msgs in unit tests

pnwamk:


samth:
with that expected message, we might as well leave it off.

PR #723, opened 2018-05-15
fix: improve 'no type' error message for format strings

bennn:
Fix the error message for when a string containing format characters
hits the 'type-of' function.

Before: the error was from applying format to a string containing the bad value.

Now: the error reports the bad string

Example program:

```
#lang racket/base

(require typed-racket/types/type-table)

(type-of #'"hello ~e world")
```

Old error:
```
format: format string requires 1 arguments, given 0; arguments were: "no type for hello ~e world 
```

New error:
```
Internal Typechecker Error: no type for "hello ~e world" 
```


samth:
LGTM

PR #724, opened 2018-05-15
Simplify contract generation.

samth:


typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -449,8 +449,7 @@
                  (not (regexp? v)))
             (flat/sc #`(quote #,v))
             (flat/sc #`(flat-named-contract '#,v (lambda (x) (equal? x '#,v))) v))]
-       [(Base-name/contract: sym ctc)
-        (flat/sc #`(flat-named-contract '#,sym (flat-contract-predicate #,ctc)) sym)]
+       [(Base-name/contract: sym ctc) (flat/sc ctc)]
```
bennn: Why not `(flat/sc ctc sym)` ?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -1013,7 +1012,8 @@
       (prefix-in t: (types numeric-predicates))))
   (provide (all-defined-out))
 
-  (define-syntax-rule (numeric/sc name body)
+  (define-syntax-rule (numeric/sc name body) (flat/sc #'body)
```
bennn: Why not `(flat/sc #'body 'name)` ?

bennn:
The "else" case for `Struct:` around line 700 can be simpler too, I think. From:

```
(flat/sc #`(flat-named-contract '#,(syntax-e pred?) (lambda (x) (#,pred? x))))
```

to

```
(flat/sc pred?)
```

samth:
A few motivations:

- @rfindler persuaded me a while back that we should show the actual contracts.
- This avoids a bunch of allocation.
- This allows inlining a bunch of contracts, reducing the number of pointless definitions.
- Combining these makes the generated contracts in the expanded code easier to read (the immediate motivation for me).

bennn:
ping @samth : were you planning to add more to this PR?

samth:
I think this is ready now.

PR #731, opened 2018-05-25
add contract-equivalent to restrict-typed->/c and restrict-typed-field/c

rfindler:
In https://github.com/racket/racket/commit/8ec3edaa9597223f152f5ca12799c8c669e471bb, `contract-equivalent?` was added to `racket/contract` but the methods haven't been implemented yet for typed racket. This commit has some easy ones that happen to make a big difference in the @bennn's gradual typing benchmark suite.

### DIFF
--- a/typed-racket-lib/typed-racket/utils/opaque-object.rkt
+++ b/typed-racket-lib/typed-racket/utils/opaque-object.rkt
@@ -172,15 +172,17 @@
   (define name (restrict-typed->/c-name ctc))
   (build-compound-type-name 'restrict-typed->/c name))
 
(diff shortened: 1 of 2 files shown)


samth:
Although, should there be a racket version bump/TR dependency on the new version? I see that the travis build fails because it needs your most recent Racket commit.

rfindler:
I don't have the energy to deal with a racket version bump for this. It will bump itself soon enough, imo.

bennn:
@samth I'm making a pull request for `opaque-object/c-equivalent`. I'll add a version bump to that

rfindler:
I think the version bump has to go in a different repo.

Issue #729, opened 2018-05-22
default hash argument produces #<unsafe-undefined>


jbclements:
It appears that TR doesn't handle optional hash table arguments correctly; in the example below, I get an error message about #<unsafe-undefined> in code that doesn't do anything involving unsafe code.

### What version of Racket are you using?

Updated head: 

Welcome to DrRacket, version 6.90.0.29--2018-05-14(-/f) [3m].
Language: typed/racket, with debugging [custom]; memory limit: 256 MB.

### What program did you run?

bad-tmp.rkt:
```racket
#lang typed/racket

(define-type My-Table (Immutable-HashTable Symbol Symbol))

;; provided function has an optional argument that defaults to (hash)
(define (provided-fun [structs : My-Table (hash)]) : Natural
  (inner-fun structs))

;; inner function uses hash table
(define (inner-fun [structs : My-Table]) : Natural
  (define bogo (hash-ref structs 'abc (λ () #f)))
  3)

;; call provided fun
(provided-fun)
```

### What should have happened?

call returns 3

### If you got an error message, please include it here.

```
hash-ref: contract violation
  expected: hash?
  given: #<unsafe-undefined>
  argument position: 1st
  other arguments...:
```

pnwamk:
Thanks, John!

It looks like the error occurs any time the default argument is not an _immediate_ value (which isn't too surprising since I think that's what some of the recent changes to optional arguments are specifically looking for).

This program suffers from the same problem:

```racket
#lang typed/racket

(define b (box 42))

;; provided function has an optional argument that defaults to (hash)
(define (provided-fun [n : Natural (unbox b)]) : Natural
  (inner-fun n))

;; inner function uses hash table
(define (inner-fun [n : Natural]) : Natural
  (+ n 0))

;; call provided fun
(provided-fun)
```

Also, adding `#:no-optimize` to the `#lang`-line makes the error go away.

mflatt:
The problem is also related to declaring a type for a specific optional argument, instead of for the optional-argument function overall.

pnwamk:
Working on a fix for this today, FYI.

pnwamk:
@samth Externally it has that type, but not internally. The "core" lambda that we're type checking just has type `(-> (U Integer Unsafe-Undefined) Integer)`, right?

PR #733, opened 2018-05-25
add unsafe-undef to opt args types when no expected type exists

pnwamk:
Fixes https://github.com/racket/typed-racket/issues/729.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/annotate-classes.rkt
+++ b/typed-racket-lib/typed-racket/base-env/annotate-classes.rkt
@@ -223,11 +223,18 @@
   #:attributes (form id default type kw)
   #:literal-sets (colon)
   (pattern [id:id default:expr]
-           #:with form #'([id default])
+           #:with form #`([#,(if (immediate-default? #'default)
+                                 (optional-immediate-arg-property #'id #t)
+                                 (optional-non-immediate-arg-property #'id #t))
+                           default])
            #:attr type #f
            #:attr kw #f)
   (pattern [id:id : type:expr default:expr]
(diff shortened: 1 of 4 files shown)


samth:
This looks sensible to me, but one worry -- if you import `lambda` from `racket/base`, don't you get the unsoundness back? Is there a way we could fail-closed instead of fail-open here?

pnwamk:
@samth I'm a little skeptical you could use that strategy to re-open this unsoundness.

Should we ship this PR and, if we can demonstrate an additional bug, address that then?

samth:
Can you add this test and reassure me?

```
(require (prefix-in r: racket/base))
(define b (box 42))
;; make sure Unsafe-Undefined is added to the core-lambda
;; argument type when a partial type annotation exists
(r:define (foo [#{n : Integer} (unbox b)])
  (+ n 0))

(unless (equal? 42 (foo))
  (error "oops! a bug!"))

**Comment by pnwamk** [comment]
That example fails to type check both in 6.12:

```
Type Checker: No function domains matched in function application:
Domains: False False
         Integer True
Arguments: Any True
 in: (r:define (foo (n (unbox b))) (+ n 0))
```

 and on HEAD w/ this PR:

```
Type Checker: type mismatch
  expected: Integer
  given: #<unsafe-undefined> in: (r:define (foo (n (unbox b))) (+ n 0))
```

I think basically it's not entering the "lambda with optional arguments" code path _because_ it's not a Typed Racket optional arg definition, and so it uses normal lambda type checking and notices that obviously things won't work as typed.

I'll add it as an expected failure with a note about why.

samth:
Ok, good. I am reassured.

PR #735, opened 2018-05-25
add contract-equivalent? for prefab/c

pnwamk:
I added some tests for `contract-equivalent?` for `prefab/c`, but more eyes to sanity check are always welcome.


typed-racket-lib/typed-racket/utils/prefab-c.rkt
```diff
@@ -160,6 +161,19 @@
...
+;; equivalent
+(define (prefab/c-equivalent this that)
+  (cond
+    [(not (and (base-prefab/c? that)
+               (equal? (base-prefab/c-key this) (base-prefab/c-key that))))
+     #f]
+    [else
+     (define these-field-contracts (base-prefab/c-field-contracts this))
+     (define those-field-contracts (base-prefab/c-field-contracts that))
+     (and (eqv? (length these-field-contracts)
```
bennn: Is this check redundant? --- does the `base-prefab/c-key`s being equal mean they'll have the same number of fields?

bennn: Hm, there's no check on the `mutability-bits` is that because we already checked the prefab keys are equal?

typed-racket-lib/typed-racket/utils/prefab-c.rkt
```diff
@@ -160,6 +161,19 @@
             [else
```
bennn: oh hey, the two `contract-stronger?`s above this line can be replaced with one `contract-equivalent?`

typed-racket-lib/typed-racket/utils/prefab-c.rkt
pnwamk: 1. No the check is not redundant; equal keys does not imply equal number of fields.

2. Mutability bits are not checked because if the prefab keys and number of fields are equal, it should suffice to check whether each field is equivalent (there's no _stronger_ check to perform if the fields are mutable, equivalent is already as strong as it gets and should be fine (unless I'm missing something obvious)).

pnwamk: Also on (2), if their keys and field lengths are equal, they necessarily have the same field mutability.

PR #736, opened 2018-05-27
simplify opt-arg function types

pnwamk:


samth:
I think merging this broke the `fail/wrong-arity2.rkt` test, so it will probably have to be reverted.

PR #737, opened 2018-05-28
refactor refinement code, add more basic refinements

pnwamk:
This cleans up and moves the code which adds basic refinements to linear integer arithmetic functions provided by `racket/base` and adds refinements for a few more functions. Instead of having an ever-growing syntax parse clause which each thing we recognize, now we lookup function ids in a free-id-table to see if we have linear refinements for them. This also allows us to quickly determine if a function should be treated like a "dependent function" when we're checking it's arguments -- see changes to `tc-app-main.rkt`.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/integer-refinements.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/integer-refinements.rkt
@@ -1,0 +1,228 @@
+#lang racket/base
+
+(require "../utils/utils.rkt"
+         (prefix-in c: (contract-req))
+         (types abbrev subtype numeric-tower prop-ops)
+         (types tc-result type-table)
+         racket/match
+         syntax/private/id-table
+         syntax/parse
+         (for-syntax racket/base))
+
+(provide/cond-contract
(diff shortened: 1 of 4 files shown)


typed-racket-lib/typed-racket/typecheck/integer-refinements.rkt
```diff
@@ -0,0 +1,199 @@
...
+(define (equality-function supported-type)
+  (λ (args-stx result)
+    (syntax-parse args-stx
+      [((~var e1 (w/type supported-type)) (~var e2 (w/type Univ)))
+       (define p (-eq (attribute e1.obj) (attribute e2.obj)))
+       (add-p/not-p result p)]
+      [((~var e1 (w/type Univ)) (~var e2 (w/type supported-type)))
+       (define p (-eq (attribute e1.obj) (attribute e2.obj)))
+       (add-p/not-p result p)]
+      [_  result])))
```
pnwamk: change to correctly handle if other argument was not a fixnum/integer/etc

PR #738, opened 2018-05-28
avoid infinite looping in subtyping and when extending an environment

pnwamk:
I was tinkering with refinements yesterday and hit an infinite loop in subtyping and environment extension. Basically the `env+` metafunction recurs when it sees that new info has been derived, and it was getting into a loop where it kept "learning" the same "new" info over and over. These changes fixed it, BUT, I can't find the snippet that was causing the bug >_<.

So... here's the fix... I'll see if I can find the test case again...


pnwamk:
Okay -- test added, minor bug fixed. If we pass integration testing I think it's good to go.

PR #741, opened 2018-06-03
tc-hash: better inference for hashes in syntax objects

bennn:
Improve type inference for hash values embedded in syntax objects

- if embedded value is a weak hash, give a `Weak-HashTable` type
- if embedded value is not-immutable and not-weak, give a
  `Mutable-HashTable` type

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/signatures.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/signatures.rkt
@@ -34,7 +34,7 @@
 
 (define-signature tc-literal^
   ([cond-contracted tc-literal (->* (syntax?) ((or/c Type? #f)) Type?)]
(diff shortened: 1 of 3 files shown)


samth:
When @pnwamk and I were talking about this, we decided there wasn't any point in supporting this use of 3d syntax in TR. Do you have a use case?

bennn:
No I don't have a use-case.

I just have the example program in the pull request --- I expected that program to typecheck.

Is there something missing in this pull request that would be needed to support non-immutable hashes in a syntax object? If not, I don't see why TR shouldn't support this.

(And if TR isn't going to support this use of 3d syntax, is that choice documented anywhere?)

samth:
I'm ok with merging this if you fix the conflicts.

PR #742, opened 2018-06-03
add contract-equivalent? for sealing->/c

bennn:
- fix bug: stop ignoring 'subset?' checks in sealing-contract-stronger?
- add contract-equivalent? for sealing contracts
- add some unit tests

### DIFF
--- a/typed-racket-lib/typed-racket/utils/sealing-contract.rkt
+++ b/typed-racket-lib/typed-racket/utils/sealing-contract.rkt
@@ -7,6 +7,7 @@
 (require racket/class
          racket/contract
          racket/match
+         racket/set
          (for-syntax racket/base
                      syntax/parse))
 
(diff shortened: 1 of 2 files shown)


bennn:
can I merge this for racket 7?

typed-racket-lib/typed-racket/utils/sealing-contract.rkt
```diff
@@ -7,6 +7,7 @@
 (require racket/class
          racket/contract
          racket/match
+         racket/set
```
samth: One question -- where exactly are the sets coming from in this code?

bennn: Good question. The code is using lists, I think these list represent sets (of method/field/init-field names), so I imported `subset?` and `set=?`

bennn: This file could just ask for ordered lists and use `equal?`

samth: Oh, this uses lists as sets via generics? I would prefer to avoid generic dispatch here.

bennn: Ok, I'll remove the generic dispatch

bennn:
daf30dd8e153a538942a4e39c0e8b24ddc40ddf9

PR #745, opened 2018-06-07
improve in-sorted-dict and use it more

pnwamk:


pnwamk:
@samth I think I made the changes you suggested today in person.

@akuhlens these changes may help the non-determinism you guys were noticing in Grift (if it doesn't make sure to let me know so we can't hunt that down ;-)

typed-racket-lib/typed-racket/env/env-utils.rkt
```diff
@@ -1,18 +1,20 @@
...
+(require "../utils/utils.rkt" racket/private/dict)
+(provide id< sorted-dict-map sorted-dict-for-each in-sorted-dict in-sorted-dict-keys)
 
 (define (id< a b) (symbol<? (syntax-e a) (syntax-e b)))
 
-(define (sorted-dict-map dict f <)
-  (define sorted (sort (dict-map dict cons) (λ (x y) (< (car x) (car y)))))
-  (map (lambda (a) (f (car a) (cdr a))) sorted))
+(define-syntax-rule (in-sorted-dict dict <?)
+  (in-assoc (sort (dict->list dict) (λ (entry1 entry2) (<? (car entry1) (car entry2))))))
```
bennn: Would it be better to`(sort .... <? #:key car)`?

pnwamk: In past benchmarking w/ sort, I found not using `#:key` and just providing a richer comparison function to always be faster... so you won't find `#:key` in any of my TR code.

samth:
Can you try running some program that takes a long time in contract generation and see if it's slower/faster/the same?

pnwamk:
Absolutely -- I'm about to run the tr-performance benchmarks which include some Grift programs whose types are large/interesting enough that they spend a non-trivial amount of time in contract generation.

pnwamk:
The latest improvements seem to do okay in general, but programs that spend significant time in contract generation get hit, e.g.:

- [schml-interp-casts-help](https://github.com/pnwamk/tr-performance/tree/master/schml-interp-casts-help) appears to go from 6.7 seconds to 7.7 seconds on my machine.

- [schml-flatten-values](https://github.com/pnwamk/tr-performance/tree/master/schml-flatten-values) now _consistently_ locks up in contract generation instead of non-deterministically... yay?

pnwamk:
Okay, so this PR initially had two features:

1. Improve the code that helps us iterate free-id-tables in a sorted fashion
2. Get rid of some of the non-determinism that has affected [some files](https://github.com/pnwamk/tr-performance/tree/master/schml-flatten-values) in the past.

Item (1) I feel is achieved and a worthwhile improvement. I'm spent several hours tinkering with item (2) and do not feel much closer to understanding what is causing the issue I was trying to address.

So... I think this PR for now should address (1), and later we can take a stab at (2) again.

pnwamk:
Also, with the most recent commit, there shouldn't be any overhead added w.r.t. type checking or contract generation time. The only changes now in this PR use more direct/efficient mechanisms for sorting and iterating over free-id-tables.

PR #746, opened 2018-06-12
doc: cite Typed Racket papers

bennn:
Add a bibliography to the TR reference.

At first, I just wanted to link to POPL 2008, ICFP 2010, and SNAPL 2017. But after I got started, it made sense to just link everything I could think of.

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/ts-reference.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/ts-reference.scrbl
@@ -1,5 +1,7 @@
 #lang scribble/manual
 
+@require[(for-label racket/control)]
+
 @title[#:tag "top"]{The Typed Racket Reference}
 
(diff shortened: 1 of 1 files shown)


bennn:
"in-progress", because this is either:

- too short of a list, and we should add more before merging ... or
- too long of a list, and should cut back to the top 2/3 papers

pnwamk:
I like this idea and would be in favor of adding more papers (since I think often each one has different insights/details that help describe the whole system).

Also, I think _where possible_ we should give a more detailed (but still brief) "why you might be interested in this paper" one-liner for each paper. e.g. ICFP 2010 has the high-level description of how TR type checking _currently works_, PADL 2012 (not listed in the PR yet) describes how this system was extended to handle the numeric tower, PLDI 2016 describes how linear integer refinements fit into this framework, etc.

bennn:
I shortened the long intro text & added a note to most entries in the bibliography.

Issue #747, opened 2018-06-28
unsound numeric comparisons (NaN)


pnwamk:
This type checks and should not:

```racket
#lang typed/racket

(let ([x : Negative-Fixnum -42]
      [y : Positive-Float +nan.0])
  (if (< x y)
      (error "should be unreachable")
      42))
;; errors in what should be unreachable code because
;; TR thinks the comparison must produce `#true`
```

This really is the same issue as https://github.com/racket/typed-racket/issues/112 -- our fix for that issue (https://github.com/racket/typed-racket/commit/52b450758524464213bf341785db007a10719b18)  removed some of the problem, but it should have gone further. i.e. even for the integer use cases, [this](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/base-env/base-env-numeric.rkt#L607) type for `<` (generated by the helper function `<-type-pattern`) is unsound.

i.e. cases such as `(-> base -NonNegReal B : (-PS* -tt (-is-type 0 non-neg)))` (i.e. which include a comparison against a type which includes NaN) either need to 

a) omit NaN explicitly, e.g. have `-NonNegRealNoNaN` instead of `-NonNegReal` (because even when `base` does not include `NaN`, the rhs of that comparison does), or 

b) not have a false proposition.

Perhaps there's additional options, but those are the two obvious ones that come to mind. Clearly we also need to assess how each solution will affect existing programs.

P.S. Although the above example has annotations, it's actually quite trivial to get an example of the same problem using standard, unannotated arithmetic mixing floats and non-floats:

```racket
(let ([x -42]
      [y (+ 1.0 +nan.0)])
  (if (< x y)
      (error "should be unreachable")
      42))
;; errors in what should be unreachable code because
;; TR thinks the comparison must produce `#true`
```

PR #748, opened 2018-06-29
fix numeric comparisons for NaN

pnwamk:
Fixes https://github.com/racket/typed-racket/issues/747 in the way that is least likely to break any programs (I think).

Basically, any comparison which has a non-trivial negative proposition needs to not include `NaN` in the argument types (since `NaN` in a comparison always produces `#false`).

Still need to add some tests specific to the issue.


stamourv:
Looks fine.

PR #753, opened 2018-07-05
fix #569

### LINKED ISSUE #569: Unbound type 't' in let form
6.9
### What program did you run?
(let ([id (lambda #:forall (t) ([x : t]) x)])
  id)
#<procedure:id>
Type Checker: parse error in type;
 type name `t' is unbound in: t  

if I change let into letrec, the program will run correctly.

yjqww6:
This fixed #569 , I think there are more ideal ways to fix it, such as check the `typechecker:plambda` syntax properties before syntax-parse, but the tail-position^ looks like just a hack for racket/match, so I make this pull request.


samth:
Thanks! This is pretty clearly an improvement.

PR #759, opened 2018-08-09
doc: caveat with match + occurrence typing

bennn:
Document a thing about `match` and occurrence typing.

The `@tech{}` in here assumes https://github.com/racket/racket/pull/2224 is merged

@greghendershott do you remember other surprises with match that should go in the docs?

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
@@ -180,3 +180,25 @@ circumvent the protections of the type system.
 
 Contract boundaries installed for typed-untyped interaction may cause
 significant slowdowns. See @secref{contract-costs} for details.
+
+
+@section{Pattern Matching and Occurrence Typing}
+
+The @racket[match] macro does not support occurrence typing (specifically,
+negative propositions) as well as @racket[cond] and @racket[if] do.
+For example, the following function safely applies @racket[add1] to integers
+and @racket[string-length] to strings but the type checker rejects it:
+
+@examples[#:label #f #:eval the-eval
+  (eval:error
+    (define (f (x : (U String Integer))) : Integer
+      (match x
+        [(? integer?) (add1 x)]
+        [_ (string-length x)])))]
+
+The reason for the type error is that Typed Racket checks the expanded form
+of the pattern-match, and @racket[match] expands to code where the call to
+@racket[string-length] occurs before the @racket[integer?] test.
+This out-of-order expansion makes backtracking (via a
+@tech[#:doc '(lib "scribblings/reference/reference.scrbl")]{failure procedure})
+possible, but unfortunately causes the type checker reject more programs.
--- a/typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
@@ -184,8 +184,8 @@ significant slowdowns. See @secref{contract-costs} for details.
 
 @section{Pattern Matching and Occurrence Typing}
 
-The @racket[match] macro does not support occurrence typing (specifically,
-negative propositions) as well as @racket[cond] and @racket[if] do.
+The @racket[match] macro does not support occurrence typing as well as
+@racket[cond] and @racket[if] do.
 For example, the following function safely applies @racket[add1] to integers
 and @racket[string-length] to strings but the type checker rejects it:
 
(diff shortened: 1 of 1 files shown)


greghendershott:
It was a long time ago. I remember bring frustrated trying to use match on x-expressions. Not at a real computer now where I can try.

typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
```diff
@@ -180,3 +180,25 @@ circumvent the protections of the type system.
 
 Contract boundaries installed for typed-untyped interaction may cause
 significant slowdowns. See @secref{contract-costs} for details.
+
+
+@section{Pattern Matching and Occurrence Typing}
+
+The @racket[match] macro does not support occurrence typing (specifically,
```
samth: I think phrasing this to involve less technical type system detail would be better.

typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
```diff
@@ -180,3 +180,25 @@ circumvent the protections of the type system.
...
+@examples[#:label #f #:eval the-eval
+  (eval:error
+    (define (f (x : (U String Integer))) : Integer
+      (match x
+        [(? integer?) (add1 x)]
+        [_ (string-length x)])))]
+
+The reason for the type error is that Typed Racket checks the expanded form
+of the pattern-match, and @racket[match] expands to code where the call to
+@racket[string-length] occurs before the @racket[integer?] test.
```
samth: I think "before" is a bad word here. Typed Racket can't see the relationship between the test and the call.

typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
pnwamk: I threw together something that I think is a little less technical but still detailed (albeit quickly in emacs... so it's probably syntactically and semantically broken here and there... sorry):

```racket
The expansion of the @racket[match] is difficult for Typed Racket
to completely reasoning about. In particular, the type of an
identifier is often @emph{not} updated to reflect the fact that
a "previous" pattern failed to match because of how the code is
expanded. For example, in the following function, the type
checker is "unaware" that in the last clause the
@racket[integer?]  predicate has already failed to match on the
value for @racket[x], and so @racket[(string-length x)] fails to
type check:

@examples[#:label #f #:eval the-eval
  (eval:error
    (define (f (x : (U String Integer))) : Integer
      (match x
        [(? integer?) (add1 x)]
        [_ (string-length x)])))]

Whereas a similar @racket[cond] expression does type check (since
it expands into much simpler code):


@examples[#:label #f #:eval the-eval
  (define (f (x : (U String Integer))) : Integer
      (cond
        [(integer? x) (add1 x)]
        [else (string-length x)]))]


One work around is to simply not relying on a catch-all "else"
clause in the match needing to know which previous patterns
have failed to match:

@examples[#:label #f #:eval the-eval
  (define (f (x : (U String Integer))) : Integer
      (match x
        [(? integer?) (add1 x)]
        [(? string?) (string-length x)]))]

However, note that @racket[match] @emph{always} inserts a
catch-all failure clause if one is not provided, and so the type
checker will not inform the programmer if all cases are not
properly check for, e.g.:

@examples[#:label #f #:eval the-eval
  (define (f (x : (U String Integer))) : Integer
      (match x
             [(? integer?) (add1 x)]))

  (f "oops no clause for me")]
```

pnwamk: Feel free to use or ignore =)

bennn: Thanks. I like adding the work-around with `string?`. For the rest ... its talking about too many issues for me. I'd rather keep this guide section focused on: "if you expected something like this `match` to work, it doesn't and heres a quick reason why". 

I'll think about changing the first sentence in each paragraph of this PR to be more like the first sentence in this suggestion.

samth: I have to say that I prefer @pnwamk's version.  In particular, it provides the work-around(s) that people want once they learn that this doesn't work.

bennn: @pnwamk  do you want to open a pull request, and I'll close this? (Or I can edit this with a cleaned-up version of yours, but I probably won't get to it this weekend and I think you should have credit for the commit.)

pnwamk: Sure, I'll get it cleaned up and submit it.

pnwamk:
See https://github.com/racket/typed-racket/pull/761.

Issue #768, opened 2018-09-06
typed/json doesn't account for json-null


LiberalArtist:
The `json` library is tricky to type (or protect by contract) because most functions take an optional `#:null` argument that defaults to the current value of `(json-null)`, which may be of any type and is recognized by `eq?`. The default value is `'null`, which of course works as a type for `JSExpr`.

The current `typed/json` interface partially addresses this by not permitting `#:null` arguments. However, it doesn't go far enough, and it is possible to provoke errors from untyped code that blame the interface:
```racket
#lang racket

(module flawed typed/racket
  (require typed/json)
  (provide parse)
  (: parse (-> String JSExpr))
  (define (parse str)
    (string->jsexpr str)))

(require json 'flawed)

(parameterize ([json-null '|something odd|])
  (parse "null"))
```

samth:
Fortunately, this can't provoke unsoundness in Typed Racket, since the `typed/json` library is just another Typed Racket program. I don't see a way to do better here, unfortunately -- do you have a suggestion? If not, we should just document that this won't work.

mfelleisen:
Why is _this_ not a general problem, where _this_ is 

 (1) require typed/foo a wrapper for foo 
 (2) require foo, too 

Use functions from the latter to circumvent the former’s invariants and cause havoc?

mfelleisen:
@LiberalArtist wrote "it doesn't go far enough, and it is possible to provoke errors from untyped code that blame the interface:"

This is the beauty of "gradual typing" and is an incredibly beautiful example of how the Wadler slogan  "blame the untyped code" is complete nonsense. On occasion it's the type that we assign to untyped code that's to be blamed.

samth:
@mfelleisen We shouldn't let ourselves off so easily: "the typed portion of the program can’t go wrong;
only the untyped portions may trigger run-time exceptions" is a quote from our paper.

LiberalArtist:
One alternative I've thought of is that `typed/json` could implement wrapper functions that always supply `#:null 'null`. That would make the variants exported by `typed/json` actually have the types they currently claim to, and it would fit will the current decision to always use the type `'null` in `JSExpr`.

For example:
```racket
#lang racket

(module ignore-parameter typed/racket
  (provide JSExpr
           (rename-out
            [string->jsexpr* string->jsexpr]))
  (require (only-in typed/json JSExpr))
  (require/typed
   json
   [string->jsexpr (-> String #:null 'null JSExpr)])
[...]

PR #771, opened 2018-09-15
typed/json: handle json-null

LiberalArtist:
Previously, untyped code could use the `json-null` parameter to cause
the functions from `typed/json` not to obey their declared types.
This commit prevents that problem by having the typed wrapper functions
explicitly pass `'null` as the `#:null` argument.

If this approach sees promising, I can add documentation (and probably tests, somewhere?).

Closes https://github.com/racket/typed-racket/issues/768


samth:
This looks great!

LiberalArtist:
Would the way to add tests be to create a module "typed-racket-test/unit-tests/json.rkt" and add it to the list in "typed-racket-test/unit-tests/all-tests.rkt"?

LiberalArtist:
I've updated the pull request to include tests and documentation.

PR #776, opened 2018-09-30
Fixes for bugs reported at RacketCon.

samth:


typed-racket-lib/typed-racket/types/abbrev.rkt
```diff
@@ -300,3 +300,11 @@
            [_ rep])))
      (values new-type props)]))
 
+
+
+(define (-list-set s) (Un (-lst s) (-set s)))
+
```
pnwamk: Would `-list-or-set` be clearer for future maintainers?

samth: Yes, I think so.

typed-racket-lib/typed-racket/types/abbrev.rkt
```diff
@@ -300,3 +300,11 @@
...
+(define-syntax-rule (set-abs s e)
+  (cl->*
```
pnwamk: A brief comment on why `set-abs`  exists for posterity?

PR #784, opened 2018-10-25
cover polymorphic types when printing

pnwamk:
Previously the algorithm which tried to tidy-up large union types by covering them with type names in the current environment only considered monomorphic types. This adds support for polymorphic types as well.

e.g., previously hash-set had a very verbose type when printed because the `HashTable` abstraction wasn't considered:

```
> hash-set
- : (All (a b)
      (-> (U (Immutable-HashTable a b)
             (Mutable-HashTable a b)
             (Weak-HashTable a b))
          a
          b
          (Immutable-HashTable a b)))
#<procedure:hash-set>
```

and now it is:

```
> hash-set
- : (All (a b) (-> HashTable a b) a b (Immutable-HashTable a b)))
#<procedure:hash-set>
```

This of course also works for any user-defined types that are polymorphic (yay!).

### DIFF
--- a/typed-racket-lib/typed-racket/types/printer.rkt
+++ b/typed-racket-lib/typed-racket/types/printer.rkt
@@ -7,6 +7,7 @@
          racket/pretty
          racket/list
          racket/set
+         (except-in (path-up "utils/utils.rkt") infer)
          (path-up "rep/type-rep.rkt" "rep/prop-rep.rkt" "rep/object-rep.rkt"
                   "rep/core-rep.rkt" "rep/values-rep.rkt" "rep/fme-utils.rkt"
                   "rep/rep-utils.rkt" "rep/free-ids.rkt"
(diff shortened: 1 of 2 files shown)


pnwamk:
Ah, I forgot about `filter-map`... I'll see if I can replace the let-loop with something clearer while fixing the errors.

typed-racket-lib/typed-racket/types/printer.rkt
```diff
@@ -287,32 +287,28 @@
...
-           [_ (loop remaining valid-names)])])))
+    ;; We keep only the unions and (instantiated) polymorphic unions
+    ;; which are subtypes of t. It's no use attempting to cover t with
+    ;; things that go outside of t.
+    (filter-map
+     (match-lambda
+       [(and name/type (cons name t*))
+        (match t*
+          [(or (? Union?) (? BaseUnion?))
+           #:when (not (member name ignored-names))
```
samth: Can't these two identical `#:when` clauses be moved to the top match pattern?

PR #785, opened 2018-10-26
update Sequenceof docs

pnwamk:
The docs have been behind the implementation for a while---this updates the docs so the two are in sync.


pnwamk:
@samth I waffled over wording w.r.t. describing the "values" produced at each iteration and went for terse + examples. Let me know if you prefer a different description.

PR #788, opened 2018-12-03
fix 506: parent structs error when #:type-name is provided

### LINKED ISSUE #506: parent structs error when `#:type-name` is provided
6.0.8.2 (Current TR head as of yesterday)

### What program did you run?
```
#lang typed/racket
(struct a ([f : Integer])
  #:type-name A)
(struct b a ())
```
The program should have behaved like this one:

```
#lang typed/racket
(struct a ([f : Integer]))
(struct b a ())
```

```
Type Checker: parent type not a valid structure name: a in: a
```

default-kramer:
Fix #506 

Note - this change means that the struct name can now be used as a type name. For example, consider `(struct a () #:type-name A)`

**Today:** Only `A` can be used as a type name. A new type named `a` cannot be created (the identifier is already in use.)

**With this change:** Both `A` and `a` can be used as type names to refer to this struct.

If this is a problem, my only idea is to introduce a new lookup table to map `a` to `A`. (There might be a better solution, but I don't know the codebase well enough.)

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
@@ -72,7 +72,6 @@
     [v:parent
       (if (attribute v.par)
           (let* ([parent0 (parse-type #'v.par)]
-                 ;; TODO ensure this is a struct
                  [parent (let loop ((parent parent0))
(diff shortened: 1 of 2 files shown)


samth:
Awesome!

PR #791, opened 2018-12-30
Introduce SequenceTop

lwhjp:
While implementing PolyDot sequences, I noticed that we'd have trouble generating contracts for unconstrained sequences. So here is a `SequenceTop` type to handle that case, as a prelude.

The expression
```
(sequence-length (in-parallel '(1 2 3) '(4 5 6)))
```
now typechecks.

This is my first time fiddling with TR internals, so apologies if this is totally the wrong approach!

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
@@ -519,6 +519,13 @@ or sets that are implemented using @racket[gen:set].
 is a sequence of strings, @racket[(Sequenceof Number String)] is a sequence which
 produces two values---a number and a string---on each iteration, etc.}
 
+@defidform[SequenceTop]{is the type of a @rtech{sequence} with unknown element
+  type and is the supertype of all sequences. This type typically
+  appears in programs via the combination of ocurrence typing ang
+  @racket[sequence?].
+@ex[(lambda: ([x : Any]) (if (sequence? x) x (error "not a sequence!")))]
+}
+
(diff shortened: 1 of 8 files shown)


typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -610,6 +610,16 @@
                (nbits-overlap? nbits bits)
                (bbits-overlap? bbits bits))
            A)]
+     [(SequenceTop:)
```
samth: I'm a little concerned about the duplicate code here. Can we combine these cases?

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -519,6 +519,13 @@ or sets that are implemented using @racket[gen:set].
 is a sequence of strings, @racket[(Sequenceof Number String)] is a sequence which
 produces two values---a number and a string---on each iteration, etc.}
 
+@defidform[SequenceTop]{is the type of a @rtech{sequence} with unknown element
```
samth: This should have an `@history` annotation.

lwhjp:
Thanks; I'll address those points and update.

samth:
Looks great. My one final concern is that we might get something about multiple valued sequences wrong -- can you add a test that fails when we try to get a single value from a multiple valued sequence using `sequence?`.

lwhjp:
(Sorry about the force-pushes; I messed up my git branch)

PR #795, opened 2019-01-20
Prevent failure generating constraints for values without psets

lwhjp:
I found this while working on dotted sequence types, but there is a simple case to reproduce:
```
(define #:forall (A ...) (f [v : (List A ... A)])
  (apply values v))
(f '(1 2 3))
```

When typechecked as a module form, the last line has expected type `AnyValues`, which ends up in a call `cgen` with types `(Values A ... A)` and `AnyValues`. The former (as an instance of ValuesDots with no non-var prefix) has an empty list of psets, which is joined to form a "null" cset (equal to `no-cset`), which results in a typecheck failure.

This change includes the empty cset in the join so that this case passes even when there are no psets, which seems to me to be the correct behavior.

Incidentally, if the example above is entered into the toplevel REPL instead of as a module, `tc-toplevel-form` passes `#f` as the expected type to `infer`, and typecheck succeeds with a default empty expected-cset.

### DIFF
--- a/typed-racket-lib/typed-racket/infer/infer-unit.rkt
+++ b/typed-racket-lib/typed-racket/infer/infer-unit.rkt
@@ -480,12 +480,14 @@
               (ValuesDots: (list (Result: _ psets _) ...) _ _))
           (AnyValues: q))
          (cset-join
-          (for*/list ([pset (in-list psets)]
-                      [cs (in-value (% cset-meet
-                                       (cgen/prop context (PropSet-thn pset) q)
-                                       (cgen/prop context (PropSet-els pset) q)))]
-                      #:when cs)
-            cs))]
+          (cons
+           empty
+           (for*/list ([pset (in-list psets)]
+                       [cs (in-value (% cset-meet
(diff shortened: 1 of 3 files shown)


typed-racket-lib/typed-racket/infer/infer-unit.rkt
```diff
@@ -480,12 +480,14 @@
               (ValuesDots: (list (Result: _ psets _) ...) _ _))
           (AnyValues: q))
          (cset-join
```
bennn: Is this still correct when there `psets` is non-empty?

Either way, I think this code would be easier to understand as an if-statement:

```
(if (null? psets)
  empty
  (cset-join ....))
```

lwhjp: My understanding is that the empty cset (corresponding to ∅ in the theory) serves as an identity element to `cset-meet` so joining it should have no effect - but using an `if` instead will certainly save some unnecessary steps.

PR #801, opened 2019-02-18
doc: add example for is-a? occurrence typing limitation

### LINKED ISSUE #792: `is-a?` doesn't support occurrence typing
For example:
```
#lang typed/racket
(define with-field%
  (class object%
    (init-field that)
    (super-new)))

(define (maybe-that o)
  (if (is-a? o with-field%)
      (get-field that o)
      #f))

(maybe-that (make-object with-field% 'bingo))
```

This fails typecheck with:
```
Type Checker: type mismatch;
 expected an object value for get-field
  given: Any in: (get-field that o)
```
Whereas I'd expect the last line to return `bingo`.

Presumably `is-a?` should have a type something like `(All (A) (-> Any A Boolean : (Instance A)))`, if that expression makes sense in TR.

bennn:
Add a note about `is-a?` and occurrence typing to the "Caveats and Limitations" section of the guide.

Closes #792

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
@@ -243,3 +243,98 @@ anticipate with their pattern matching, e.g.:
 
   (eval:error
    (size 42))]
+
+
+@section{@racket[is-a?] and Occurrence Typing}
+
+Typed Racket does not use the @racket[is-a?] predicate to refine object types
+ because the target object may have been created in untyped code and
+ @racket[is-a?] does not check the types of fields and methods.
+
+For example, the code below defines a class type @racket[Pizza%], a subclass
+ type @racket[Sauce-Pizza%], and a function @racket[get-sauce] (this
+ function contains a type error).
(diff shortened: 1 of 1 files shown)


typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
```diff
@@ -243,3 +243,98 @@ anticipate with their pattern matching, e.g.:
...
+ @racket[is-a?] does not check the types of fields and methods.
+
+For example, the code below defines a class type @racket[Pizza%], a subclass
+ type @racket[Sauce-Pizza%], and a function @racket[get-sauce] (this
+ function contains a type error).
+The @racket[get-sauce] function uses @racket[is-a?] to test the class of its
+ argument; if the test is successful, the function expects the argument to have
+ a field named @racket[topping] that contains a value of type @racket[Sauce].
+
+@codeblock{
```
bennn: I wanted this to be an `@examples`, but I couldn't get the error message to point to the right expression. I'd see this:

```
eval:33:0: Type Checker: type mismatch
expected: Sauce
given: Any
in: bbq
```

instead of `in: (get-field topping pizza)`

PR #803, opened 2019-02-19
[RFC] Support type checking struct properties in struct definitions

capfredf:


rfcs/0000-template.md
```diff
@@ -49,7 +49,6 @@ A few examples of what this can include are:
 - Does this feature exist in other programming languages and what experience have their community had?
 - Papers: Are there published papers, books, great blog posts, etc that discuss this? Be _specific_!
 
-
 # Unresolved questions
```
pnwamk: Probably better to not have a this arbitrary whitespace change in the template (people might wonder what changed in the template, take the time to check, and discover the answer is "nothing").

rfcs/text/0002-type-ann-for-struct-properties-in-structs.md
```diff
@@ -0,0 +1,97 @@
+- Feature Name: type-annotations-for-struct-properties-in-structs
+- Start Date: TODO
+- RFC PR: (leave this empty)
+- Feature Commit(s): (leave this empty)
+
+# Summary
+
+Typecheck a value expression for a struct property in a struct definition with
+expected type information in the property.
+
```
pnwamk: My knee-jerk reaction is that this summary is too specific and will be confusing for anyone not intimately familiar with the development story behind all of this (which shouldn't be relevant, IMHO). I.e., I think the fact that there is some (undocumented?) support for struct properties without type checking currently shouldn't be conveyed in the RFC title and summary. I think it would better communicate the intent to just say something like this RFC aims to "Provide initial support for type checking some struct properties" and then later on in the fine grained details we can discuss some of the nuance if that adds value. Just my two bits.

pnwamk: Or perhaps I'm misreading it and imposing some of my own understanding of this work, but either way I still think having a simpler and more clear summary would add value.

rfcs/text/0002-type-ann-for-struct-properties-in-structs.md
```diff
@@ -0,0 +1,97 @@
...
+# Motivation
+By adding types for struct properties, we further complete the type annotations
+for struct types.
+
+# Guide-level explanation
```
pnwamk: This is (IMO) a potentially tricky subject with all the subtle details. I played around with the guide-level wording you had to try and remove potential ambiguities. I'm just going to insert them below the associated paragraph; please feel free to completely ignore or use any of them as you see fit!

rfcs/text/0002-type-ann-for-struct-properties-in-structs.md
```diff
@@ -0,0 +1,97 @@
...
+struct type properties are type-annotated with `Struct-Property`, where the only
+argument is the expected type of property value expressions. Here is an example
+of `prop:input-port`:
+
```
pnwamk: To type-annotate a struct property we use the `Struct-Property`
form, which takes a single argument to describe the type of
that property's values. For example, here is a type-annotation
for the `prop:input-port` struct property:

rfcs/text/0002-type-ann-for-struct-properties-in-structs.md
```diff
@@ -0,0 +1,97 @@
...
+(: prop:input-port (Struct-Property (U Input-Port Nonnegative-Integer)))
+```
+
+This specifies the value in structs for `prop:input-port` to be either an input
+port or a nonngeative integer. Usually a value for a struct property is a
+function which takes the struct instance that the property would be extracted
+from as the first argument. To construct such struct property, we use `Self` to
+specify type of expected instances in the type annotation. Here is an example of
+`prop:custom-write`:
+
```
pnwamk: This specifies the value of a struct's `prop:input-port` property
to be either an input port or a nonngeative integer. Many struct
properties, however, are intended to contain a function which
then takes the associated struct instance as an argument. To
construct an appropriate type for such a function, we use `Self`
to specify which argument must be the associated struct
instance. For example, here is a type annotation for the
`prop:custom-write` struct property:

rfcs/text/0002-type-ann-for-struct-properties-in-structs.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+This specifies the value in structs for `custom-write` to have the function type
+`(-> Self Port Boolean Void)` where `Self` must be the struct instance that the
+accessor to `custom-write` is invoked with. In the definition of a struct where
+we attach properties that will operate on the struct instances, we specify the
+type of the value using the struct type to match `Self` in the type annotation
+of the struct property. Here is an example of using `prop:custom-write` in the
+definition of the struct `Point` :
+
```
pnwamk: This specifies that the value in a struct's `prop:custom-write`
property should have the function type `(-> Self Port Boolean
Void)`, i.e. a function which will be given the associated struct
as its first argument. Note that `Self` is only a valid type
within a `Struct-Property` declaration. When defining a struct
with a struct property whose type features `Self`, we instead use
the name of the struct currently being defined wherever `Self`
appeared in the struct property's type declaration. For example,
here is a struct `Point` which defines the `prop:custom-write`
property mentioned above):

rfcs/text/0002-type-ann-for-struct-properties-in-structs.md
```diff
@@ -0,0 +1,97 @@
...
+        (lambda ([self : Point]
+                 [p : Port]
+                 [mode : Boolean])
+             (print (cons (Point-x self) (Point-y self) p))))
+
+```
+
+Note that using `Self` in value expressions for properties like
+`prop:custom-write` fails the typechecker.
+
```
pnwamk: (In my tweaks I believe I incorporated this warning/note above within the paragraph, just FYI)

rfcs/text/0002-type-ann-for-struct-properties-in-structs.md
```diff
@@ -0,0 +1,97 @@
...
+
+# Reference-level explanation
+## Struct Type Property Types
+`(Struct-Property ty)` is a new structral type, where its argument can be any
+valid type in Typed Racket. User aren't allowed to use this form as the type
+constructor to define a struct property type. All built-in struct type
+properties are type-annotated.
+
```
pnwamk: "Users aren't allowed ..." this is intended to be a temporary restriction, am I remembering that right? If so it might be useful to make that clear, i.e. "Initially/currently users cannot use this form..." or "currently this form is reserved for ..." etc. And I wonder if such a note would be useful in the guide level explanation? I dunno.

rfcs/text/0002-type-ann-for-struct-properties-in-structs.md
```diff
@@ -0,0 +1,97 @@
...
+# Drawbacks and Alternatives
+[drawbacks]: #drawbacks
+
+## Drawbacks
+Adds complexity to the type checker.
+
+## Backward Compatibility
+The old `Struct-Type-Property` becomes an alias to `(Struct-Property Any)`
+
```
pnwamk: I think we had a brief discussion about how to handle legacy code in which struct properties were not type checked at all. Did you guys decide on a solution (and the above `Backward Compatibility` note reflects that) or should this RFC describe the options?

rfcs/text/0002-type-ann-for-struct-properties-in-structs.md
```diff
@@ -0,0 +1,97 @@
...
+# Prior art
+[prior-art]: #prior-art
+
+As of Racket 7.2, most struct properties doesn't work in Typed Racket. Type
+information of struct type properties is incomplete. For example, the type of
+`prop:custom-write` is just `Struct-Type-Property`, which contains no specific type
+information it expects. Also, there is no type-checking property values against
+properties.
+
```
pnwamk: Would it be clearer there at the end of the prior art paragraph to say "Also, provided struct properties within struct definitions are not type checked in versions <= 7.2"?

PR #804, opened 2019-02-19
typechecking struct propety values in struct definitions

capfredf:
Related to #803.


typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
```diff
@@ -76,14 +76,16 @@
...
+  #:attributes (mutable prefab type-only maker extra-maker [prop 1] [prop-val 1])
  (pattern
   (~seq
     (~or (~optional (~and #:mutable (~bind (mutable #t))))
          (~optional (~and #:prefab (~bind (prefab #t))))
          (~optional (~and #:type-only (~bind (type-only #t))))
          (~optional (~seq #:extra-maker extra-maker))
-         (~optional (~seq #:maker maker))) ...)))
+         (~optional (~seq #:maker maker))
+         (~seq #:property prop prop-val))
```
capfredf: This may not need `prop-val`.

typed-racket-lib/typed-racket/typecheck/internal-forms.rkt
```diff
@@ -92,15 +94,18 @@
...
   (pattern ((~optional (tvars:id ...) #:defaults (((tvars 1) null)))
             nm:struct-name type-name:id ([fields:id : types:expr] ...) options:dtsi-fields)
            #:attr name #'nm.nm
            #:attr mutable (attribute options.mutable)
            #:attr prefab (attribute options.prefab)
            #:attr type-only (attribute options.type-only)
            #:attr maker (or (attribute options.maker) #'nm.nm)
-           #:attr extra-maker (attribute options.extra-maker)))
+           #:attr extra-maker (attribute options.extra-maker)
+           #:attr properties (for/list ([p (attribute options.prop)]
```
capfredf: Don't need the prop-val here.

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -64,6 +68,35 @@
     [(~or t:typed-struct t:typed-struct/exec)
      #'t.type-name]))
 
+(define-syntax-class expaneded-props
```
capfredf: spelling

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -64,6 +68,35 @@
...
+  #:literals (null list #%plain-app)
+  (pattern null #:attr prop-names null #:attr prop-vals null)
+  (pattern (#%plain-app list (#%plain-app cons pn pv) ...)
```
capfredf: This could be `(#%plain-app cons prop-names prop-vals) ...`

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -64,6 +68,35 @@
...
+         (#%plain-app values args-v ...)))
+     (for ([p (in-list (attribute props.prop-names))]
+           [pval (in-list (attribute props.prop-vals))])
+       (match (single-value p)
+         [(tc-result1: (StructProperty: ty))
+          (define sty (lookup-type-alias name parse-type))
+          (match-define (F: var) -Self)
+          (tc-expr/check pval (ret (subst var sty ty)))]
+         [(tc-result1: ty)
+          (tc-error "expected a struct type property but got something else")]))]
```
capfredf: Improve error message.

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -433,41 +470,43 @@
...
          (define sty (mk/inner-struct-type names desc concrete-parent))
-         
+
          (parsed-struct sty names desc (struct-info-property nm/par) type-only)]))
 
 ;; register a struct type
 ;; convenience function for built-in structs
 ;; tc/builtin-struct : identifier Maybe[identifier] Listof[identifier]
-;;                     Listof[Type] Maybe[identifier] Listof[Type]
+;;                     Listof[Type] Maybe[identifier] Listof[Type] ListOf[Pairof[identifier, Type]]
```
capfredf: Only need a list of the property names, but they are expressions (not just ids).

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -433,41 +470,43 @@
...
 ;;                     -> void
 ;; FIXME - figure out how to make this lots lazier
-(define/cond-contract (tc/builtin-struct nm parent fld-names tys kernel-maker)
-     (c:-> identifier? (c:or/c #f identifier?) (c:listof identifier?)
-           (c:listof Type?) (c:or/c #f identifier?)
-           c:any/c)
+(define/cond-contract (tc/builtin-struct nm parent fld-names tys kernel-maker properties)
```
capfredf: Why does this need another argument?

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -5,7 +5,7 @@
          racket/list racket/match racket/sequence
          (prefix-in c: (contract-req))
          (rep core-rep type-rep values-rep)
-         (types utils abbrev type-table struct-table resolve)
+         (types utils abbrev type-table struct-table resolve substitute)
```
capfredf: Is substitute still needed?

typed-racket-test/fail/bad-struct-prop.rkt
```diff
@@ -0,0 +1,4 @@
+#lang typed/racket
```
capfredf: Add an error message predicate.

typed-racket-test/unit-tests/init-env-tests.rkt
```diff
@@ -52,13 +52,13 @@
         (convert-type -field)
         '-field)
       (check-equal?
-        (convert-type (make-StructType (make-Struct #'foo #f null #f #f #'foo?)))
+        (convert-type (make-StructType (make-Struct #'foo #f null #f #f #'foo? null)))
         '(make-StructType
-          (make-Struct (quote-syntax foo) #f (list) #f #f (quote-syntax foo?))))
+          (make-Struct (quote-syntax foo) #f (list) #f #f (quote-syntax foo?) (list))))
```
capfredf: Use `null` consistently.

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -2289,7 +2289,8 @@
 [special-comment-value (-> -Special-Comment Univ)]
 
 ;; Section 13.8
-[prop:custom-write -Struct-Type-Property]
+#;[prop:custom-write -Struct-Type-Property]
+[prop:custom-write (-struct-property (-> -Self -Output-Port B Univ))]
```
pnwamk: Do we want to leave the commented out type there?

typed-racket-lib/typed-racket/env/global-env.rkt
```diff
@@ -8,7 +8,7 @@
          "env-utils.rkt"
```
pnwamk: It seems like the code in this file is unchanged, but there are two extremely minor white space edits. Is it worth keeping these edits to muck up this file's history a little?

typed-racket-lib/typed-racket/env/type-alias-env.rkt
```diff
@@ -62,12 +62,12 @@
```
pnwamk: See comment for `.../global-env.rkt`.

typed-racket-lib/typed-racket/private/parse-type.rkt
```diff
@@ -138,6 +138,7 @@
 (define-literal-syntax-class #:for-label Vector)
 (define-literal-syntax-class #:for-label Struct)
 (define-literal-syntax-class #:for-label Struct-Type)
+#;(define-literal-syntax-class #:for-label StructProperty)
 (define-literal-syntax-class #:for-label Prefab)
```
pnwamk: Do we want to delete this commented out code?

typed-racket-lib/typed-racket/rep/base-types.rkt
```diff
@@ -170,7 +170,7 @@
   [-Subprocess Subprocess #'subprocess? subprocess?]
   [-Security-Guard Security-Guard #'security-guard? security-guard?]
   [-Thread-Group Thread-Group #'thread-group? thread-group?]
-  [-Struct-Type-Property Struct-Type-Property #'struct-type-property? struct-type-property?]
+  ;;[-Struct-Type-Property Struct-Type-Property #'struct-type-property? struct-type-property?]
   [-Impersonator-Property Impersonator-Property #'impersonator-property? impersonator-property?]
```
pnwamk: Keep or delete commented out code?

typed-racket-lib/typed-racket/rep/core-rep.rkt
```diff
@@ -208,4 +208,3 @@
    #:property prop:custom-print-quotable 'never
```
pnwamk: See comment for `.../global-env.rkt`.

typed-racket-lib/typed-racket/rep/type-mask.rkt
```diff
@@ -4,7 +4,7 @@
 ;; Type Masks
```
pnwamk: Is a `Self` really something that is never described by any other type mask? I.e., can't a `Self` be a struct?

I would have assumed `Self` would have the same type mask as structs (or similar).

typed-racket-lib/typed-racket/rep/type-mask.rkt
```diff
@@ -149,6 +149,8 @@
  mask:struct
  mask:prefab
  mask:struct-type
+ mask:self
+ mask:struct-property
  mask:syntax
```
pnwamk: I haven't put much thought into the validity of `mask:struct-property`.... I could believe it's disjoint from all the other descriptors. Might want to double check that.

Also, is it worth having it's own type mask? I do not know, but I just feel like they'll probably not be frequently encountered in type computations (but I could be wrong).

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -714,6 +714,8 @@
 ;; Structs
 ;;************************************************************
 
+(def-structural StructProperty ([elem #:invariant])
+  [#:mask mask:struct-property])
```
pnwamk: See comments in `type-mask.rkt`.

typed-racket-lib/typed-racket/rep/values-rep.rkt
```diff
@@ -60,7 +60,7 @@
         (combine-frees (map free-vars* (cons dty results))))]
```
pnwamk: See comment for `.../global-env.rkt`.

typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt
```diff
@@ -157,6 +157,7 @@
   ((listof/sc (#:covariant)) listof #:flat)
   ((cons/sc (#:covariant) (#:covariant)) cons/c #:flat)
   ((set/sc (#:covariant #:chaperone)) set/c #:flat)
+  ((struct-property/sc (#:invariant)) struct-type-property/c #:flat)
   ((immutable-vectorof/sc (#:covariant)) immutable-vectorof/c #:flat)
```
pnwamk: Is this always a flat contract? What if it's a function type within the struct-property? Also, off the top of my head I can't decide if `#:invariant` makes sense and what the implications of that would be vs covariant, so if you haven't already maybe double check that is fine.

typed-racket-lib/typed-racket/static-contracts/structures.rkt
```diff
@@ -4,7 +4,7 @@
```
pnwamk: See comment for `.../global-env.rkt`.

typed-racket-lib/typed-racket/typecheck/signatures.rkt
```diff
@@ -1,6 +1,6 @@
 #lang racket/base
```
pnwamk: See comment for `.../global-env.rkt`.

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -64,6 +68,35 @@
...
+         (#%plain-app values args-v ...)))
+     (for ([p (in-list (attribute props.prop-names))]
+           [pval (in-list (attribute props.prop-vals))])
+       (match (single-value p)
+         [(tc-result1: (StructProperty: ty))
+          (define sty (lookup-type-alias name parse-type))
+          (match-define (F: var) -Self)
+          (tc-expr/check pval (ret (subst var sty ty)))]
+         [(tc-result1: ty)
+          (tc-error "expected a struct type property but got ~a") ty]))]
```
pnwamk: is returning `ty` in the error case here right? Should it be the weird `Error` type instead?

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -271,6 +276,10 @@
        'no-type]
       ;; these forms we have been instructed to ignore
       [stx:ignore^
+       (cond
+         [(syntax-property #'stx 'tc-struct)
```
pnwamk: Is it worth having a comment about why we don't ignore some syntax that's labeled as such?

typed-racket-lib/typed-racket/types/abbrev.rkt
```diff
@@ -27,6 +27,7 @@
 
 ;; Convenient constructors
 (define -App make-App)
+(define -Self (make-F 'self-))
 (define -mpair make-MPair)
```
pnwamk: Are we sure nothing bad is going to happen by hand-inserting in the symbol `'self-` here? I can't think of a specific concern, but it just seems a little funny to have that "magic symbol" there.

typed-racket-lib/typed-racket/types/base-abbrev.rkt
```diff
@@ -345,4 +345,3 @@
     [(_ (var) consts ty)
```
pnwamk: See comment for `.../global-env.rkt.`

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
capfredf: This is a bug. I will fix it soon

typed-racket-lib/typed-racket/static-contracts/combinators/structural.rkt
AlexKnauth:
```
> (flat-contract? (struct-type-property/c exact-nonnegative-integer?))
#f
> (chaperone-contract? (struct-type-property/c exact-nonnegative-integer?))
#f
> (impersonator-contract? (struct-type-property/c exact-nonnegative-integer?))
#t
```

capfredf:
> LGTM overall, modulo minor comments.
> 
> Would a few more tests be valuable? E.g.,I thought I saw some changes to contract generation code but no tests to exercise those changes.

Good catch. I will add more tests for contract generation.

typed-racket-lib/typed-racket/types/abbrev.rkt
capfredf: @pnwamk, I can't come up with a better solution to representing `self` as a hygienic singleton type var. 
Also, it looks like I have to put it elsewhere since it is not a constructor.

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -2289,7 +2289,7 @@
 [special-comment-value (-> -Special-Comment Univ)]
 
 ;; Section 13.8
-[prop:custom-write -Struct-Type-Property]
+[prop:custom-write (-struct-property (-> -Self -Output-Port B ManyUniv))]
```
AlexKnauth: The third argument in this function type should not be `B` / `Boolean`. It should allow a boolean, `0`, or `1`.

The [documentation for `prop:custom-write`](https://docs.racket-lang.org/reference/Printer_Extension.html#(def._((quote._~23~25kernel)._prop~3acustom-write))) through the [documentation for `gen:custom-write`](https://docs.racket-lang.org/reference/Printer_Extension.html#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._gen~3acustom-write%29%29) specifies that it

> takes three arguments: the structure to be printed, the target port, and an argument that is `#t` for write mode, `#f` for display mode, or `0` or `1` indicating the current quoting depth for print mode.

This seems to specify the type `(U #t #f 0 1)` equivalently `(U Boolean 0 1)` for that third argument.

AlexKnauth: You can construct the `(U #t #f 0 1)` type with `(Un -True -False -Zero -One)` for the "base-env" declaration.

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -2289,7 +2289,7 @@
...
 [custom-write? (-> Univ B)]
 [custom-write-accessor (-> Univ (-> Univ -Output-Port B ManyUniv))]
```
AlexKnauth: This would have to change as well, to use `(Un -True -False -Zero -One)` instead of `B` for the third argument of the result.

typed-racket-test/fail/bad-struct-prop.rkt
```diff
@@ -0,0 +1,6 @@
+#;
+(exn-pred " Type Checker: type mismatch")
+#lang typed/racket
+(struct foo ([x : Number]) #:property prop:custom-write
+  (lambda ([n : Number] [p : Output-Port] [b : Boolean]) : Void
```
AlexKnauth: This type annotation is wrong. The third argument `b` (the "mode" argument) should have the type `(U Boolean 0 1)`, not `Boolean`.

typed-racket-test/fail/bad-struct-prop1.rkt
```diff
@@ -0,0 +1,8 @@
+#;
+(exn-pred " Type Checker: type mismatch")
+#lang typed/racket
+(struct foo ([x : Number]))
+
+(struct dummy ([y : Number]) #:property prop:custom-write
+  (lambda ([self : dummy] [p : Output-Port] [b : Boolean]) : Void
```
AlexKnauth: Same here, the "mode" argument should have type `(U Boolean 0 1)`.

typed-racket-test/succeed/struct-props.rkt
```diff
@@ -0,0 +1,38 @@
+#lang typed/racket
+(struct dummy ([x : Number]))
+
+(struct foo-dummy dummy ([x : Number]) #:property prop:custom-write
+  (lambda ([self : dummy] [p : Output-Port] [b : Boolean]) : Void
```
AlexKnauth: Same here, the "mode" argument should have type `(U Boolean 0 1)`.

typed-racket-test/succeed/struct-props.rkt
```diff
@@ -0,0 +1,38 @@
...
+          (displayln (+ (dummy-x self) 20))))
+
+(struct foo ([x : Number]) #:property prop:custom-write
+  (lambda ([self : foo] [p : Output-Port] [b : Boolean]) : Void
```
AlexKnauth: Same here, the "mode" argument should have type `(U Boolean 0 1)`.

typed-racket-test/succeed/struct-props.rkt
```diff
@@ -0,0 +1,38 @@
...
+          (displayln (+ (foo-x self) 20))))
+
+(struct foobar foo ([y : Number]) #:property prop:custom-write
+  (lambda ([self : foo] [p : Output-Port] [b : Boolean]) : Void
```
AlexKnauth: Same here, the "mode" argument should have type `(U Boolean 0 1)`.

typed-racket-test/succeed/struct-props.rkt
```diff
@@ -0,0 +1,38 @@
...
+          (displayln (+ (foo-x self) 20) p)))
+
+(struct foobar^ foo ([y : Number]) #:property prop:custom-write
+  (lambda ([self : foobar^] [p : Output-Port] [b : Boolean]) : Void
```
AlexKnauth: Same here, the "mode" argument should have type `(U Boolean 0 1)`.

typed-racket-test/succeed/struct-props.rkt
```diff
@@ -0,0 +1,38 @@
...
+          (displayln (+ (foobar^-y self) 20) p)))
+
+(struct foobarabc foobar ([z : Number]) #:property prop:custom-write
+  (lambda ([self : foo] [p : Output-Port] [b : Boolean]) : Void
```
AlexKnauth: Same here, the "mode" argument should have type `(U Boolean 0 1)`.

typed-racket-test/succeed/struct-props.rkt
```diff
@@ -0,0 +1,38 @@
...
+          (displayln (+ (foo-x self) 20) p)))
+
+(struct foobarabc^ foobar ([z : Number]) #:property prop:custom-write
+  (lambda ([self : foobar] [p : Output-Port] [b : Boolean]) : Void
```
AlexKnauth: Same here, the "mode" argument should have type `(U Boolean 0 1)`.

typed-racket-test/succeed/struct-props.rkt
```diff
@@ -0,0 +1,38 @@
...
+          (displayln (+ (foobar-y self) 20) p)))
+
+
+(struct (A B) poly-foo ([x : A] [y : B])
+  #:property prop:custom-print-quotable 'never
+  #:property prop:custom-write
+  (lambda ([self : poly-foo] [p : Output-Port] [b : Boolean]) : Void
```
AlexKnauth: Same here, the "mode" argument should have type `(U Boolean 0 1)`.

typed-racket-lib/typed-racket/base-env/base-env.rkt
AlexKnauth: The third argument in this function type should not be `B`. It be should be `(Un -True -False -Zero -One)` to allow `#t`, `#f`, `0`, or `1`.

The [documentation for `prop:custom-write`](https://docs.racket-lang.org/reference/Printer_Extension.html#(def._((quote._~23~25kernel)._prop~3acustom-write))) through the [documentation for `gen:custom-write`](https://docs.racket-lang.org/reference/Printer_Extension.html#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._gen~3acustom-write%29%29) specifies that it

> takes three arguments: the structure to be printed, the target port, and an argument that is `#t` for write mode, `#f` for display mode, or `0` or `1` indicating the current quoting depth for print mode.

This seems to specify the type `(U #t #f 0 1)` equivalently `(U Boolean 0 1)` for that third argument, which is represented in the "base-env" as `(Un -True -False -Zero -One)`.

typed-racket-lib/typed-racket/base-env/base-env.rkt
AlexKnauth: This would have to change as well, to use `(Un -True -False -Zero -One)` instead of `B` for the third argument of the result.

capfredf:
> You misunderstood the type of `prop:custom-write` because the Racket documentation for `custom-write-accessor` was out of date. Both `gen:custom-write` and `prop:custom-write` use `#t` to indicate write mode, `#f` to indicate display mode, and `0` or `1` to indicate print mode, with `0` for printing outside of a quoted expression, and `1` for printing inside of a quoted expression.
> 
> I've made a pull request to fix the documentation for `custom-write-accessor`: [racket/racket#2633](https://github.com/racket/racket/pull/2633)

Thank you. I will fix the type annotation for custom-write.

typed-racket-lib/typed-racket/base-env/base-env.rkt
AlexKnauth: The type for `custom-write-accessor` is still wrong/out-of-date. It should be
```racket
(-> Univ (-> Univ -Output-Port (Un B (-val 1) (-val 0)) ManyUniv))
```
with the `(Un B (-val 1) (-val 0))` instead of `B` in the mode argument. The updated function type should be a subtype of the previous one, since the unioned mode argument is a supertype in a contravariant position.

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -61,7 +61,7 @@
 [equal?/recur (-> Univ Univ (-> Univ Univ Univ) B)]
 [immutable? (asym-pred Univ B (-PS (-is-type 0 (Un -Bytes -BoxTop -String (-Immutable-HT Univ Univ) (-ivec Univ)))
                                    (-not-type 0 (Un (-Immutable-HT Univ Univ) (-ivec Univ)))))]
-[prop:equal+hash -Struct-Type-Property]
+[prop:equal+hash (-struct-property Univ)]
```
AlexKnauth: Should this `prop:equal+hash` type be something like
```
(-lst* (-> -Self -Self (-> Univ Univ B) Univ)
       (-> -Self (-> Univ -Int) -Int)
       (-> -Self (-> Univ -Int) -Int))
```
instead of `Univ` inside the `-struct-property`?

capfredf: Yes. I fixed it in my local branch but I haven't push it yet.

AlexKnauth:
Should there be a `fail` test to show that `-Imp` doesn't make assumptions about type-arguments being the same when the struct is polymorphic?

Such as in this case:
```racket
#lang typed/racket

(struct (A) wrap ([a : A]) #:transparent
  #:property prop:equal+hash
  (list
   ; This should fail because it is not guaranteed that the values inside `self` and `other`
   ; will be of the same type `A`. They must be allowed to be different types `A` and `B`.
   (lambda ([self : (wrap A)] [other : (wrap A)] [rec : (-> Any Any Boolean)])
     (rec (wrap-a self) (wrap-a other)))
   (lambda ([self : (wrap A)] [rec : (-> Any Integer)])
     (rec (wrap-a self)))
   (lambda ([self : (wrap A)] [rec : (-> Any Integer)])
     (rec (wrap-a self)))))
```

PR #806, opened 2019-02-24
add some disarms

mflatt:
Two extra calls to `disarm*` allow TR to keep working if some libraries are changed to include `syntax-protect`s (that seem needed).


samth:
The test failures look like they're about the number parsing changes and not this, but I'd prefer to know that the test passed for someone.

mflatt:
I think I've fixed those failures, but when I combine with the `syntax-protect` additions, I still see failures related to those additions. So, I'll try again.

mflatt:
The revised patch (along with Racket repairs) now passes the tests for me.

samth:
Sounds good.

PR #811, opened 2019-03-20
[WIP] attempt to fix types for `raise-argument-error`

abmclin:
This is connected to https://github.com/racket/racket/pull/2544

Currently this fix does not work but here you go.

Based on various examples, I extended `->key` in `base-abbrev.rkt` to
support rst pattern variable since the new `raise-argument-error`
defintion has a rest argument.

I then move the old `raise-argument-error` type to `base-env.rkt` and
replace the type abstraction with the extended `->key` abstraction.

My interpretation of the `->key` pattern is that since `#:more-info` is
an optional keyword it needs to have the following signature
`#:more-info (Un -String (-val #f)) #t` since it has default argument of
`#f` if no strings are provided. I think `#t` is used to indicate that
`#:more-info` is optional but I could be entirely wrong.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env-indexing-abs.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env-indexing-abs.rkt
@@ -328,10 +328,6 @@
    [real->decimal-string (-Real [index-type] . ->opt .  -String)]
 
 
-   [raise-argument-error
-    (cl->*
-     [-> Sym -String Univ (Un)]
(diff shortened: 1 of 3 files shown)


samth:
I've converted this to a draft PR since you said it doesn't work yet.

abmclin:
Yes thank you, the other PR on racket is still in progress. I will update when that PR is ready to be merged.

abmclin:
Unfortunately, due to lack of sufficient time, I am closing this PR.

PR #817, opened 2019-04-04
typed/web-server/http: update types (typed-racket-more v1.10)

LiberalArtist:
- Weaken the types of `response-output` and
  the `#:output` argument to `response/output`
  from `Void` to `Any`, corresponding to
  <https://github.com/racket/web-server/commit/a93685>
  and v1.2 of the `web-server-lib`.

  The return type should really be `AnyValues`
  to match the contract `any`,
  but Typed Racket can't generate a contract for `AnyValues`.

- Update the types for cookie-related functions
  to reflect the re-implementation based on `net/cookies/server`
  from <https://github.com/racket/web-server/commit/7974d4d>.
  This commit adds only as much of `typed/net/cookies/server`
  and `typed/net/cookies/common` as needed for `typed/web-server/http`.

- Document that `net/cookie` is deprecated.

- Update the types for `valid-id-cookie?` and `request-id-cookie`
  to support the `#:shelf-life` argument added in
  <https://github.com/racket/web-server/commit/626ad02>.

- Add the value `temporarily/same-method` with
  the type `Redirection-Status`, corresponding
  to <https://github.com/racket/web-server/commit/4601f4>.

- Correct the types of `response-seconds` and the
  corresponding arguments to `response/full`,
  `response/output`, and `response/xexpr`
  to use `Real`.
  While this corresponds to
  <https://github.com/racket/web-server/pull/61>,
  using response values containing complex numbers
  would cause runtime errors even before this:
  the `response-seconds` is passed to `seconds->date`.

### DIFF
--- a/typed-racket-doc/info.rkt
+++ b/typed-racket-doc/info.rkt
@@ -3,6 +3,7 @@
 (define collection 'multi)
 
 (define build-deps '("net-doc"
+                     "net-cookies-doc"
                      "scheme-lib"
                      "srfi-lite-lib"
                      "r6rs-doc"
@@ -14,7 +15,7 @@
                      "pict-lib"
                      ("typed-racket-lib" #:version "1.10")
                      "typed-racket-compatibility"
-                     "typed-racket-more"
+                     ("typed-racket-more" #:version "1.10")
                      "racket-doc"
                      "draw-lib"))
 (define deps '("base"))
--- a/typed-racket-more/typed/web-server/http.rkt
+++ b/typed-racket-more/typed/web-server/http.rkt
@@ -1,6 +1,7 @@
 #lang typed/racket
 
-(require typed/net/url)
+(require typed/net/url
+         typed/racket/date)
 
 (provide (all-defined-out))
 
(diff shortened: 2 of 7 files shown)


LiberalArtist:
**Note:** This should build against any version of the `web-server-lib` package since <https://github.com/racket/web-server/commit/4601f4>, but I didn't increment the package version number when I made the cookie-related changes or added `temporarily/same-method`. In <https://github.com/racket/web-server/pull/61> I finally increment the version to 1.3, so that is the earliest `#:version` the `typed-racket-more` package can test for to be sure of getting all of the relevant changes.

jeapostrophe:
LGTM

typed-racket-more/typed/web-server/http.rkt
```diff
@@ -57,65 +58,103 @@
...
+                            [#:seconds Real]
                             [#:mime-type (Option Bytes)]
                             [#:headers (Listof Header)]
                             Response)]
                        [TEXT/HTML-MIME-TYPE Bytes])
 
+;; For backwards compatability:
+;; these bindings are not exported by web-server/http,
+;; but they were exported by previous versions of
+;; web-server/http based on net/cookie.
```
samth: Not relevant to this PR, but was this a backwards-incompatible change to `web-server/http` in the past?

samth:
This build failure seems like a problem, though: https://travis-ci.org/racket/typed-racket/jobs/515528872#L497

typed-racket-more/typed/web-server/http.rkt
LiberalArtist: If you mean the exported names: no. Line 86 should have said `;; typed/web-server/http based on net/cookie.` (I will submit a correction shortly.) The names `cookie?`, `cookie-name?`, `cookie-value?`, and `valid-domain?` are not and never were provided by untyped `web-server/http`: they were only provided by `typed/web-server/http`.

If you mean about the underlying change: this was discussed at <https://github.com/racket/web-server/issues/20>. IIRC the conclusion was basically that `web-server/http` had tried very hard to avoid making promises about its internal representation of cookies (e.g. it did not provide `cookie?`). So we took pains to keep the `web-server/http` API compatible (e.g. parsing dates provided as strings …), but ultimately wanted to track the underlying change in the web.

typed-racket-doc/info.rkt
```diff
@@ -14,7 +15,7 @@
                      "pict-lib"
                      ("typed-racket-lib" #:version "1.10")
                      "typed-racket-compatibility"
-                     "typed-racket-more"
+                     ("typed-racket-more" #:version 1.10)
```
LiberalArtist: I think this is the issue with the build (the version should be a string, so `("typed-racket-more" #:version "1.10"). I'll push a patch shortly.

typed-racket-more/typed/web-server/http.rkt
LiberalArtist: The fact that `typed/web-server/http` previously provided a _different_ opaque `Cookie` type than `typed/net/cookie` (the deprecated one) seems to be consistent with that rationale.

LiberalArtist: @samth One difference I just thought of is that previously, the predicates could be used for occurrence typing (e.g. `cookie-name?` for `Cookie-Name`). The actual representation hasn't changed (they were and are still `String`s, with slightly different well-formedness checks), but that might make some programs now fail to typecheck. I can add little wrappers to attach a `String` proposition.

LiberalArtist:
@samth I've fixed the build failure you linked to. This new failure should just be because https://github.com/racket/web-server/pull/61 hasn't gone into the snapshot yet: https://travis-ci.org/racket/typed-racket/jobs/515881390#L630-L635

LiberalArtist:
The build now seems to be passing: https://travis-ci.org/racket/typed-racket/builds/515881389

PR #820, opened 2019-05-22
avoid separate `hash-keys` and `hash-values`

mflatt:
The list returned by `hash-keys` is not guaranteed to be parallel to the list returned by a separate call to `hash-values`.

Different orders in Racket Cs cause TR's contract generation to mix up bindings, as illustrated by this "Insert Large Letters" failure:

http://drdr.racket-lang.org/52168/cs/racket/share/pkgs/drracket-test/tests/drracket/insert-large-letters-test.rkt

The contracts shown in the error message will vary, but they're almost always the wrong class contract.

### DIFF
--- a/typed-racket-lib/typed-racket/static-contracts/instantiate.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/instantiate.rkt
@@ -121,11 +121,14 @@
            (hash-set! memo-table sc result)
            result]))
   (define constraints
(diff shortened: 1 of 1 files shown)


samth:
Looks good to me.

mflatt:
I see that there was a test failure flagged by Travis, but the failure appears to pre-date this change, so I pushed the commit.

PR #824, opened 2019-05-26
any-wrap: avoid looping on cyclic values

### LINKED ISSUE #823: value serialization loops forever in TR
## Summary

It appears to me that certain cyclic Typed Racket values cause check-equal?'s error message to loop infinitely. I think that this is probably actually a racket bug, so I haven't tried to minimize the bug at all. If one of y'all can confirm that a) there aren't any known isssues around infinite-time printing of cyclic values, and b) this is probably a Racket problem rather than a TR problem, I'll submit this as a Racket bug instead.

7.3

### What program did you run?

[lab7.txt](https://github.com/racket/typed-racket/files/3218207/lab7.txt)

The test case failure should have been displayed in finite time and memory.

bennn:
When wrapping a value,
- keep a seteq of already-seen values
- return if we see the same value twice

~Fixes #823~

### DIFF
--- a/typed-racket-lib/typed-racket/utils/any-wrap.rkt
+++ b/typed-racket-lib/typed-racket/utils/any-wrap.rkt
@@ -77,7 +77,7 @@
      v 
      "Attempted to use a higher-order value passed as `Any` in untyped code: ~v" v))
   
-  (define (wrap-struct neg-party s [inspector (current-inspector)])
+  (define (wrap-struct neg-party s seen [inspector (current-inspector)])
     (define blame+neg-party (cons b neg-party))
     (define (extract-functions struct-type)
       (define-values (sym init auto ref set! imms par skip?)
@@ -93,7 +93,7 @@
                (list* (make-struct-field-accessor ref n)
                       (lambda (s v) (with-contract-continuation-mark
                                      blame+neg-party
-                                     (any-wrap/traverse v neg-party)))
+                                     (any-wrap/traverse v neg-party seen)))
                       res)
                (cdr imms))
               ;; field is mutable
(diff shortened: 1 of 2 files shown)


rfindler:
Don't you have to return the wrapped/copied value when you hit the same input the second time?

Also, I think that you won't preserve the sharing in the same form as the input value unless you use a mutable thing and update it.

bennn:
Robby is right. The PR as-is changes `#0=(1 . #0#)` into `(1 . #0=(1 . #0#))` and doesn't protect the `cdr`.

I'm planning to:
1. do the sharing correctly for mutable data & things that `make-reader-graph` works for
2. raise an exception for other data (immutable set, immutable struct field, promise, evt) that makes a cycle
3. see how this affects performance ... if it's bad then maybe we can spend one pass to detect cycles?

rfindler:
Doing something where you optimistically don't do much but then bail over and start over when you need to sounds like the right approach to me. There are similar concerns in the implementation of `equal?` and I recall Ken had a nice paper about it (a more sophisticated version of that, I believe).

samth:
What's the status here?

bennn:
Nothing new --- I've yet to try writing a deep-copy for cyclic values.

samth:
Is the current status an improvement over the status quo?

bennn:
No, the current status quietly gives a wrong answer.

I think the shortest path to an improvement is to error if there's a cyclic value

samth:
Is this now ready to go?

bennn:
This is ready to merge if we're happy raising an exception for cyclic values

(Today I edited the comments in the changed files)

PR #831, opened 2019-06-16
Remove tests that rely on single-floats

stamourv:
These error now that single-floats are disabled by default.


bennn:
Does it make sense to move some of these tests into a context where single-floats are enabled?

stamourv:
The one thing I can think of is to move these tests to files that aren't run by default, like say in the xfail directory.

stamourv:
My general feeling on this, though, is that since single-floats are going away, I don't think there's much value in these tests anymore, so I don't think we're losing much by just getting rid of them.

samth:
I don't think single floats are going away, just that they're not available most of the time in the default reader (or on RacketCS). I agree with Ben that we should keep these tests somewhere, and run them by default with single flonum reading enabled (similar to how the Racket tests work).

rfindler:
If they are kept probably good to make sure they are being run somewhere somehow so that they don't bit-rot

stamourv:
I could see guarding with a dynamic `single-flonum-supported?` check working for, e.g., the typechecking tests, which call the typechecker directly. I can change them to do that.

For integration and (especially) optimizer tests, the only way I can think of doing that would involve creating separate test suites that are run conditionally. That would complicate the harness, which doesn't seem worth it for such a niche feature. I'm open to better ideas, but I'm also not interested in spending much time on this.

(Also, because of a git slip-up, this PR does not include all the tests I removed. I'll force-push to fix that.)

samth:
I think dropping the optimizer tests is reasonable (and maybe dropping optimization of single floats entirely).

stamourv:
Ok, what about the couple integration tests that use single-flonums?
Also, there are no single-flonum optimizations.

samth:
Let's keep those -- single flonum support is there so we need to test it. I'll work on fixing the harness to enable that.

stamourv:
Sounds good. I'll fix the unit tests and random tests, and delete the optimizer tests.
I'll just push those directly when I'm done.

stamourv:
Actually, looks like the unit tests would also require changes to harness code, so I'll leave those to you too.

stamourv:
Pushed the updated version.

stchang:
Should these changes be in the 7.4 release?

I installed 7.4 CS pre-release and `main-distribution-test` and got the following errors:
```
raco setup: error: during making for <pkgs>/typed-racket-test (Typed Racket Test Suite)
raco setup:   real->single-flonum: unsupported
raco setup:     compiling: <pkgs>/typed-racket-test/historical-counterexamples.rkt
raco setup: error: during making for <pkgs>/typed-racket-test (Typed Racket Test Suite)
raco setup:   real->single-flonum: unsupported
raco setup:     compiling: <pkgs>/typed-racket-test/unit-tests/typecheck-tests.rkt
raco setup: error: during making for <pkgs>/typed-racket-test/unit-tests
raco setup:   real->single-flonum: unsupported
raco setup:     compiling: <pkgs>/typed-racket-test/unit-tests/subtype-tests.rkt
raco setup: error: during making for <pkgs>/typed-racket-test/unit-tests
raco setup:   real->single-flonum: unsupported
raco setup:     compiling: <pkgs>/typed-racket-test/unit-tests/typed-units-tests.rkt
```

Not sure where the problem lies but though I would report it (did a quick search and didnt see any other mentions of this problem for pre-release builds).

PR #832, opened 2019-06-30
Fixnum

mflatt:
The implementation of `portable-fixnum?` is not correct for Racket CS, which has one less bit for fixnums on 32-bit platforms. I noticed because the implementation of `case` needs a similar function to replace its use of `fixnum?`. The `racket/fixnum` library now provides `fixnum-for-every-platform?`, so this patch uses and supports that function.


samth:
This looks good to me, but is the `portable-index?` function below still right?

mflatt:
I didn't think too hard about `portable-index?`, but I assumed that 28 bits was meant to be fairly conservative, and it is covered by the range of possible indices.

PR #833, opened 2019-06-30
RFC for initial implementation of types for untyped contracts

btlachance:
I'd appreciate any feedback you all have on the RFC.

(The section "Reference-level explanation" is probably not detailed enough, but I'm having trouble figuring out what more needs to be said or what should be excerpted from another source.)


samth:
This is great. I hope that @capfredf @pnwamk @bennn can offer thoughts as well.

samth:
A few initial thoughts:

1. Can you given an example with a function contract and say what the result type is?
2. Can you say more about the status of dependent contracts?
3. Can you say how the `FlatContract` type is handled?

btlachance:
(Should these go in the RFC?)

*1. Can you given an example with a function contract and say what the result type is?*

Sure, the contract `(->/c positive-real? positive-real?)` has type 

    (Contract (-> Positive-Real Real)
              (-> Real Positive-Real))

Another: `(->/c (>/c 9) (between/c 0 99))` has type

    (Contract (-> Real Any)
              (-> Any Real))

The type I'm using for `>/c` is `(-> Real Real (FlatContract Any Real))`

*2. Can you say more about the status of dependent contracts?*

The only dependent contract I have is `->i`, and as far as I can tell the current implementation supports:
- mandatory and optional domain contracts, both the named and keyword variants
- `#:rest` contracts
- all variants of the pre- and post-condition contracts
- *named* single- and multi-value range contracts (i.e. the `id+ctc` variants in the contract docs) 

Big things missing:
- The `any` range contract
- The`the-unsupplied-arg` and `unsupplied-arg?` values. Giving types to those shouldn't be hard, but it does make me question how unions and contract types should interact (needed to support conditionals that produce contracts, like in the `->i` example that uses `unsupplied-arg?`). 

And type-system wise: dependent function contracts just have a contract type with ordinary function types in the input and output positions.

*3. Can you say how the `FlatContract` type is handled?*

Single-argument function types are a subtype of `FlatContract` (which makes them a subtype of `Contract`). And a couple contract combinators (e.g. `not/c`) only allow `FlatContract` arguments. The domain of the predicate is the contract's input type, and the contract's output type is the predicate's proposition or the predicate's domain type if the proposition was trivial.

In the current implementation the function application code (`tc-funapp.rkt`) allows a `FlatContract` in operator position. This maybe isn't the right design choice since `flat-contract?` is true for things that aren't usable as functions; If more of the atomic types should be subtypes of `FlatContract`, then the funapp code should change.

samth:
Adding the first two answers to the RFC would be good.

For the `FlatContract` type, I'm somewhat worried. Could we represent `FlatContract` as just a union of various types? And also add a specific type for the result of `coerce-contract` when it produces a flat contract, so that those can be applied?

rfcs/text/0000-basic-typed-contracts.md
```diff
@@ -0,0 +1,249 @@
...
+applied to, ensuring that a caller provides arguments that satisfy the
+domain contract and that the function returns a result that satisfies
+the range contract; Values flow from the caller through the domain
+contract into the function, and out of the function and through the
+range contract.
+
+This value flow is reflected in the type of a function contract. For
+our `(->/c positive? positive?)` the type is
+
+    (Contract (-> {2} Positive-Real Real {3})
```
bennn: This notation is a little confusing. How about `(Contract (-> #|2|#Positive-Real #|3|#Real) ....)` so that:
1. it's valid TR syntax
2. the labels always come first
3. the labels touch the names they belong to

rfcs/text/0000-basic-typed-contracts.md
```diff
@@ -0,0 +1,249 @@
...
+
+In the input type of the contract, the range of the type is `Real`
+(marked `{3}`) because applying the contract to a function is only
+allowed if that function returns a value that can flow through the
+range contract (whose input type is `Real`). Finally, in the output
+type of the contract the range type is `Positive-Real` (marked `{4}`)
+because that's the output type of the range contract. This means that
+if we apply this contract to a function, then the resuling function
+returns at least a `Positive-Real` if it returns at all.
+
```
bennn: Add quick examples for lists, structs, and/or vectors?

I'm guessing: functions are a great example because a function type has input & output. Other contract-types should be simpler, but it'd be nice to have that confirmed.

rfcs/text/0000-basic-typed-contracts.md
```diff
@@ -0,0 +1,249 @@
...
+# Reference-level explanation
+
+## Interaction with other features
+
+As far as I can tell contracts are mostly self-contained. Some
+exceptions:
+
+  - Predicates are subtypes of the `FlatContract` type.
```
bennn: Does `exact-integer?` have a predicate type and does that mean `(Contract Any Integer)` is a subtype of `FlatContract`?

rfcs/text/0000-basic-typed-contracts.md
```diff
@@ -0,0 +1,249 @@
...
+    contract's output type with the range type of the function that
+    will be monitored. In the code this is called "pairwise
+    intersection" and in the thesis this is the `comb` operator.
+
+As far as subtyping goes, `(Contract S T)` and `(FlatContract S T)` is
+contravariant in `S` and covariant in `T`.
+
+## How the feature would be implemented
+
+The MS thesis linked up top gives type rules formalizing what I walked
```
bennn: Is there anything to implement that goes beyond what's "standard" for adding a new type?

If not, then I guess this section can say: we need to add the new types, and that means a syntax, printer, typechecking, subtyping, and ~type->contracts~ [oh! guess not]. (Without going into further detail)

rfcs/text/0000-basic-typed-contracts.md
```diff
@@ -0,0 +1,249 @@
...
+contract for typed/untyped interaction, typed code can't let untyped
+code use a value of contract type. At the moment I believe the current
+implementation raises an error (maybe an internal one?) when providing
+a value of contract type because I haven't defined a way to compile
+those types to contracts.
+
+I haven't properly looked at how these contract types interact with
+polymorphic types. Currently the only way to get primitive contracts
+with such a type are a predicates on values of some polymorphic type,
+e.g. a function like `null?`.
```
bennn: For now lets say both things aren't supported, and change the PR to give useful error messages?

btlachance:
> Could we represent `FlatContract` as just a union of various types?

This is a great idea! I've only thought about this a little but I think using a union type for `FlatContract` is doable.

> And also add a specific type for the result of `coerce-contract` when it produces a flat contract, so that those can be applied?

Is your thought that the "specific type" for the result should be a subtype of the `FlatContract` (union) type? Or does the specific type use some new type constructor?

btlachance:
(I'll add these to the RFC too. But replying here first.)

> Can the RFC explain why there aren't types for chaperone & impersonator contracts?

AFAIK there's no reason to *not* have them, it just wasn't part of the initial work. It was something we talked about eventually wanting though. After adding them, I think the `Contract` type could then just be defined as a union of flat, chaperone, and impersonator types.

> I like the 2-arg contract type. But why not use TR's prop syntax and have `(Contract Any #:+ Integer)`? Because it's a chore to write `#:+` everywhere?

No complaints here about the proposition syntax, but I haven't used it often enough in TR to have formed much of an opinion; We just never thought of it. But I like this idea a lot since the output type seems analagous to a proposition.

Since there is a default proposition for a function type, using the proposition syntax for contract types suggests I might need to figure out a sensible default there, too.

> Allowing `FlatContract` as an operator sounds good. Maybe the error message can suggest adding `contract-first-order-passes?` for the edge cases.

I'm not sure how `contract-first-order-passes?` fits here---did you mean `contract-first-order`? I had forgotten about that operator, so at the very least I should get my head around how `contract-first-order` and `coerce-contract` differ at first-order types.

rfcs/text/0000-basic-typed-contracts.md
btlachance: The type of `exact-integer?` should be something like `(-> Any Boolean : Integer)`, which would mean it also has type `(FlatContract Any Integer)`.

And `FlatContract` is only a subtype of `Contract`, not a supertype of it.

rfcs/text/0000-basic-typed-contracts.md
btlachance: Good idea. Will update the RFC

rfcs/text/0000-basic-typed-contracts.md
btlachance: I'll add a list example.

As far as structs go: No support yet for `struct/c`, and I might need to provide a variant of `contract-out` that does some additional checks; At the moment the PR just reprovides Racket's `contract-out`. I updated the RFC with notes about that.

For vectors, are you interested in mutable vectors?

rfcs/text/0000-basic-typed-contracts.md
btlachance: AFAIK defining the types was pretty standard. RFC updated.

rfcs/text/0000-basic-typed-contracts.md
btlachance: Updated the RFC!

rfcs/text/0000-basic-typed-contracts.md
bennn: Mutable specifically

bennn:
> (btlachance) Since there is a default proposition for a function type, using the proposition syntax for contract types suggests I might need to figure out a sensible default there, too.

I wouldn't worry about the default for now, and just require `#:+` everywhere.

> > (bennn) Allowing FlatContract as an operator sounds good. Maybe the error message can suggest adding contract-first-order-passes? for the edge cases.

> (btlachance) I'm not sure how contract-first-order-passes? fits here---did you mean contract-first-order? I had forgotten about that operator, so at the very least I should get my head around how contract-first-order and coerce-contract differ at first-order types.

Either one:

```
> (contract-first-order-passes? 4 4)
#t
> ((contract-first-order 4) 4)
#t
```

PR #834, opened 2019-07-02
typed/web-server/http: response message inference (typed-racket-more v1.11)

LiberalArtist:
This commit weakens the type of the message argument
of `response/full`, `response/output`, and `response/xexpr`
from `Bytes` to `(Option Bytes)`.
A `#false` argument causes the message to be inferred from the
numeric response code: note that the type of `response-message`
*does not* change.

Corresponds to v1.4 of the `web-server-lib` package,
https://github.com/racket/web-server/commit/9262724, and
https://github.com/racket/web-server/pull/62.

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/libraries.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/libraries.scrbl
@@ -251,7 +251,11 @@ and the @racket[URL] and @racket[Path/Param] types from
 
 @history[#:changed "1.10"
          @elem{Updated to reflect @racketmodname[web-server/http]
-          version 1.3.}]
+          version 1.3.}
+         #:changed "1.11"
+         @elem{Updated to reflect @racketmodname[web-server/http]
+          version 1.4.}
+         ]
 
 @defmodule/incl[typed/db]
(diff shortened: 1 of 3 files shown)


samth:
Looks good to me.

LiberalArtist:
Aside from this particular update to the code, I wonder if there's something more we could do to remind people to update `typed/web-server/http` (and I guess `typed/web-server/configuration/responders`, also) when they update the corresponding untyped `web-server` module. I didn't even realize `typed/web-server/http` existed until just before I submitted https://github.com/racket/typed-racket/pull/817, which is why so many changes from my previous PRs were bundled into that commit.

The best idea I've had so far is a comment at the top of the relevant Scribble file in `web-server-doc`: I think that's the most central file one would have to edit to make a change to the exported types.

samth:
I like that idea.

LiberalArtist:
It doesn't look like any of the CI failures are related to this PR: most seem to be about the changes to fixnums and single-flonums. @samth, should this go in 7.4?

samth:
If it would help you to have this in 7.4, I can ask the release managers about it.

PR #841, opened 2019-07-09
Testing on RacketCS.

samth:


samth:
This currently has errors due to racket/racket#2732 racket/racket#2729 and racket/racket#2730.

PR #842, opened 2019-07-09
Improve typechecking for for/last and for*/last

AlexKnauth:
This fixes a problem raised about `for/last` and `for*/last` in this thread on the Racket Users List:
https://groups.google.com/d/topic/racket-users/0tOGWZ9O57c/discussion

It uses the existing internal `define-for/acc:-variants` macro to transform
```
(for/last : type (clause ...)
  body)
```
into
```
(for/fold ([acc : type #f])
          (clause ...)
  body)
```
which type-checks in more situations because it assigns a type for `acc` other than `False`.


samth:
Can you add some test cases?

AlexKnauth:
Just added a few test cases in [`typed-racket-test/succeed/for-last.rkt`](https://github.com/racket/typed-racket/pull/842/files#diff-ccd1c15198e095eda727edf7e8d4d7fb)

AlexKnauth:
I have added that to the tests now as well.

AlexKnauth:
I have also now adjusted the documentation, moving `for/last` and `for*/last` into the the same section as the other supported `for` forms.

Issue #798, opened 2019-02-07
expt type has a slightly incorrect case


pnwamk:
### What version of Racket are you using?
7.2
### What program did you run?
```racket
#lang typed/racket

(ann (expt -0.0 -3.0) Nonnegative-Flonum) ;; => -inf.0
```
### What should have happened?
The type following type is assigned to `expt`:

```
(-> Nonnegative-Flonum Flonum Nonnegative-Flonum)
```

But, as the above program indicates, this is not true.

The solution _might_ be to simply subtract `-0.0` from the first argument type for this particular case... I'll come back to this.

pnwamk:
Similar issue, different incorrect case:

```racket
(ann (expt -0.0 -1) (U Nonnegative-Flonum One)) ;; => -inf.0
```

PR #844, opened 2019-07-12
Changes proposed to merge for 7.4

samth:
These changes do several things.

1. 68c43db addresses the problem that improvements to Racket's complex division of floating point numbers made it out of sync with Typed Racket's optimizer. This repairs the optimizer. Tests are in `historical-counterexamples.rkt`, and now all pass on traditional Racket.
2. a3699a2 27b81c1 417d279 together fix a bug in the type of `expt` on negative floating point numbers, which could result in unsound behavior from Typed Racket (see #798). The latter two commits are because I did the first one wrong.
3. ab0faf6 makes typechecking files with literal extflonums work on RacketCS (otherwise there's an internal error).
4. 534e1a2 and 4e78164 make Typed Racket's tests work on RacketCS (with one exception due to a RacketCS bug). There are a lot of lines changed here, mostly mechanical and related to disabling tests for features that RacketCS does not support. The second commit is needed so that RacketCS can be detected in the tests.

Each of these 4 groups could be separately merged to the release or not. I'm proposing to merge them all.


stamourv:
For 1, what was missing from the code you showed yesterday?

For 2, adding this new `NonNegSignFlonum` type makes the (already way too complicated, IMO) numeric type hierarchy yet more complicated. It also breaks the symmetry between positive and negative. Would it have been possible to instead make the operation types more conservative (accept tighter argument types, produce looser return types)?

3 looks good.

4 looks good also, though I'd split off the optimizer-related changes in 534e1a2 to a separate commit; they cause huge diffs, and make it harder to see the other changes to the rest of the TR test suite, IMO.

samth:
@stamourv we talked about this some at RacketCon, but for posterity:

1. I was missing a conditional on the `swap?` boolean.
2. I agree that this is non-symmetric, but the existence of a union of all flonums with a negative sign didn't seem that useful.
3. Yay.
4. I agree but the commit is cherry-picked from master and so I don't want to make them different.

PR #845, opened 2019-08-06
Fix xvfb in travis-ci

gus-massa:
There was some change in Travis-ci, and the test are failing. I tried with the new recommended method to start `xvfb`, but I got an error in the `integration` test. 

    exception-message:  "Gtk initialization failed for display \":99.0\"" 

(See [travis-ci/gus-massa/typed-racket/#L766](https://travis-ci.org/gus-massa/typed-racket/jobs/568025328#L766) )

After some googling ans some random tries, I got this solution, mostly from https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-xvfb-directly , but I'm not sure about the details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,11 @@ matrix:
   - TESTS=extra VARIANT=cs PATH=~/racket/bin:$PATH
   
 
+services:
+- xvfb
+
 before_install:
-- "export DISPLAY=:99.0"
-- "sh -e /etc/init.d/xvfb start"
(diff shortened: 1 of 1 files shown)


samth:
The only failure is some lingering RacketCS floating point issues, so this looks good.

gus-massa:
It is strange that I also got a "*10 minutes without output*" in the test that is failing. I run the test again and it got fixed so I thought it was a fluke, but it may be an intermittent problem ...

PR #850, opened 2019-08-24
rename contracts in tests

bennn:
Change the names in `check-exn`s to match the new names from cb0fb44


bennn:
The latest build should not have passed; there are errors in https://travis-ci.org/racket/typed-racket/jobs/576236718

Maybe the `allow_failures` needs to change?
https://stackoverflow.com/questions/20905969/travisci-how-to-allow-failures-for-environment-variable

PR #851, opened 2019-08-24
Struct-Property doc

bennn:
@capfredf

### DIFF
--- a/rfcs/text/0002-typechecking-struct-proper-values.md
+++ b/rfcs/text/0002-typechecking-struct-proper-values.md
@@ -55,15 +55,15 @@ declaration. For example, here is a struct `Point` which has the
 
 In addition, some struct properties requires a function to take not only the
 associated struct instance, but also other instance created by the same struct
-constructor. To construct appropriate type for such a function, we use `Self` to
+constructor. To construct appropriate type for such a function, we use `Imp` to
 specify the implementor type. For example, here is a type annotation for
-`prop:equal` struct property:
+`prop:equal+hash` struct property:
 
 ```racket
-(: prop:equal (Struct-Property (-> Self Imp (-> Any Any Boolean) Boolean)))
+(: prop:equal+hash (Struct-Property (-> Self Imp (-> Any Any Boolean) Boolean)))
 ```
 
-This specifies that the a value of a struct's `prop:equal` property should be a
+This specifies that the a value of a struct's `prop:equal+hash` property should be a
 function, whose second argument will be an instance of the same implementator
 type but not the associated struct instance. Note that `Imp` is only a valid
 type within `Struct-Property` declaration. When defining a struct with a struct
@@ -71,12 +71,12 @@ property whose type features `Imp`, we instead use the name of struct currently
 being defined wherever `Imp` appeared in the struct property's type
 declaration. But for polymorphic struct types, we introduce new type variables
 and use them with type constructors for `Imp`. For example, here is a
-polymorphic struct `Point` which has the `prop:equal` struct property
+polymorphic struct `Point` which has the `prop:equal+hash` struct property
 mentioned above:
 
 ```racket
 (struct (T V) Point ([x : T][y : V])
-  #:property prop:equal
+  #:property prop:equal+hash
   (lambda (A B)
     ([self : (Point T V)]
      [other : (Point A B)]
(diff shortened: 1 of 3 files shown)


typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -974,6 +974,31 @@ prefab types with the (implicitly quoted) prefab-key
 
 @history[#:added "1.7"]}
 
+@defform[(Struct-Property ty)]{
+  Internal type that describes a property value that is attached to a
+   structure type.
```
AlexKnauth: This documentation should specify what `ty` means; that values associated with the property must have type `ty`. Would an example like this help?
```racket
;; prop : (Struct-Property ty)
(struct s [] #:property prop expr)
;; requires expr : ty
```

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -974,6 +974,31 @@ prefab types with the (implicitly quoted) prefab-key
...
+
+  @ex[(:print-type prop:input-port)]
+
+  @history[#:added "1.10"]}
+
+@defidform[Self]{
+  Internal type that can only appear in a @racket[Struct-Property] type.
+  Refers to an instance of a structure type that has a value for the property.
```
AlexKnauth: Refers to "an" instance vs refers to "the" instance? I thought `Self` was only meant for "the" exact instance that triggered the property lookup, while `Imp` was for "an" instance of the same structure type.

bennn:
@AlexKnauth I tried again.

I didn't add the example because these types are internal only --- I'm worried people will get confused if they try to run it.

AlexKnauth:
Okay. It looks good to me now

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -974,6 +974,32 @@ prefab types with the (implicitly quoted) prefab-key
 
 @history[#:added "1.7"]}
 
+@defform[(Struct-Property ty)]{
+  Internal type that describes a property that can be attached to a structure
```
samth: This is not an internal type.

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -974,6 +974,32 @@ prefab types with the (implicitly quoted) prefab-key
...
+  the @racket[Self] type refers to this instance.
+
+  @ex[(:print-type prop:custom-write)]
+
+  @history[#:added "1.10"]}
+
+@defidform[Imp]{
+  Internal type that can only appear in a @racket[Struct-Property] type.
+  In contrast to @racket[Self], refers to another instance created by the
+  same struct constructor.
```
samth: Should say something about sub-structs here.

typed-racket-lib/typed-racket/HISTORY.txt
```diff
@@ -1,6 +1,10 @@
+7.4
+- Add Struct-Property, Self, and Imp types.
```
samth: This isn't in 7.4.

rfcs/text/0002-typechecking-struct-proper-values.md
```diff
@@ -55,28 +55,28 @@ declaration. For example, here is a struct `Point` which has the
 
 In addition, some struct properties requires a function to take not only the
 associated struct instance, but also other instance created by the same struct
-constructor. To construct appropriate type for such a function, we use `Self` to
+constructor. To construct appropriate type for such a function, we use `Imp` to
 specify the implementor type. For example, here is a type annotation for
-`prop:equal` struct property:
+`prop:equal+hash` struct property:
```
capfredf: I came up with `prop:equal` because `prop:equal+hash` expects a list of 3 functions, which, I think, would make the example a little more complicated. But it looks like `prop:equal` is confusing.  What about adding a footnote to `prop:equal`,  or changing the name to something like `prop:fake-equal`?

bennn: Ok, I'll go back to `prop:equal` and add a note.

Is there a better example than `prop:equal+hash` to use in the docs for `Imp` ?

capfredf: I haven't found any other built-in properties that rely on `Imp`.

bennn:
I think I've addressed all the comments.

PR #861, opened 2019-09-19
adapt to `*in-directory` rename as `in-directory` from `racket/private/for`

mflatt:
Sync with racket/racket@3907f35d1d.


samth:
LGTM

PR #863, opened 2019-09-30
typecheck make-struct-type-property, properties predicates and accessors

capfredf:
1. add a new type constructor
2. typecheck make-struct-type-property
3. typecheck sp predicates and accessors


typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -1228,14 +1229,15 @@
 [make-struct-field-mutator  (->opt top-func -Nat [(-opt -Symbol)] top-func)]
 
 ;; Section 5.3 (Structure Type Properties)
+#;
```
samth: Why is this commented out?

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -1948,8 +1952,8 @@
 [pipe-content-length (-> (Un -Input-Port -Output-Port) -Nat)]
 
 ;; Section 13.1.8
-[prop:input-port (-struct-property (Un -Input-Port -Nat))]
-[prop:output-port (-struct-property (Un -Output-Port -Nat))]
+[prop:input-port (-struct-property (Un -Input-Port -Nat) #'input-port?)]
+[prop:output-port (-struct-property (Un -Output-Port -Nat) #'out-port?)]
```
samth: `output-port?`

typed-racket-lib/typed-racket/private/type-annotation.rkt
```diff
@@ -87,8 +87,9 @@
      (let ([anns (for/list ([s (in-list stxs)])
                    (type-annotation s #:infer #t))])
        (if (for/and ([a (in-list anns)]) a)
-           (match (tc-expr/check expr (ret anns))
-             [(tc-results: tcrs _) tcrs])
+           (begin
```
samth: Is this needed?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -505,8 +505,11 @@
                [safe-spp/sc (flat/sc #'struct-predicate-procedure?/c)]
                [optimized/sc (if (from-typed? typed-side)
                                  unsafe-spp/sc
-                                 safe-spp/sc)])
-          (or/sc optimized/sc (t->sc/fun t)))]
+                                 safe-spp/sc)]
+               [spt-pred-prodceure?/sc (if (from-typed? typed-side)
```
samth: spelling

typed-racket-lib/typed-racket/typecheck/error-message.rkt
```diff
@@ -58,6 +59,10 @@
      (class-mismatch r1 r2)]
     [((Instance: (app resolve (? Class? c1))) (Instance: (app resolve (? Class? c2))))
      (class-mismatch c1 c2 #t)]
+    [((Has-Struct-Property: prop-name) (Struct: n2 _ _ _ _ _ _))
+     (type-mismatch t1 t2 (format "Struct ~a doesn't has the property ~a"
```
samth: "doesn't have"

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -538,3 +546,16 @@
...
+  (match-define (Struct-Property: ty bpred) struct-property-ty)
+  (cond
+    [(not (unbox bpred))
+     (set-box! bpred pred-id)
+     (define has-spty (make-Has-Struct-Property prop-name))
+     (values->tc-results (make-Values (list (-result struct-property-ty -true-propset -empty-obj)
+                                            (-result (make-pred-ty has-spty))
+                                            (-result (make-Fun (list (-Arrow (list has-spty) ty))))))
+                                             '())]
+    [else (tc-error "the predicate for ~a was set before typechecking the creation of ~a" prop-name)]))
```
samth: Should this be an internal error?

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -203,7 +208,7 @@
 (define (tc-toplevel/pass1.5 form)
   (parameterize ([current-orig-stx form])
     (syntax-parse form
-      #:literals (define-values begin)
+      #:literals (define-values begin #%plain-app)
```
samth: Is this new literal used?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -713,7 +716,14 @@
...
                                "untyped code"))
             (struct-type/sc null))]
-       [(Struct-Property: s) (struct-property/sc (t->sc s))]
+       [(Struct-Property: s _) (struct-property/sc (t->sc s))]
+       [(Has-Struct-Property: orig-id)
+        (match-define (Struct-Property: _ (box pred?)) (lookup-id-type/lexical orig-id))
+        (define orig-name (syntax-property orig-id 'original-name))
+        (flat/sc #`(flat-named-contract '#,(syntax-e pred?) (lambda (x)
+                                                              (unless (struct-type-property-predicate-procedure? #,pred? #,orig-name)
+                                                                (error 'hi (format "~v is not the right property predicate of ~v" #,pred? #,orig-name)))
```
samth: Error message.

typed-racket-lib/typed-racket/private/parse-type.rkt
```diff
@@ -159,6 +159,7 @@
 (define-literal-syntax-class #:for-label Intersection)
 (define-literal-syntax-class #:for-label Refine)
 (define-literal-syntax-class #:for-label Self)
+(define-literal-syntax-class #:for-label Exist)
```
bennn: Is `Exist` a general existential type?

If so, then I think it should accept parenthesis too, like All does `Exist (A) ....` (and someday, accept multiple names)
If not, then rename it to suggest why/how its different?

typed-racket-lib/typed-racket/types/subtype.rkt
```diff
@@ -797,6 +799,12 @@
    (match t2
      [(Evt: result2) (subtype* A result1 result2)]
      [_ (continue<: A t1 t2 obj)])]
+  [(case: Exist (Exist-names: names b1))
+   ;;(eprintf "here am I ..................\n \n \n") 
+   (match t2
+     [(Fun: arrows1)
+      ;;(eprintf "I am here \n \n \n")
+      #f])]
```
bennn: Otherwise return `(void)`?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -713,7 +716,17 @@
...
-       [(Struct-Property: s) (struct-property/sc (t->sc s))]
+       [(Struct-Property: s _) (struct-property/sc (t->sc s))]
+       [(Has-Struct-Property: orig-id)
+        ;; we can't call syntax-local-value/immediate in has-struct-property case in parse-type
+        (define-values (a prop-name) (syntax-local-value/immediate orig-id (λ () (values #t orig-id))))
+        (match-define (Struct-Property: _ (box pred?)) (lookup-id-type/lexical prop-name))
+        ;; if original-name is only set when the type is added via require/typed
+        (define real-prop-var (or (syntax-property prop-name 'original-name) prop-name))
+        (flat/sc #`(flat-named-contract '#,(syntax-e pred?) (lambda (x)
+                                                              (unless (struct-type-property-predicate-procedure? #,pred? #,real-prop-var)
```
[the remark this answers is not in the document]
capfredf: I think so. I think I can move it outside the `flat-name-contract

typed-racket-lib/typed-racket/private/parse-type.rkt
capfredf: Good point. My original intention was to add an existential function type but something like Exist-> didn't look right to me. On the other hand, Exist is not good either since it may imply TR enables standard pack/unpack. Any other thoughts? @bennn @samth

typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt
```diff
@@ -79,6 +79,14 @@
       [(Fun: (list arrow))
        #:when (not (RestDots? (Arrow-rst arrow)))
        (tc/funapp1 f-stx args-stx arrow args-res expected)]
+      [(Exist-names: _ (Fun: (list arrow)))
+       (match-define (list (tc-results: (list (tc-result: t (PropSet: _+ _-) o)) _)) args-res)
+       (let ([checked-ret (tc/funapp1 f-stx args-stx arrow args-res expected)])
+         (match checked-ret
+           [(tc-results: (list (tc-result: t (PropSet: p+ p-) o__)) _)
+            (lexical-env (env+ (lexical-env) (list p+)))]
```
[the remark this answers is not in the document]
capfredf: `add-props-to-current-lexical` doesn't fully work the way as its name implies. It [swaps](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/env/lexical-env.rkt#L40) the old proposition for the argument. But what we want here is to "merge" old and new proposition. e.g., code like [this](https://github.com/racket/typed-racket/pull/863/files/c91ad308f931540e795949ab08a9b58da2d612e0#diff-14a1bd9fe6385f30ff41befea55d849cR20) wouldn't be able to typecheck if `add-props-to-current-lexical!` was used. 

Also, the comment for `add-props-to-current-lexical!` says the function is used at the top level

typed-racket-lib/typed-racket/types/printer.rkt
```diff
@@ -655,11 +657,11 @@
     [(? improper-tuple? t)
      `(List* ,@(map type->sexp (improper-tuple-elems t)))]
     [(Opaque: pred) `(Opaque ,(syntax->datum pred))]
-    [(Struct: nm par (list (fld: t _ _) ...) proc _ _ property-tys)
+    [(Struct: nm par (list (fld: t _ _) ...) proc _ _ properties)
      `#(,(string->symbol (format "struct:~a" (syntax-e nm)))
         ,(map t->s t)
         ,@(if proc (list (t->s proc)) null)
-        ,@(list (map t->s (unbox property-tys))))]
+        ,@(free-id-set->list properties))]
```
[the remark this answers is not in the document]
capfredf: Do you mean `(map t->s  (free-id-set->list properties))`? 
`properties` is a set of the identifiers for property descriptors, not types

typed-racket-lib/typed-racket/types/subtype.rkt
capfredf: I think simply `return #false` for the whole case would be better

typed-racket-lib/typed-racket/private/parse-type.rkt
bennn: `Exist (A ....) T` still sounds good. Can you draft docs for this type first, and pick a name after?

typed-racket-lib/typed-racket/utils/eq-contract.rkt
```diff
@@ -0,0 +1,12 @@
+#lang racket/base
+(provide eq/c)
+(require racket/contract/combinator)
+
+(define (eq/c name)
+  (make-contract #:name (format "not the same as ~a" name)
```
samth, line 6: Maybe "not an identical value to ~a"

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -864,17 +864,7 @@ functions and continuation mark functions.
           (list-length (list 1 2 3))]}
           
 @defform*[[(Exist (a ...) t)]]{ is an existential type where fresh variables
-@racket[a ...] may appear in @racket[t]. Currently there are no packing or
```
samth: Keep this but drop the non-working example.

typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -1021,6 +1011,11 @@ prefab types with the (implicitly quoted) prefab-key
 
   @history[#:added "1.10"]}
 
+@deffrom[(Has-Struct-Property prop)]{
```
samth: Is the a tech link for struct type property?

typed-racket-test/succeed/test-spt-untyped-typed-untyped.rkt
```diff
@@ -18,7 +20,8 @@
   #;(bar (world)))
 
 
-(require (submod "." typed))
-(struct world [] #:property prop:foo (λ (self) 10))
-(define x (world))
-((foo-ref x) x)
+;; (require 'untyped)
```
samth: Why are these commented out?

samth:
@bennn Any thoughts on this, which is now ready to merge?

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -511,15 +517,42 @@
...
+                                                            (same eq-name))))
+        (exist/sc (list #'name) lhs rhs)]
        [(F: v)
-        (triple-lookup
-         (hash-ref recursive-values v
-                   (λ () (error 'type->static-contract
-                                "Recursive value lookup failed. ~a ~a" recursive-values v)))
-         typed-side)]
+        (cond
+          [(string-prefix? (symbol->string v) "self-")  (if (not (from-untyped? typed-side))
```
bennn, line 546: style: indentation

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -511,15 +517,42 @@
...
+                                                            ;; if self is in negative position, we can't generate a contract yet. 
+                                                            (fail #:reason "contract generation not supported for Self")
+                                                            any/sc
+                                                           )]
+          [(triple-lookup
```
bennn, line 551: please add `else`

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -511,15 +517,42 @@
...
         (prompt-tag/sc (map t->sc ts) (t->sc s))]
-       [(F: v) #:when (string-prefix? (symbol->string v) "self-")
-               (fail #:reason "contract generation not supported for Self")]
-       ;; TODO
+       [(Exist: (list n) (Fun: (list li-arrs ...)))
+        (define (occur? t)
+          (if (or (not t) (empty? t)) #f
+              (set-member? (free-vars-names (free-vars* t)) n)))
+
+        (define (err) (fail #:reason "contract generation only supports Exist Type in this form: (Exist (X) (-> ty1 .. (-> X ty ... ty2) : X)"))
```
bennn, line 525: the message is missing a `)` ... before the `:`?

and change `..` to `...`

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -683,7 +716,30 @@
...
+        ;; so we need to check if it is produced by the property
+        (flat/sc #`(if (not (struct-type-property-predicate-procedure? #,pred? #,real-prop-var))
+                       (raise-arguments-error 'struct-property
+                                              "predicate does not match property"
+                                              "predicate"
+                                              #,pred?
+                                              "property"
+                                              #,real-prop-var)
+                       (flat-named-contract '#,real-pred-var
+                                            #,pred?)))]
```
bennn, line 742: Should this be `(lambda (x) pred?)`? That`s what `Struct: ` uses, and it's important when the contract def comes above the struct def.

I don't have an example for why Struct: needs a lambda on hand, but `require/typed` might be important to make one.

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -683,7 +716,30 @@
...
+        ;; the original-name of `prop-name` is its original referece in the unexpanded program.
+        (define real-prop-var (or (syntax-property prop-name 'original-name) prop-name))
+
+        ;; a property is wrapped so we need its original reference
+        (define real-pred-var (or (syntax-property pred? 'original-name) (syntax-e pred?)))
+        
+        ;; the `pred?` could be provided to a property through require/typed, 
+        ;; so we need to check if it is produced by the property
+        (flat/sc #`(if (not (struct-type-property-predicate-procedure? #,pred? #,real-prop-var))
+                       (raise-arguments-error 'struct-property
```
bennn, line 735: This error should probably go under a lambda, so it's not a bother unless the code actually uses the contract.

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -543,6 +552,16 @@
...
 
+;; body is a type
+(def-type Exist ([n exact-nonnegative-integer?]
+                 [body Type?])
+  #:no-provide
+  [#:frees (f) (f body)]
+  [#:fmap (f) (make-Exist n (f body))]
+  [#:for-each (f) (f body)]
+  #;
+  [#:mask (λ (t) (mask (Exist-body t)))])
```
bennn, line 563: delete this line and the `#;` ?

typed-racket-lib/typed-racket/static-contracts/combinators/exist.rkt
```diff
@@ -0,0 +1,51 @@
...
+         racket/list
+         racket/match
+         racket/syntax
+         (contract-req)
+         (for-template racket/base racket/contract/base)
+         (for-syntax racket/base syntax/parse))
+
+(provide exist/sc:)
+(provide exist/sc)
+#;
```
bennn, line 14: time to re-enable this contract?

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -538,3 +543,18 @@
...
+    [(Struct-Property: ty #f)
+     (set-struct-property-pred! struct-property-ty pred-id)
+     (define has-spty (make-Has-Struct-Property prop-name))
+     (values->tc-results (make-Values (list (-result struct-property-ty -true-propset -empty-obj)
+                                            (-result (make-pred-ty has-spty))
+                                            (-result (make-Fun (list (-Arrow (list has-spty) ty
+                                                                             #:props (-PS (-is-type 0 -Self)
+                                                                                          (-not-type 0 -Self))))))))
+                                             '())]
+    [(Struct-Property: ty _) (tc-error "the predicate for ~a should not be set mannualy " (syntax-e prop-name))]
```
bennn, line 559: typo, want "manually"

it'd also be good to add "use #f instead" if that's what the user should do

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -511,15 +517,42 @@
...
+                                                   (PropSet: (TypeProp: _ (F: n1)) _) _)))))
+             #:when (and (not (ormap occur? (list rst kw rst_i kw_i)))
+                         (eq? n1 n))
+             (values dom rng)]
+            [_ (err)]))
+        
+        (define-values (dom rng) (get-dom-rng li-arrs))
+        (define/with-syntax name n)
+        (define lhs (t->sc/neg dom))
+        (define eq-name (flat/sc #'(eq/c name)))
```
bennn, line 540: `eq/c` is probably much slower than `(lambda (x) (eq? x name))` ... do we really need `make-contract`?

typed-racket-lib/typed-racket/private/type-contract.rkt
capfredf: like `(lambda _ (raise-arguments-error ...))` ? If so, we would need wrap the lambda into a contract because flat/sc only accepts a contract. But that would look a little weird to me, since I don't usually see a contract always raise an errro.

My intention was to fail fast when an unmatched predicate was given, but delaying the error looks better for this code. 

So I think maybe the whole if expression should go under a lambda, which we turn into a flat-named-contract

typed-racket-lib/typed-racket/private/type-contract.rkt
capfredf: we probably need make-contract to make related error messages readable.

capfredf:
@bennn I have fixed all the issues you mentioned except the last one. you can take a look when you are available.

typed-racket-lib/typed-racket/private/type-contract.rkt
bennn: does this file still need to import `racket/contract/combinator` ?

if so, please move the require up with the others (from line 54)

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -184,8 +185,10 @@
       typed-racket/utils/sealing-contract
       typed-racket/utils/promise-not-name-contract
       typed-racket/utils/simple-result-arrow
+      typed-racket/utils/eq-contract
       racket/sequence
-      racket/contract/parametric))
+      racket/contract/parametric
+      racket/contract/combinator))
```
bennn, line 191: Unused import?

PR #864, opened 2019-10-02
Avoid using `call/ec` in typed `for/vector` expansion.

samth:


gus-massa:
Is it correct to use `unsafe-fx=` in https://github.com/racket/typed-racket/pull/864/commits/c1fe11f683fce7267e8374ad7001a76c219da9c4#diff-72cf4e4479067e30d7f7287eeb580300R820 ? I think the `#:break` checks runs before the creation of the vector.

samth:
This caused several tests to take much much longer, such as `math-test/math/tests/array-tests.rkt`. Need to investigate what's going on, but this may have to be reverted.

Issue #457, opened 2016-11-12
Unsoundness from stateful #:opaque predicates


AlexKnauth:
##### What version of Racket are you using?

I'm using version 6.7.0.1

### If you got an error message, please include it here.

I didn't get an error message. I added the number `16` to the byte string `#"abcdefg"` and got a `box` pointing to the number `3136` as the answer.

##### What should have happened?

I'm not sure. Maybe occurrence typing has to work differently for `#:opaque` predicates?

##### What program did you run?

```racket
#lang typed/racket

(module untyped racket
  (provide x? reset!)
  (define b (make-parameter #true))
  (define (x? x)
    (cond [(b) (b #false) #true]
          [else #false]))
  (define (reset!) (b #true)))

(require/typed 'untyped
               [#:opaque X x?]
               [reset! (-> Void)])

(: ->false : (All (T) [Any -> Boolean : #:+ T]))
(define (->false v) #false)

(: transmute : (All (T) Any -> T))
;; This ends up always returning v
(define (transmute v)
  (reset!)
  ;; (x? v) returns #true the first time
  (if (or (x? v) ((inst ->false T) v))
      (if (not (x? v)) ; but #false the second time
          v
          (error "this never happens"))
      (error "this never happens")))

(+
           #b00000000000000000000000000010000
           (transmute ;(inst transmute Positive-Index)
            (string->bytes/utf-8 "abcdefg")))
```

AlexKnauth:
What should happen?

It depends on the definition of `(Opaque x?)`. Defining it as the set of values for which `x?` returns (present tense) true is not specific enough. There are four options that I see for defining it more precisely:

1. The set of values for which `x?` has ever returned true.

2. The set of values for which `x?` has always and will always return true.

3. The set of values for which `x?` has returned true and has not returned false (yet).

4. The set of values for which `x?` has ever or will ever return true but has not and will never return false.

Under definition 1, `x?` could have the type `[Any -> Boolean : #:+ X]`, but not `[Any -> Boolean : X]` (where `X` is `(Opaque x?)`). Under definition 2, `x?` could have the type `[Any -> Boolean : #:- (! X)]`. With definitions 3 and 4, I'm not sure.

The current system assumes that `x?` is not stateful, so it ends up looking a little bit like definition 3, and when state is involved it causing this unsoundness. The combination of definition 3 with the current way of dealing with occurrence typing using intersections causes this unsoundness.

@pnwamk Do you have any ideas?

[...]

PR #867, opened 2019-10-23
WIP: Add Type/Predicate and make-positive-predicate

AlexKnauth:
Addresses https://github.com/racket/typed-racket/issues/457 by adding `Type/Predicate` and `#:type/predicate` as sound versions of what's most useful about `Opaque` and `#:opaque`.

- [x] 1. A type constructor `Type/Predicate` that takes a predicate id and produces a type. If `foo?` is a predicate, then `(Type/Predicate foo?)` produces a type for values that have passed `foo?` even just once. So if `(foo? x)` returns `#true` and then returns `#false`, `x` still has type `Foo` after that.
 
- [x] 2. A required/typed keyword `[#:type/predicate Foo foo?]` that imports the predicate `foo?` and defines the type `Foo` as `(Type/Predicate foo?)`. The `foo?` predicate is imported with type `(-> Any Boolean : #:+ Foo)`. It can _not_ have type `(-> Any Boolean : Foo)`.

- [x] 3. A new expression form `make-positive-predicate` that takes a type and produces a predicate with only the positive `#:+` proposition. So `(make-positive-predicate X)` will produce a predicate with the type `(-> Any Boolean : #:+ X)`.
   * `(make-positive-predicate (Type/Predicate foo?))` produces a predicate equivalent to `foo?` with type `(-> Any Boolean : #:+ (Type/Predicate foo?))`
   * `(make-positive-predicate (Opaque foo?))` also produces a predicate equivalent to `foo?` with type `(-> Any Boolean : #:+ (Opaque foo?))`

- [ ] 4. Make sure `make-predicate` errors on `Type/Predicate` types, with an error message that suggests using `make-positive-predicate` instead. It should give a warning on `Opaque` types, similarly suggesting `make-positive-predicate` instead.

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/libraries.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/libraries.scrbl
@@ -287,7 +287,7 @@ following steps are required:
   @item{Determine the data manipulated by the library, and how it will
   be represented in Typed Racket.}
   @item{Specify that data in Typed Racket, using @racket[require/typed]
-  and @racket[#:opaque] and/or @racket[#:struct].}
+  and @racket[#:type/predicate] and/or @racket[#:struct].}
   @item{Use the data types to import the various functions and constants
   of the library.}
   @item{Provide all the relevant identifiers from the new adapter module.}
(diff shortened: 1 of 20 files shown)


samth:
What happens if we just change `#:opaque` to only have a positive predicate? How much breaks?

I'd really prefer not to add a lot more mechanism here.

samth:
That's also true of your proposal, I believe.

AlexKnauth:
Yes, with my proposal `make-predicate` would be a warning on `Opaque` types, which would guide programmers toward the sound version.

samth:
I should also say that this change will _definitely_ require an RFC, and I encourage you to first write that and go through the process before working too much more on the implementation.

AlexKnauth:
Closed to be replaced by https://github.com/racket/typed-racket/pull/882

PR #869, opened 2019-11-06
fix a bug where #:construct-name is used in struct definitions

capfredf:


capfredf:
@samth I have added two testcases.
The package `psd' was built successfully with this fixed patch.

PR #880, opened 2019-11-24
Only allow 29 bits of mask size, to support 32-bit Racket CS.

samth:
Drop `mask:future` since subtyping with future values seems
uncommon.

Related to racket/racket#2925.


bennn:
Was it necessary to delete `mask:future`, or was this pre-emptive to have 3 slots available in the future?

(IIUC before this PR there were 27 mask bits in use --- I'm thinking I misunderstand something)

samth:
When I counted it looked like there were 30.

samth:
I was wrong, see 10222f20da2a3.

PR #886, opened 2019-12-11
more docs for #:opaque

bennn:
- add a small example to the guide
- edit the reference paragraph on `#:opaque` to:
  1. say "opaque" in the first sentence (also using a `deftech`)
  2. avoid starting/ending sentences with code
  3. add a link to the `define-predicate` section to explain what a predicate is

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/guide/typed-untyped-interaction.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/guide/typed-untyped-interaction.scrbl
@@ -73,6 +73,49 @@ typed/racket
 is a valid use of the @racket[require/typed] form and imports @racket[add1]
 from the @racketmodname[racket/base] library.
 
+
+@subsection{Opaque Types}
+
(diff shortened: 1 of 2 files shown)


MarcKaufmann:
This looks good to me, given the extent to which I can parse these git pulls and merges. On the example, I think the main issue isn't that it's simple. Rather it doesn't make clear why I would want or have to define a type via a predicate, rather than directly as a type. In the case of `gregor`, I had to do it because no module exports the `time` or `Time` struct. Therefore I would either have to type  the whole struct in order to check that it works, or use `#:opaque`. But in this case, defining the type for the struct wouldn't work, because I can't peek inside of the struct, so I have to use opaque types.

It is not quite clear to me what other implications the `#:opaque` version has that might be different from defining the types when I have a transparent struct. I guess in the case of `point` I could either define it as an opaque type with `point?` or define my own point structure and use that, right? Are there cases where I need opaque types that are not about structs? It might be less work to use opaque types, but it might not be necessary. 

Thanks a lot for taking the time to do the pull request.

PR #888, opened 2019-12-16
Provide TR analogue of the web-server's response/empty convenience function

jessealama:
Should I also bump the version of typed-racket-more?

### DIFF
--- a/typed-racket-more/typed/web-server/http.rkt
+++ b/typed-racket-more/typed/web-server/http.rkt
@@ -231,3 +231,13 @@
                             [#:headers (Listof Header)]
                             [#:cookies (Listof Cookie)]
                             Response)])
+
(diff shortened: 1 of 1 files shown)


samth:
Will this work on tomorrow's snapshot?

jessealama:
Hmm, it looks like I'm one step ahead: the current version of the web server is 1.5, not 1.6 as I thought. I'll dump the commit that broke the build.

cc @jeapostrophe

samth:
This broke tests here: http://drdr.racket-lang.org/53392/racket/share/pkgs/typed-racket-more/typed/web-server/http.rkt

@jessealama do you know what went wrong?

jessealama:
Looks like you fixed this -- thanks! Sorry for the trouble.

PR #889, opened 2019-12-16
add index-of to base-env.rkt

jbclements:
This PR adds the "index-of" primitive to typed racket. I couldn't find a place to put any test cases; here's the code I would use, though:

``` racket
#lang typed/racket

(require typed/rackunit)

(check-equal? (index-of '(a 12 c d) 'd) 3)
(check-equal? (index-of '(a 12 c d) 'e) #f)
(check-equal? (index-of
               '("ab" "cde" "fhgi")
               3
               (λ ([str : String] [len : Integer])
                 (= (string-length str) len)))
              1)
```


samth:
Unit tests would belong in `typed-racket-test/unit-tests/typecheck-tests.rkt`. Integration tests (such as the file you have there) can just be put in `typed-racket-test/succeed` and they get run automatically.

jbclements:
Okay, test added. Let me know if you want me to squash them.

PR #891, opened 2019-12-18
Remove inaccurate keyword from response/empty

jessealama:


jessealama:
mistaken repo state; an improved pull request will follow

PR #899, opened 2020-01-13
update typed/web-server/http for v1.6 (typed-racket-more v1.13)

LiberalArtist:
- Add types related to `binding:file/port-in`.

- Strengthen the type of the `#:timeout` arguments to
  `valid-id-cookie?` and `request-id-cookie` from `Number` to `Real`,
  analogously to <https://github.com/racket/typed-racket/commit/acf7707>.
  (A non-real `#:timeout` has always been a runtime error and in one
  case was already prohibited by a contract on the untyped side.)
  Usually you don't actually want to be using the `#:timeout` argument:
  see <https://github.com/racket/web-server/commit/626ad02>.

Related to <https://github.com/racket/web-server/pull/84>.

### DIFF
--- a/typed-racket-more/typed/web-server/http.rkt
+++ b/typed-racket-more/typed/web-server/http.rkt
@@ -31,6 +31,13 @@
                                   [headers : (Listof Header)]
                                   [content : Bytes])
                                  #:extra-constructor-name make-binding:file]
+                       ;; It would be better if we could tell Typed Racket
+                       ;; that `binding:file/port` is a subtype of `binding:file`.
+                       [#:opaque binding:file/port binding:file/port?]
+                       [binding:file/port-in (-> binding:file/port Input-Port)]
+                       [make-binding:file/port
+                        (-> Bytes Bytes (Listof Header) Input-Port
+                            binding:file/port)]
                        [#:struct request
                                  ([method : Bytes]
                                   [uri : URL]
@@ -107,6 +114,10 @@
          Cookie-Value
          Valid-Domain)
 
+;; N.B.: The untyped cookie functions will implicitly convert bytes to
+;; strings on input, partially for compatability, but that doesn't seem
+;; very desirable in a typed context.
+
 (require/typed/provide
  web-server/http/cookie
  [make-cookie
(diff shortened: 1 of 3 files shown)


LiberalArtist:
I'm interested in a better solution for `binding:file/port?` if there is one.

typed-racket-more/typed/web-server/http.rkt
```diff
@@ -145,14 +156,14 @@
                         (-> String
                             Bytes
                             Request
-                            [#:timeout Number]
```
samth: This change isn't backwards-compatible so I think we should skip it.

typed-racket-more/typed/web-server/http.rkt
```diff
@@ -31,6 +31,13 @@
                                   [headers : (Listof Header)]
                                   [content : Bytes])
                                  #:extra-constructor-name make-binding:file]
+                       ;; It would be better if we could tell Typed Racket
+                       ;; that `binding:file/port` is a subtype of `binding:file`.
+                       [#:opaque binding:file/port binding:file/port?]
```
samth: The better solution would require writing out the type environment manually.

typed-racket-more/typed/web-server/http.rkt
LiberalArtist: I'll update the PR momentarily. I've been conflicted about this change (e.g. <https://github.com/racket/web-server/pull/75#issuecomment-552677557>). As you say, it isn't strictly backwards-compatible, though for various reasons it seems unlikely to be a problem: in particular, <https://github.com/racket/typed-racket/commit/acf7707> had an analogous change from `Number` to `Real` that seems to have gone fine.

LiberalArtist: The fact that this seems to come up quite often with `Number`/`number?` makes me wonder if we should think about what to call it in Rhombus …

LiberalArtist:
The CI will fail until racket/web-server#84 makes it into a snapshot.

samth:
That should happen tonight, and then we can check CI.

PR #907, opened 2020-02-12
require/untyped-contract: adjust for relative submod path

bennn:
Add a ".." to any relative submodule paths.

- - -

Example program (test.rkt):

```
  #lang racket/base
  ;; test.rkt

  (module a typed/racket
    (define (f (x : (U Symbol String))) : Void (void))
    (provide f))

  (module b racket/base
    (require typed/untyped-utils typed/racket/base)
    (require/untyped-contract (submod ".." a) (f (-> Symbol Void)))
    (void))
  (require 'b)
```

Without adjustment, `require/untyped-contract` expands to an invalid
require. This new require appears in a new submodule but points to
the same path `(submod ".." a)`, which is not the same goal.

The old error message is:

```
  syntax-local-module-exports: unknown module
    module name: #<resolved-module-path:(submod 'test[7305] b a)>
```

### DIFF
--- a/typed-racket-more/typed/untyped-utils.rkt
+++ b/typed-racket-more/typed/untyped-utils.rkt
@@ -39,13 +39,18 @@
                     [*typed/racket/base (datum->syntax #'from-module-spec
                                                        'typed/racket/base)]
                     [*require (datum->syntax #'from-module-spec
-                                             'require)])
+                                             'require)]
+                    [from-module-spec-for-submod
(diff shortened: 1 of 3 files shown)


samth:
This looks like it matches any submod require at all, not just ones that are relative to the current module.

bennn:
Oops, thats right.

Fixed now to not modify non-relative submods

AlexKnauth:
What Racket version does this pull request correspond to, between 7.7 and 7.8?

PR #911, opened 2020-03-03
type-contract: look for lists, make list/sc

bennn:
Check for proper list types and output a matching static contract,
 instead of using cons/sc for all kinds of pairs

- - -

Looks like we never make `list/sc` contracts currently --- even though there are sc-optimizer cases for them.

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -529,8 +529,21 @@
        [(Mutable-VectorTop:)
         (only-untyped mutable-vector?/sc)]
        [(Box: t) (box/sc (t->sc/both t))]
-       [(Pair: t1 t2)
-        (cons/sc (t->sc t1) (t->sc t2))]
+       [(Pair: t-car t-cdr)
+        ;; look ahead as long as t-cdr is a Pair
+        (define-values [t-last rev-sc*]
+          (let loop ((t t-cdr)
+                     (sc* (list (t->sc t-car))))
+            (match t
+             [(Pair: t-car t-cdr)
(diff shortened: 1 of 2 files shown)


typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -529,8 +529,16 @@
        [(Mutable-VectorTop:)
         (only-untyped mutable-vector?/sc)]
        [(Box: t) (box/sc (t->sc/both t))]
-       [(Pair: t1 t2)
-        (cons/sc (t->sc t1) (t->sc t2))]
+       [(Pair: _ _)
+        (match type
+         [(List: elem-t*)
+          (apply list/sc (map t->sc elem-t*))]
+         [(List: elem-t* #:tail rest-t*)
```
samth: This seems like it has quadratic complexity in the length of the list for improper lists.

samth:
Is there an advantage to the `list/c` contracts?

rfindler:
I forget the details now but list-contract? Returning true has some
implication that I think we done to support TR.

bennn:
I tried grepping for `list-contract?` in my Racket clone & packages, but didn't see any implications. Either way, though, the current "nested pairs" contract is also a `list-contract?` ([link](https://github.com/racket/racket/blob/master/racket/collects/racket/contract/private/list.rkt#L438)).

So I think the only benefits of this PR are:
1. give a faster answer to the `list-contract?` question
2. enable two [sc optimizations](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/static-contracts/optimize.rkt#L28): `(list any ...n) -> (list-length? n)` and `(list x ... none/c x ...) -> none/c`

PR #912, opened 2020-03-13
tc-app-lambda: register type for ((lambda ....) ...)

bennn:
Register a type for the anonymous function in a ((lambda ....) ....)
 expression after inferring its type from the arguments.

Also add -true-propset to the "golden" types in a few tests.
(The tests pass now, but look uglier.)

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-lambda.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-lambda.rkt
@@ -6,7 +6,7 @@
          (prefix-in - (contract-req))
          syntax/parse racket/match racket/list
          racket/sequence
-         (typecheck signatures find-annotation)
+         (typecheck signatures find-annotation tc-metafunctions)
          (types abbrev utils generalize type-table)
          (private type-annotation syntax-properties)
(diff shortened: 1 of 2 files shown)


typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-lambda.rkt
```diff
@@ -35,7 +35,14 @@
    #:fail-unless (= (syntax-length #'(x ...))
                     (syntax-length #'(args ...))) #f
    #:fail-when (andmap type-annotation (syntax->list #'(x ...))) #f
-   (tc/let-values #'((x) ...) #'(args ...) #'body expected))
+   (let* ([res (tc/let-values #'((x) ...) #'(args ...) #'body expected)]
+          [dom-ts
+           (for/list ([arg (in-list (syntax-e #'(args ...)))])
+             (match (type-of arg)
```
samth: Do we really not need annotations from the xs here?

bennn: Do you mean: should we ask for `(type-of x) ...` instead?
I'm not sure; they might not have a registered type after `tc/let-values` finishes.

PR #922, opened 2020-04-01
Change source of web-server's response/empty function

jessealama:
This is the TR analogue of a proposed change to the
web-server (https://github.com/racket/web-server/pull/85),
in which `web-server/http/empty` goes away, having been
incorporated into `web-server/http/response-structs`.


jessealama:
It's expected that the build will fail until the web-server's pull request gets merged.

PR #923, opened 2020-04-03
typed/pict: add missing struct field

bennn:
Currently, `(pict-children (blank 50 50))` = `0`


samth:
This is fine, but how did a type specification bug change that behavior?

bennn:
The bad type specification renames the struct accessors. Typed `pict-children` gets the real `pict-descent`, typed `pict-panbox` gets `pict-children` ... and the untyped `pict-last` has no accessor.

This seems like something the type environment should be able to catch. I'll try to look at that tomorrow.

PR #925, opened 2020-04-08
consistency check for #:struct clauses

bennn:
Add two checks to require/typed and type-environment:

1. is the struct name bound to compile-time struct info?
2. do the fields match what we know about the struct?

There have been two big complications so far, though.

First & most serious, some files need to opt-out of the checking.
`typed/racket/sandbox` is one example

- it needs to opt-out because it describes a struct that isn't exported
  (exn:fail:resource)
- it's safe to opt-out because it only re-provides things that really
  are exported (exn:fail:resource? exn:fail:resource-resource)

Currently using the #:no-provide keyword to opt-out rather than making
 a new one.

Second, `contract-out` mangles selector names. Other tools might do that
too, so check 2 above does nothing if we cannot find selectors that
match a simple regexp.


bennn:
The build failed because of the pict `child` struct field-ordering issue (https://github.com/racket/pict/pull/60) --- of course.

I changed the order in `typed/pict` to match the current fields (6f2202e); let's see if there are other errors.

PR #926, opened 2020-04-10
require/typed struct: provide the static struct info

bennn:
require/typed creates a `define-syntax` for struct info;
 make sure this is provided directly, and not redirected to
 a type-name-error syntax transformer

- - -

For example, this program currently works:

```
  #lang racket/base

  (module t typed/racket
    (struct foo ((a : Symbol)))
    (provide (struct-out foo)))

  (require 't racket/match)
  (match-define (foo a0) (foo 'xxx))
```

And so this next program ought to work, but raises a syntax error in the
match pattern because `foo` is bound to a syntax transformer and not
a `struct-info?`

```
  #lang racket/base

  (module u typed/racket/base
    (struct foo ((a : Symbol)))
    (provide (struct-out foo)))

  (module t typed/racket
    (require/typed/provide (submod ".." u)
      (#:struct foo ((a : Symbol)))))

  (require 't racket/match)
  (match-define (foo a0) (foo 'xxx))
```

- - -

At first I tried to solve this problem with some a `def-struct-stx-binding`, but I don't think this require/typed code can fit into that path ... we'd need a new kind of `def-binding`

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -546,10 +546,15 @@
                                                        (id-drop orig-sels orig-muts num-fields)))
                                            (struct-info-list new-sels new-muts)))))))
 
-                         (define-syntax nm
-                              (if id-is-ctor?
-                                  (make-struct-info-self-ctor #'internal-maker si)
-                                  si))
+                         #,(ignore
+                             ;; provide the static struct info directly, unlike other define-syntax forms
+                             ;; this is similar to the struct quad code in `typecheck/provide-handling.rkt`
+                             #'(begin
+                                 (define-syntax nm
+                                   (if id-is-ctor?
+                                     (make-struct-info-self-ctor #'internal-maker si)
+                                     si))
+                                 (provide nm)))
 
                          (dtsi* (tvar ...) spec type (body ...) #:maker maker-name #:type-only)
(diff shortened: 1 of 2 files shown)


samth:
I'm confused. This seems like it would leave the resulting module with a `provide` even if nothing were intentionally provided by the author. Is that right?

bennn:
You're right. This needs work.

capfredf:
Based on my recent experience with tinkering structure handling in TR, my gut reaction is a new type of def-binding would be more or less identical to the `def-struct-stx-binding`, so I am not sure it would be worth it. 

my understanding of this issue is `require/typed(/provide)` introduces structure types but doesn't go through all the whole process of structure handling like a typed structure definition does. So I made a bold move to prove my hypothesis:

https://github.com/capfredf/typed-racket/commit/840545ac2767d20e4205346c57a7817676c87f51

(alternatively, I could remove `#:type-only` on https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/base-env/prims-contract.rkt#L523, but the result should be the same)

bennn:
Thanks for looking into this @capfredf 

With your patch (or without `type-only`), does the `require/typed` still add the right contracts?

In general, we need to make sure that nothing unsafe gets imported & nothing extra gets exported (and write missing regression tests).

capfredf:
@bennn My patch has other problems, e.g. `(foo 10)` in the main untyped module won't work. It should not be too hard to fix though. I will come back to this problem later

bennn:
#1078 is much better

PR #927, opened 2020-04-11
type-environment: optionally provide struct constructor

bennn:
Let `type-environment` accept the struct #:kernel-maker option,
 and pass it on to `d-s`.

Without this option, type-environment does not register a type for
 a struct constructor.

Also, delete some comments that have been replaced by contracts.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
@@ -172,8 +172,6 @@
 
 
 ;; construct all the various types for structs, and then register the appropriate names
-;; identifier listof[identifier] type listof[fld] listof[Type] boolean ->
-;;  (values Type listof[Type] listof[Type])
 (define/cond-contract (register-sty! sty names desc)
   (c:-> (c:or/c Struct? Prefab?) struct-names? struct-desc? void?)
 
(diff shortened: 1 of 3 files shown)


typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -517,7 +512,7 @@
   (define sty (mk/inner-struct-type names desc parent-type))
 
   (register-sty! sty names desc)
-  (register-struct-bindings! sty names desc #f)
+  (void (register-struct-bindings! sty names desc #f))
```
bennn, line 515: This may need a comment (but I'm not sure how to word it) because,

- first thought: `register-struct-bindings!` must return `(void)` because of the `!`
- second thought: this must be a bug, because we're ignoring a list of bindings

It took a while for me to realize that this function `tc/builtin-struct` runs at a different time than the similar-named `tc/struct`. The latter runs during type checking and helps make the `#%type-decl` module. This function appears in the `#%type-decl` module and so cannot use the list of bindings.

bennn:
The latest test failure is interesting:

```
Run racket -l typed-racket-test -- --unit
--------------------
Tests for infer > All inference tests > Unit Tests > Typed Racket Tests > ((List b ... b)) ((Listof Symbol))
FAILURE
location:   <pkgs>/typed-racket-test/unit-tests/infer-tests.rkt:201:3
actual:     (Listof Nothing)
expected:   (Listof Nothing)

Did not infer the expected result.
--------------------
3318 success(es) 1 failure(s) 0 error(s) 3319 test(s) run
Typed Racket Tests did not pass.
```

PR #936, opened 2020-06-08
support test submodule in integrated tests

capfredf:
Looks like this [test](https://github.com/racket/typed-racket/blob/master/typed-racket-test/succeed/submodules.rkt) has not properly run for a while ...


samth:
Turns out this revealed a problem with the type of `current-command-line-arguments`, which had been there for quite a while.

PR #939, opened 2020-06-14
Fix <method not found> and in-fxvector in for form

sorawee:
- Correct the method names that were "not found".
  See also racket/gui#185
- Add unsafe-fxvector-length and unsafe-fxvector-ref
  so that in-fxvector in the for form works correctly.
  Add corresponding tests.
  Thanks to James Cooper who reported the issue in Slack.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -3174,6 +3174,8 @@
 [unsafe-struct*-ref top-func]
 [unsafe-struct-set! top-func]
 [unsafe-struct*-set! top-func]
+[unsafe-fxvector-length (-FxVector . -> . -Fixnum)]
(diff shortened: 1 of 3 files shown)


typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -3174,6 +3174,8 @@
 [unsafe-struct*-ref top-func]
 [unsafe-struct-set! top-func]
 [unsafe-struct*-set! top-func]
+[unsafe-fxvector-length (-FxVector . -> . -Fixnum)]
```
samth, line 3177: The result type can be `-Index`

sorawee:
@samth do you know why if I put them in `base-env-indexing-abs.rkt`, then the type checker doesn't seem to recognize them?

samth:
Probably there's a missing require for-template.

Issue #944, opened 2020-06-26
Unaesthetic error when domain of variable argument function does not match given


maueroats:
### What version of Racket are you using?
7.7
### What program did you run?
```racket
#lang typed/racket
(: max* (-> Integer Integer * Integer))
(define (max* . xs)
  (apply max xs))
(define (one-more-than-max-bad [xs : (Listof Integer)])
    : Integer
    (add1 (apply max* xs)))
```
### What did happen?

An appropriate error message occurred, but note spacing in Domain.

```
Type Checker: Bad arguments to function in `apply':
Domain: IntegerInteger *
Arguments:  (Listof Integer)
 in: (apply max* xs)
```

### What should have happened?
Error message, but with a space between the two `Integer` types. 

### If you got an error message, please include it here.
(see "what did happen")

### Blame

If there is no `*` in the type signature, then the error message has the expected spaces in the domain.
Perhaps [stringify-domain](https://github.com/racket/typed-racket/search?q=stringify-domain&unscoped_q=stringify-domain) does not work properly in this particular case?

PR #946, opened 2020-06-29
fix issue #944 by adding an appropriate space in error message

maueroats:
Certain type errors with rest arguments lacked a space in their output, making them confusing for beginners. (What is the type `StringString`?)


maueroats:
This should close #944 .

PR #948, opened 2020-07-03
transient semantics

bennn:
Goal: add a new way to run typed/racket programs. The new types are weaker against untyped code, but should run faster.

2021-11-01: This PR is the main development.
~For now, this PR is a preview of the main development at <https://github.com/bennn/typed-racket/tree/transient> . I'm planning to keep a todo list here, listen to feedback, and push changes once I have them. Eventually, I want to merge this branch.~

### Notes for code review

- `utils/transient-rewrite.rkt` ~`defender/defender.rkt`~ rewrites expanded, unoptimized typed code with flat contract checks everywhere that an untyped value can enter at runtime.

- `private/type-contract.rkt` has a new function `type->static-contract/transient` that turns a type into a flat contract

- `cast` and other exports from `base-env/prims-contract.rkt` come in multiple versions with different semantics (Deep, Shallow, Optional)

- Other changes follow the `type-enforcement-mode?` box from `utils/tc-utils.rkt`. This new piece of state gets instantiate like the `typed-context?` box. It decides how to compile types to runtime checks.

### RESOLVED Design questions

Q. Transient code should be able to use typed/rackunit macros, but in general we can't allow guarded macros in non-guarded code. I propose using `unsafe-provide` for macros that are safe in transient (DONE https://github.com/bennn/rackunit/commit/32495bff0badecf6d1d3eea5c7be476fc3c7fcb1).

Q. `define-typed/untyped-identifier` should probably take a 3rd optional argument. Here's the problem: Transient needs to `(require/typed plot (plot-pict ....))` with a huge type because `plot` defines that as a typed/untyped id and so Transient gets an untyped identifier (it can't use the guarded one). That huge type should be defined once and for all with a require/typed on the library side. Sounds good? (YES, DO IT)

Q. The type environment files have changes to let the `defender` know whether transient can trust the results of a base-env function. These annotations are not enough for higher-order library functions; for example, `make-do-sequence` returns a function that we need to trust because TR's types are not dependent enough. For now, I have a hack to trust that function --- should we keep the hack for now, or try changing the type-rep to encode transient trust? (KEEP FOR NOW)

### Issues / TODO

- [x] write an RFC for the transient & erasure modes (#952 )
- [x] add `#lang`s ... instead of `#lang typed/racket #:transient` want `#lang typed/racket/transient` so we can use a repl and so `module+` is no longer unsound
- [x] ~add "expected" results to any `typed-racket-test/succeed` and `..../fail` files that have a different outcome on transient; change the tester to run each file on each mode (ditto for other tests)~ 2021-09-20: to keep this PR moving, I made a new folder `typed-racket-test/fail/transient/` ; the tester changes look difficult
- [x] in the `defender` review all "ignore" expressions ... maybe transient rewriting should never ignore code EDIT 2021-09-20: yes indeed, transient rewriting should not ignore expressions but rather dive into them looking for typed code
- [x] all tests pass
- [x] rename typed/racket/shallow/base to typed/racket/base/shallow etc. so that "how to enforce types" is always the last modifier (just like how no-check works)
- [x] docs for the new languages
  * add reference pages for Shallow and Optional, similar to the no-check page
  * consider splitting ref. section 2 (special forms) into typechecking & runtime forms; extend the docs for runtime forms
  * consider documenting the shape / "transient view" of static types alongside the type ... and maybe do the same for Deep
  * decide how to update the guide
- [ ] measure performance once things are in shape
  * might want to implement a cheap `and/c` to avoid racket/contract inside transient
  
### Future

Long-term issues for Transient. Record these on a project page?

- Transient-specific optimizations to avoid checks and/or improve their representation.
- Deep can optimize the contracts on exports to Shallow (need to change `typecheck/renamer.rkt` with another arg.)
  * `integer?` = no contract
  * `(-> integer? integer?)` = no contract
  * `(-> integer? (vectorof integer?))` = `(-> any (vectorof integer?))`
- edit: DONE ~Instead of keeping a "Transient env" of identifiers that don't need a result check (e.g. `map`), store that info IN the type itself. That way, a function can return a trusted function (e.g. `(negate void?)`).~

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/guide/typed-untyped-interaction.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/guide/typed-untyped-interaction.scrbl
@@ -6,6 +6,9 @@
 
 @(define the-eval (make-base-eval))
 @(the-eval '(require typed/racket))
+@(define shallow-eval (make-base-eval #:lang 'typed/racket/shallow))
+@(define optional-eval (make-base-eval #:lang 'typed/racket/optional))
+
 
 @title[#:tag "typed-untyped-interaction"]{Typed-Untyped Interaction}
 
@@ -14,7 +17,21 @@ that are entirely typed. One of the key features of Typed Racket is that
 it allows the combination of both typed and untyped code in a single
 program.
 
-@section{Using Untyped Code from Typed Code}
+From a static typing perspective, combining typed and untyped code is
+straightforward.
+Typed code must declare types for its untyped imports to let the type checker
+validate their use (@secref{untyped-in-typed}).
+Untyped code can freely import bindings from typed code (@secref{typed-in-untyped}).
+
+At run-time, combining typed and untyped code is complicated because there is a
+tradeoff between strong type guarantees and the performance cost of checking
+that untyped code matches the types.
+Typed Racket provides strong @emph{Deep} type guarantees by default, but offers two
+weaker options as well: Shallow and Optional types
+(@secref{protecting-interaction}).
+
+
+@section[#:tag "untyped-in-typed"]{Using Untyped Code in Typed Code}
 
 Suppose that we write the untyped module from @secref["quick"] again:
 
@@ -38,7 +55,7 @@ it. Since the untyped module did not specify any types, we need to annotate
 the imports with types (just like how the example in @secref["quick"] had
 additional type annotations with @racket[:]):
 
-@margin-note{Note that a typed module @emph{does not} need to use @racket[require/typed]
+@margin-note{Note that a typed module @emph{should not} use @racket[require/typed]
 to import from another typed module. The @racket[require] form will work in such
 cases.}
 
@@ -116,7 +133,7 @@ typed/racket
 
 
 
-@section{Using Typed Code in Untyped Code}
+@section[#:tag "typed-in-untyped"]{Using Typed Code in Untyped Code}
 
 In the previous subsection, we saw that the use of untyped code from
 typed code requires the use of @racket[require/typed]. However, the
@@ -128,28 +145,26 @@ to use the bindings defined in the typed module as expected. The major
 exception to this rule is that macros defined in typed modules may not
 be used in untyped modules.
 
-@section{Protecting Typed-Untyped Interaction}
+
+@section[#:tag "protecting-interaction"]{Protecting Typed-Untyped Interaction}
 
 One might wonder if the interactions described in the first two
-subsections are actually safe; after all, untyped code might be able to
+subsections are actually safe. After all, untyped code might be able to
 ignore the errors that Typed Racket's type system will catch at
 compile-time.
 
-To ensure that typed-untyped interactions are safe, Typed Racket establishes
-contracts wherever typed and untyped code interact. For example, suppose
+For example, suppose
 that we write an untyped module that implements an @racket[_increment]
 function:
 
-@margin-note{For general information on Racket's contract system
-, see @secref[#:doc '(lib "scribblings/guide/guide.scrbl")]{contracts}.}
-
-@examples[#:label #f #:eval the-eval
+@examples[#:eval shallow-eval #:hidden (module increment racket (provide increment) (code:contract increment : exact-integer? -> exact-integer?) (define (increment x) "this is broken"))]
+@examples[#:eval optional-eval #:hidden (module increment racket (provide increment) (code:contract increment : exact-integer? -> exact-integer?) (define (increment x) "this is broken"))]
+@examples[#:eval the-eval
 (module increment racket
   (provide increment)
 
   (code:contract increment : exact-integer? -> exact-integer?)
-  (define (increment x) "this is broken"))
-]
+  (define (increment x) "this is broken"))]
 
 and a typed module that uses it:
 
@@ -161,23 +176,55 @@ and a typed module that uses it:
   (increment 5))
 ]
 
-This combined program is not correct. All uses of @racket[_increment]
+This combined program has a problem. All uses of @racket[_increment]
 in Typed Racket are correct under the assumption that the
 @racket[_increment] function upholds the @racket[(-> Integer Integer)]
 type. Unfortunately, our @racket[_increment] implementation does not
 actually uphold this assumption, because the function actually produces
 strings.
 
-On the other hand, when the program is run:
+By default, Typed Racket establishes contracts wherever typed and untyped code
+interact to ensure strong types.
+These contracts can, however, have a non-trivial performance impact.
+For programs in which these costs are problematic, Typed Racket provides
+two alternatives. All together, the three options are Deep, Shallow, and Optional types.
+
+@itemlist[#:style 'ordered
+  @item{
+    @emph{Deep} types get enforced with comprehensive contract checks.
+  }
+  @item{
+    @emph{Shallow} types get checked in typed code with lightweight assertions
+    called @emph{shape checks}.
+  }
+  @item{
+    @emph{Optional} types do not get enforced in any way. They do not ensure
+    safe typed-untyped interactions.
+  }
+]
+
+@margin-note{See also: @secref["behavior-of-types" #:doc '(lib
+"typed-racket/scribblings/ts-reference.scrbl")] in the Typed Racket Reference.}
+The next subsections give examples of Deep, Shallow, and Optional behaviors.
+
+
+@subsection{Deep Types: Completely Reliable}
+
+When the @racket[_client] program above is run, standard Typed
+Racket (aka. Deep Typed Racket) enforces the @racket[require/typed]
+interface with a contract.
+This contract detects a failed type assumption when the @racket[_client]
+calls the untyped @racket[_increment] function:
 
 @examples[#:label #f #:eval the-eval (eval:error (require 'client))]
 
-we find that the contract system checks the assumption made by the typed
-module and correctly finds that the assumption failed because of the
-implementation in the untyped module (hence it is @emph{blamed} in the
-error message).
+Because the implementation in the untyped module broke the contract
+by returning a string instead of an integer, the error message
+@emph{blames} it.
 
-In the same fashion, Typed Racket checks all functions and other values
+@margin-note{For general information on Racket's contract system,
+see @secref[#:doc '(lib "scribblings/guide/guide.scrbl")]{contracts}.}
+In general, Deep Typed Racket checks all functions and other values
 that pass from a typed module to untyped module or vice versa with
 contracts. This means that, for example, Typed Racket can safely optimize
 programs (see @secref["optimization"]) with the assurance that the program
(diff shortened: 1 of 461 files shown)


bennn:
Aw hang on, I meant to have 1 commit here. May need to open a new pull request.

jackfirth:
Eagerly awaiting the RFC, since this sounds extremely cool but I don't understand it at all 🎉

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -727,6 +757,270 @@
...
+     ;; This comes before Base-ctc to use the Value-style logic
+     ;; for the singleton base types (e.g. -Null, 1, etc)
+     [(Val-able: v)
+      (cond
+       [(eof-object? v)
+        (flat/sc #'eof-object?)]
+       [(void? v)
+        (flat/sc #'void?)]
+       [(or (symbol? v) (boolean? v) (keyword? v) (null? v) (eq? unsafe-undefined v))
+        (flat/sc #`(lambda (x) (eq? x '#,v)))]
```
bennn, line 803: @samth FYI for my dissertation, this function is `(lambda (x) (void) (eq? x '#,v))` because of <https://github.com/racket/racket/issues/3339>

(If there's time to measure again later, I'll re-run without the `(void)` if possible)

bennn:
quick update: my dissertation is in and this pull request is now a top priority

I'll work on making the test suite reusable first, then get to working on this branch.

typed-racket-lib/typed-racket/base-env/base-env-indexing-abs.rkt
```diff
@@ -3,12 +3,18 @@
...
 
 (provide indexing)
 
-(define-syntax-rule (make-env* [i t] ...) (make-env [i (λ () t)] ...))
+(define-syntax (make-env* stx)
+  (syntax-parse stx
+   [(_ [id ty (~optional trusted-positive? #:defaults ([trusted-positive? #'#true]))] ...)
+    #'(make-env [id ty trusted-positive?] ...)]))
+
+(define needs-transient-check #false)
```
bennn, line 17: note: these env changes are a hack

Shallow TR usually changes `(f x)` to a `(check T (f x))` to spot-check the values that `f` returns. The env changes let us skip the check for certain identifiers `f`. Ideally we'd extend the language of types to make that decision ... but I think we should leave that for the future.

TODO write a comment somewhere that explains this

typed-racket-lib/typed-racket/env/lexical-env.rkt
```diff
@@ -95,6 +99,15 @@
...
                                    t)]
+                             [(and (let* ([fail (λ () #f)]
+                                          [v (syntax-local-value i fail)]
+                                          [orig-id (and (provide/contract-info? v)
+                                                        (provide/contract-info-original-id v))])
+                                     (and (identifier? orig-id)
+                                          (lookup-type orig-id fail))))
+                              => (λ (orig-t)
+                                   (register-type i orig-t)
+                                   orig-t)]
```
bennn, line 110: @samth this is the part of Shallow TR that goes behind contracts to get types.

Ideally we'd use the `typecheck/renamer.rkt` interface to get types, but (I believe) the typechecking code here gets run after expansion => after those rename transformers have been expanded to the contracted identifiers.

Another idea is to serialize more types for the new identifiers, but I was never able to make that work. The typed id's come from one submodule & the contract versions come from another ... and I couldn't find a meeting point.

bennn:
Right now, Shallow code (transient) exports ids using basically the same renamer strategy as normal Deep code. If a Shallow export appears in Shallow or untyped code, it expands to the plain id. In Deep code, it expands to a contract-wrapped id.

The downside of this approach is that a Shallow module needs to prepare the same kinds of contracts that Deep does. Same goes for Optional (unsound) modules.

I wonder if this could work differently. Maybe all Shallow exports can be plain identifiers with some syntax-local-value info ... or something ... that Deep code can look for. Then if a Deep module sees one of these ids, it uses the `require/typed` method to add a contract.

bennn:
Ah, okay.

Right now, Deep doesn't know where imports come from and treats them all uniformly.

Deep should know because those contracts are Deep's business, to protect itself. The current implementation is arguably backwards: a Shallow module has to prepare Deep contracts, which means that it will error when imported by a Deep client if there's a problem making the contracts.

Then again, the error only happens during a `require`. And the current implementation has the (major?) benefit that it creates one copy of those Deep contracts instead of one per Deep client. I'm leaning toward keeping things as they are.

mfelleisen:
OK, I haven't thought this thru completely, but based on the diss, every semantics of Types in TR can protect itself w/o knowing what other semantics exists, other than that they mess up the global invariants. If this is correct, I agree that changing it would be 'the right thing'.  [[ If I am wrong, we need to work this out. ]] 

I do wonder whether this generalizes to a spectrum of grad type sem.

bennn:
@samth @capfredf (and anyone else!) the tests are passing & I don't have any other changes planned. Please look this over & send comments and questions.

Major highlights:
- In `rep/type-rep.rkt` , the `Arrow` rep has a boolean flag to let Shallow TR know that it does not need to check the result of calls to this arrow. Because of the new flag, there are lots of changes to base types. I used a syntax parameter to change the default from "no" to "yes" in some files to lower the number of edits, but there are still a lot.
- Another big diff is in `base-env/prims-contract.rkt` . It makes weaker versions of `cast`, `require/typed`, etc. for Shallow and Optional TR.
- The biggest new block of code is `private/shallow-rewrite.rkt`. Shallow TR uses this to add runtime checks in well-typed code. (If you know Retic. Python / Transient, it's a Transient pass.)
- Most other changes use the new parameter `current-type-enforcement-mode` to decide how to compile based on the `#lang` at hand. In normal (Deep) TR, we do same thing as before. In Shallow TR and Optional TR, we do something weaker. For example, the optimizer does fewer optimizations on Shallow code and doesn't run at all on Optional code.

bennn:
ping @samth  @capfredf , time to merge?

ps Jack, resyntax seems to work fine on this PR!

capfredf:
Sorry for the delay. Reviewing this pr is on my to-do list, but somehow I forgot it. I will take a look asap this week

typed-racket-lib/typed-racket/utils/tc-utils.rkt
```diff
@@ -280,8 +288,28 @@ don't depend on any other portion of the system
           (current-continuation-marks))))
 
 ;; are we currently expanding in a typed module (or top-level form)?
+;; (box/c (or/c #t #f))
 (define typed-context? (box #f))
 
+;; (box/c type-enforcement-mode?)
+(define type-enforcement-mode (box #f))
```
capfredf, line 295: It looks to me that `type-enforcement-mode` should be a parameter.

typed-racket-lib/typed-racket/private/shallow-rewrite.rkt
```diff
@@ -0,0 +1,744 @@
...
+          [(list) (outer (remove-poly-layer tvarss-list))]
+          [(cons (list (list tvars ...) dotted) rest-tvarss)
+           (extend-indexes dotted
+              (extend-tvars/new tvars ns
+                (inner rest-tvarss)))]
+          [(cons tvars rest-tvarss)
+           (extend-tvars/new tvars ns
+             (inner rest-tvarss))])))))
+
+(define (readd-props! new-stx old-stx)
```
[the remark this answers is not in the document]
bennn: It's readd as in re-add, not read. `copy-props!` seems like a better name, I'll change it.

typed-racket-lib/typed-racket/private/shallow-rewrite.rkt
```diff
@@ -0,0 +1,744 @@
...
+(define (type->flat-contract* t* ctc-cache)
+  (for/fold ((extra-def* '())
+             (ctc-stx* '())
+             #:result (values (reverse extra-def*) (reverse ctc-stx*)))
+            ((t (in-list t*)))
+    (define-values [ex* ctc-stx] (type->flat-contract t ctc-cache))
+    (values (rev-append ex* extra-def*) (cons ctc-stx ctc-stx*))))
+
+(define (rev-append a* b*)
+  (let loop ((a* a*) (b* b*))
```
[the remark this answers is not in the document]
bennn: I haven't tested performance. Will try a quick test.

typed-racket-lib/typed-racket/utils/tc-utils.rkt
bennn: Should `typed-context?` be a parameter too? Can it be?
Unless we know it's a yes to both, I'd rather leave this to do after merging.

bennn:
Thanks @capfredf ! I left a few clarifying comments. For everything else, the response is basically: yes and I'll make a change.

(no-check is not exactly the same behavior as Optional ... definitely for multi-module programs and possibly for single modules too)

typed-racket-lib/typed-racket/utils/tc-utils.rkt
bennn: [`language-info.rkt`](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/language-info.rkt) seems like a no-go for parameters

typed-racket-doc/typed-racket/scribblings/guide/typed-untyped-interaction.scrbl
```diff
@@ -189,7 +236,121 @@ have a non-trivial performance impact, especially with the use of first-class
...
-values from another typed module.
+values from another Deep-typed module.
+
+
+@subsection{Shallow Types: Sound Types, Low-Cost Interactions}
+
+Changing the module language of the @racket[_client] program
+from @racketmodname[typed/racket] to @racketmodname[typed/racket/shallow]
+changes the way in which typed-untyped interactions are protected.
+Instead of contracts, Typed Racket uses shape checks to enforce
```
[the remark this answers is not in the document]
bennn: Added `@deftech`

typed-racket-lib/typed-racket/base-env/base-env-numeric.rkt
```diff
@@ -781,27 +782,27 @@
   (-> N B : (-PS (-is-type 0 (Un -RealZeroNoNan -InexactComplex -InexactImaginary))
                  (-not-type 0 -RealZeroNoNan)))]
 
-[number? (make-pred-ty N)]
+[number? (make-pred-ty N #t)]
```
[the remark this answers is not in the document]
bennn: Hang on, I think we do want edits to appear on all these lines. The `#t` is telling Shallow TR it can trust the result of `number?`. If that's wrong, we'd have a soundness bug.

How about a rename to make this more clear: `make-unsafe-pred-ty`? `make-unsafe-shallow-pred-ty` ?

typed-racket-doc/typed-racket/scribblings/guide/typed-untyped-interaction.scrbl
```diff
@@ -14,7 +17,21 @@ that are entirely typed. One of the key features of Typed Racket is that
...
+straightforward.
+Typed code must declare types for its untyped imports to let the type checker
+validate their use (@secref{untyped-in-typed}).
+Untyped code can freely import bindings from typed code (@secref{typed-in-untyped}).
+
+At run-time, combining typed and untyped code is complicated because there is a
+tradeoff between strong type guarantees and the performance cost of checking
+that untyped code matches the types.
+Typed Racket provides strong @emph{Deep} type guarantees by default, but offers two
+weaker options as well: Shallow and Optional types
```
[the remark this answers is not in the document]
bennn: I added deftechs (to the reference) but used emph for these, because I don't want to link the reference page yet --- not until the margin note.

typed-racket-lib/typed-racket/private/shallow-rewrite.rkt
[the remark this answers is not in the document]
bennn: `rev-append` is definitely faster than `(append (reverse a*) b*)`. Here're average times from 3 compiles on a few of the [`tr-performance`](https://github.com/pnwamk/tr-performance) programs:

| benchmark   | rev-append        | append reverse    | speedup (append - rev)   |
|-------------|-------------------|-------------------|--------|
| parser      |  3.46s (σ 0.05)   |  3.68s (σ 0.03)   | -0.22s |
| old-metrics |  4.13s (σ 0.01)   |  4.45s (σ 0.17)   | -0.32s |
| math-flonum |  7.3s (σ 0.05)    |  8.89s (σ 1.0)    | -1.59s |
| fsm         |  7.67s (σ 0.05)   |  8.2s (σ 0.34)    | -0.53s |
| forth       |  7.85s (σ 0.03)   |  9.03s (σ 0.26)   | -1.18s |
| dungeon     |  14.14s (σ 0.48)  |  16.56s (σ 0.37)  | -2.42s |
| bernoulli   |  104.28s (σ 1.24) |  105.91s (σ 1.51) | -1.63s |
| acquire     |  27.45s (σ 0.19)  |  27.19s (σ 0.27)  | +0.26s |

I'll move rev-append to utils/utils

typed-racket-lib/typed-racket/base-env/base-env-numeric.rkt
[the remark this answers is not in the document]
capfredf: `make-unsafe-shallow-pred-ty` looks good to me

typed-racket-lib/typed-racket/private/shallow-rewrite.rkt
```diff
@@ -736,9 +737,5 @@
              #:result (values (reverse extra-def*) (reverse ctc-stx*)))
             ((t (in-list t*)))
     (define-values [ex* ctc-stx] (type->flat-contract t ctc-cache))
-    (values (rev-append ex* extra-def*) (cons ctc-stx ctc-stx*))))
-
-(define (rev-append a* b*)
-  (let loop ((a* a*) (b* b*))
-    (if (null? a*) b* (loop (cdr a*) (cons (car a*) b*)))))
+    (values (append (reverse ex*) extra-def*) (cons ctc-stx ctc-stx*))))
```
shhyou, line 740: Not using `rev-append` anymore?

bennn: 😱

capfredf (approved):

samth (approved):

bennn:
Awesome. I'll add the alias `make-unsafe-shallow-pred-ty` and merge once the tests pass.

bennn:
So happy I'll never need to rebase this PR atop master again.

sorawee:
Yay!

yfzhe:
I found that both [kind system rfc](https://github.com/racket/typed-racket/blob/1dde78d165472d67ae682b68622d2b7ee3e15e1e/rfcs/text/0003-kind-system.md) and this [transient semantics rfc](https://github.com/racket/typed-racket/blob/1dde78d165472d67ae682b68622d2b7ee3e15e1e/rfcs/text/0003-shallow-optional-semantics.md) use the same 003 ordinal number. Should the transient semantics one be changed to 004?

capfredf:
@benn seems like `racket -l typed-racket-test -- --unit` don't run those failed tests found by DrDr

bennn:
Right, those files aren't supposed to run. The harness compiles them & looks at the expanded code. (Except for `type-annotation.rkt`,  which is unsupported.)

I tried adding compile-omit-paths. Also fixed the rfc number: https://github.com/racket/typed-racket/pull/1273

PR #950, opened 2020-07-06
magnitude of Float-Complex where one (or both) parts are inf

bdeket:
Bring in line with main racket:
https://docs.racket-lang.org/reference/generic-numbers.html?q=magnitude#%28def._%28%28quote._~23~25kernel%29._magnitude%29%29
stating that inf takes precedence.


samth:
Can you add a test? Probably just use rackunit in a file in typed-racket-test/succeed.

PR #951, opened 2020-07-18
fix: Add server protocol to ssl-listen

GriffinMB:
The optional server protocol parameter type is missing from ssl-listen, as seen here: https://docs.racket-lang.org/openssl/index.html?q=typed%2Fopenssl#%28def._%28%28lib._openssl%2Fmain..rkt%29._ssl-listen%29%29

This updates the type declaration.


GriffinMB:
Is there anything else you need from me here?

samth:
Thanks!

PR #952, opened 2020-07-28
rfc for #948

bennn:
First draft

[rendered](https://github.com/bennn/typed-racket/blob/rfc-transient/rfcs/text/0003-shallow-optional-semantics.md)


bennn:
See #948

PR #953, opened 2020-07-29
type for the optional arguments of 'in-fx/fl/extflvector'

bdeket:
add type for the optional arguments of 'in-fx/fl/extflvector' as in 'in-vector'

### DIFF
--- a/typed-racket-test/succeed/pr953-in-xvector.rkt
+++ b/typed-racket-test/succeed/pr953-in-xvector.rkt
@@ -10,13 +10,16 @@
 (check-equal? (sequence->list (in-flvector (flvector 1.0 2.0 3.0 4.0) 1)) (list 2.0 3.0 4.0))
 (check-equal? (sequence->list (in-flvector (flvector 1.0 2.0 3.0 4.0) 1 2)) (list 2.0))
 (check-equal? (sequence->list (in-flvector (flvector 1.0 2.0 3.0 4.0) 0 #f 2)) (list 1.0 3.0))
-(check-equal? (for/list : (Listof Flonum) ([i (in-flvector (flvector 1.0 2.0 3.0 4.0) 3 -1 -2)])i)
+(check-equal? (for/list : (Listof Flonum) ([i (in-flvector (flvector 1.0 2.0 3.0 4.0) 3 -1 -2)]) i)
               (list 4.0 2.0))
 
 (check-equal? (sequence->list (in-fxvector (fxvector 1 2 3 4) 0 #f 2)) (list 1 3))
-(check-equal? (for/list : (Listof Fixnum) ([i (in-fxvector (fxvector 1 2 3 4) 3 -1 -2)]) i)
-              (list 4.0 2.0))
+(check-equal? (for/list : (Listof Fixnum) ([i : Fixnum (ann (in-fxvector (fxvector 1 2 3 4) 3 -1 -2) (Sequenceof Fixnum))]) i)
+              (list 4 2))
+;; I don't understand why this fails...
+#;(check-equal? (for/list : (Listof Fixnum) ([i (in-fxvector (fxvector 1 2 3 4) 3 -1 -2)]) i)
+              (list 4 2))
 
 (when (extflonum-available?)
-  (check-equal? (for/list : (Listof extflonum) ([i (in-extflvector (extflvector 1.0t0 2.0t0 3.0t0 4.0t0) 3 -1 -2)]) i)
+  (check-equal? (for/list : (Listof ExtFlonum) ([i (in-extflvector (extflvector 1.0t0 2.0t0 3.0t0 4.0t0) 3 -1 -2)]) i)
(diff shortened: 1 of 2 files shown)


bdeket:
What would be a good test for this? Or how how to test this?

I see 
```
[tc-e (for/list : (Listof Flonum) ([a (in-flvector (flvector 1.0 2.0))]) a)
             (-lst -Flonum)]
```
in the unit-test/typecheck-tests.rkt file. Best to add an extra test over there?

sorawee:
Since you are already adding a new test file, putting it there also makes sense. But really, I don't care where it's tested (though other people might, idk).

typed-racket-test/succeed/pr953-in-xvector.rkt
```diff
@@ -0,0 +1,21 @@
...
+(check-equal? (sequence->list (in-flvector (flvector 1.0 2.0 3.0 4.0) 1 2)) (list 2.0))
+(check-equal? (sequence->list (in-flvector (flvector 1.0 2.0 3.0 4.0) 0 #f 2)) (list 1.0 3.0))
+(check-equal? (for/list : (Listof Flonum) ([i (in-flvector (flvector 1.0 2.0 3.0 4.0) 3 -1 -2)])i)
+              (list 4.0 2.0))
+
+(check-equal? (sequence->list (in-fxvector (fxvector 1 2 3 4) 0 #f 2)) (list 1 3))
+(check-equal? (for/list : (Listof Fixnum) ([i (in-fxvector (fxvector 1 2 3 4) 3 -1 -2)]) i)
+              (list 4.0 2.0))
+
+(check-equal? (for/list : (Listof extflonum) ([i (in-extflvector (extflvector 1.0t0 2.0t0 3.0t0 4.0t0) 3 -1 -2)]) i)
```
samth, line 20: This test should be behind a `(when (extflonum-available?) ...)` check.

bdeket:
I see this fails for the cs branch, and looking into it, I don't understand why it succeeds on the bc branch.
the for expands to [normalise-inputs](https://github.com/racket/typed-racket/blob/0ed63a7e2fa6b3cb2ee6934717feb159384b129e/typed-racket-lib/typed-racket/base-env/base-special-env.rkt#L76),  where the start/stop/step-indexes can be Nat/Nat/Nat.

I don't know if typed-racket does something extra but if it is just using the [racket implementation](https://github.com/racket/racket/blob/6b1dad21ebd103960822e6de7d45e9a85303982b/racket/collects/racket/private/for.rkt#L897) then I think the types for the indexes should be:
Nat/(U Nat -1)/(U Positive-Integer Negative-Integer)

I can change the test, to only count upwards in a for loop,
or look into changing the type for normalise-inputs...
And maybe also change the type for the step from Integer to (U Positive-Integer Negative-Integer) for in-vector and friends

bdeket:
@samth, I went ahead and changed the type for normalise-inputs to allow going backward through the vector.

samth:
Fantastic! I'll take a look and hopefully we can merge soon.

typed-racket-test/succeed/pr953-in-xvector.rkt
```diff
@@ -0,0 +1,25 @@
...
+(check-equal? (sequence->list (in-flvector (flvector 1.0 2.0 3.0 4.0) 1)) (list 2.0 3.0 4.0))
+(check-equal? (sequence->list (in-flvector (flvector 1.0 2.0 3.0 4.0) 1 2)) (list 2.0))
+(check-equal? (sequence->list (in-flvector (flvector 1.0 2.0 3.0 4.0) 0 #f 2)) (list 1.0 3.0))
+(check-equal? (for/list : (Listof Flonum) ([i (in-flvector (flvector 1.0 2.0 3.0 4.0) 3 -1 -2)]) i)
+              (list 4.0 2.0))
+
+(check-equal? (sequence->list (in-fxvector (fxvector 1 2 3 4) 0 #f 2)) (list 1 3))
+(check-equal? (for/list : (Listof Fixnum) ([i : Fixnum (ann (in-fxvector (fxvector 1 2 3 4) 3 -1 -2) (Sequenceof Fixnum))]) i)
+              (list 4 2))
+;; I don't understand why this fails...
```
samth, line 19: Can you say more about this failure?

bdeket: I was hoping to be able to do:
`(for/list : (Listof Fixnum) ([i (in-fxvector (fxvector 1 2 3 4) 3 -1 -2)]) i)`
but without the annotation of `(Sequenceof Fixnum)` and  `Fixnum` for `i` (both are necessary) The resulting list is seen as `(Listof Integer)`

capfredf: I took a quick look at this failure.  Seems it has something to do with unsafe-normalise-inputs and normalise-inputs. I might get back to some point later next week

samth:
I'm going to merge this now; we can try to make that example work with less annotation in the future.

PR #954, opened 2020-07-29
Bring type in line with documentation/implementation

bdeket:
Sending real values for x and y to bitmap get-argb-pixels exits with  an error
(for example)
```
(send H get-argb-pixels 1. 1. 2 2 (make-bytes 100))
ptr-ref: contract violation
  expected: fixnum?
  given: 2055.0
  argument position: 3rd
  other arguments...:
```
Note that get/set are not mirrored (set does allow real x and y according to the doc)

looking at racket/draw/bitmap it seems that x and y could be Integer (ie neg. values allowed) but since the docs say otherwise I will stick with this for now

### DIFF
--- a/typed-racket-more/typed/racket/private/gui-types.rkt
+++ b/typed-racket-more/typed/racket/private/gui-types.rkt
@@ -65,7 +65,7 @@
                        Real)
                  (List Bytes Integer Integer)))
    [get-argb-pixels
-    (->* [Real Real Exact-Nonnegative-Integer Exact-Nonnegative-Integer Bytes]
+    (->* [Exact-Nonnegative-Integer Exact-Nonnegative-Integer Exact-Nonnegative-Integer Exact-Nonnegative-Integer Bytes]
          [Any Any #:unscaled? Any]
          Void)]
    [get-backing-scale (-> Positive-Real)]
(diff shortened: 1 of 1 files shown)


typed-racket-more/typed/racket/private/gui-types.rkt
```diff
@@ -481,7 +481,7 @@
   (Class #:implements DC<%>
          (init [bitmap (Option (Instance Bitmap%))])
          [get-argb-pixels
-          (->* [Real Real Exact-Nonnegative-Integer Exact-Nonnegative-Integer Bytes]
+          (->* [Exact-Nonnegative-Integer Exact-Nonnegative-Integer Exact-Nonnegative-Integer Bytes]
```
samth, line 484: Did this intentionally remove an argument?

bdeket:
no, I removed the argument by mistake :(

The function works fine with negative integers, so I could use Integer instead of Exact-Nonnegative-Integer?
I tested some more rationals and floats, and... they sometimes work, depending on the combination: 
- if both x and y are smaller than 0 they can be any Real
- if one of them is really big (for my 512x512 image , bigger than 1e15) both can be any Real
but for all 'normal' combinations having one value not an integer triggers a bytes-set!, or ptr-ref contract violation

In the end it turned out that my real problem is that the type for Bitmap% in racket/draw and images/flomap don't match. And this is PR is not solving this...

PR #962, opened 2020-08-20
fix missing prompt-tag for call/cc

### LINKED ISSUE #961: call/cc should accept prompt tag
7.8
### What program did you run?
```
> call/cc
- : (All (a b c ...)
      (case->
       (-> (-> (-> Nothing) (values)) (values))
       (-> (-> (-> a c ... c Nothing) (values b c ... c))
           (values (U a b) c ... c))))
#<procedure:call-with-current-continuation>
```

It should accept prompt tag as the second argument, but absent from its type

yjqww6:
fix #961


capfredf:
@samth this looks good to merge

PR #968, opened 2020-09-01
fix duplicated evaluation of vector-expr after optimizing known-length vector-op

yjqww6:
```
#lang typed/racket

(define x 0)
(vector-ref (vector (set! x (+ x 1))) 0)
x
```
This program yields 2 currently.

### DIFF
--- a/typed-racket-lib/typed-racket/optimizer/vector.rkt
+++ b/typed-racket-lib/typed-racket/optimizer/vector.rkt
@@ -79,7 +79,7 @@
                          #`(when (immutable? new-v)
                              (op new-v i.opt new.opt ...)) ; produces the correct error message
                          #'(begin))
-                   (op.unsafe v.opt i.opt new.opt ...)))
+                   (op.unsafe new-v i.opt new.opt ...)))
 
   ;; we can do the bounds checking separately, to eliminate some of the checks
   (pattern (#%plain-app op:vector-op v:opt-expr i:fixnum-expr new:opt-expr ...)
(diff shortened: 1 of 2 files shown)


yjqww6:
`nonrecoverable invalid memory reference`
This error message from the failed cs ci seems like a segfault during gc, which I think is not very likely to relate to this pull request.
Probably, this expose a severe bug in racketcs.

yjqww6:
Ran ci on a branch even with master and got the same problem, so recent changes to RacketCS might be problematic.
https://github.com/yjqww6/typed-racket/runs/1060799964

samth:
cc @mflatt

mflatt:
I haven't been able to replicate the problem, so far. Is `xvfb-run racket -l typed-racket-test -- --int` the right thing to run?

yjqww6:
If it did related to racketcs changes, it should happen between Aug 28 and Sept 2. But probably we should re-run the ci to see what happen since there are many changes to racket and typed-racket last week.
In worst case, it may need to use exactly the same docker image and scripts with ci config to setup such an environment.

samth:
That is indeed the command that failed. Did you try the snapshot that would have been current on the 7th?

yjqww6:
I cannot reproduce on my machine using a snapshot from ci-snapshot.racket-lang.org either.
But it could be replicated in ci consistently,
1. https://github.com/yjqww6/typed-racket/actions/runs/247346705 - even with master, fail with same problem
2. https://github.com/yjqww6/typed-racket/actions/runs/247673432 - 1 commit behind master(to avoid failing earlier), but change ci to use racketcs 7.8,  doesn't have this problem.
3. https://github.com/yjqww6/typed-racket/actions/runs/247711415 - even with master, but change image to ubuntu 18.04, fail with same problem

yjqww6:
I have created an issue https://github.com/racket/racket/issues/3397 to let this PR serves for its original purpose.

samth:
I will run the tests manually so we can merge this.

samth:
All the tests passed; I merged.

PR #972, opened 2020-09-17
Generalize Byte to something more specific.

samth:
cc @sorawee

### DIFF
--- a/typed-racket-lib/typed-racket/types/generalize.rkt
+++ b/typed-racket-lib/typed-racket/types/generalize.rkt
@@ -18,6 +18,7 @@
         [(== -Zero) -Int]
         [(List: ts) (-lst (apply Un ts))]
         [(? (lambda (t) (subtype t -Bottom))) Univ]
+        [(? (lambda (t) (subtype t -Byte))) -NonNegFixnum]
         [(? (lambda (t) (subtype t -Int))) -Int]
         [(? (lambda (t) (subtype t -Rat))) -Rat]
         [(? (lambda (t) (subtype t -Flonum))) -Flonum]


samth:
This fixes the particular problem with racket/racket#3396 but it's not clear if (a) it breaks something else or (b) it fixes everything (what happens if we annotate with `Index`?).

sorawee:
Yes, `Index` has the same issue

```
Type Checker: type mismatch
  expected: Index
  given: Positive-Fixnum
  in: (for/sum: : Integer ((i : Index (in-range 4))) i)
```

yjqww6:
Though I didn't actually try this, I guess it breaks code like
```
#lang typed/racket

(define (f [y : Integer]) : Integer
  (let loop ([a 1])
    (if (< a y)
        (loop (* a 2))
        a)))
```
, which seems not ideal.

Probably this is a chance to remove more gaps from typechecking inlined for-clauses.

A possible way is change https://github.com/sorawee/racket/blob/1766cba6c66ecd405e5a1803d1ad0d1813e5d786/racket/collects/racket/private/for.rkt#L2099 from `[pos start]` to something like `[pos (with-typed-racket:range-hint start (start end inc))]`, which attachs syntax property eventually. And then typed-racket could generalize it according to hints.

Moreover, probably we could expect even `(for/sum ([i : Positive-Byte (in-range 10 0 -1)]) i)` type checks (with a more precise unsafe-fx+). The expanded code should be similar to following program:
```
#lang typed/racket
(require typed/racket/unsafe racket/unsafe/ops)
(unsafe-require/typed
 racket/unsafe/ops
 [unsafe-fx+ (-> (U Negative-Fixnum Byte) Nonpositive-Fixnum
                 (U Negative-Fixnum Byte))])

(let-values ([(start) (quote 10)] [(end) (quote 0)] [(inc) (quote -1)])
  (#%app
   (letrec-values (((for-loop)
                    (lambda (acc [pos : (U Negative-Fixnum Byte)])
                      (if (#%app unsafe-fx> pos end)
                          (let-values ([([i : Positive-Byte]) pos])
                            (if (quote #t)
                                (let-values ([(acc)
                                              (let-values (((acc) acc))
                                                (let-values ([(acc)
                                                              (let-values ()
                                                                (let-values (((new)
                                                                              (let-values ()
                                                                                i)))
                                                                  (#%app
                                                                   +
                                                                   acc
                                                                   new)))])
                                                  (#%app values acc)))])
                                  (if (if (quote #t)
                                          (#%app not (quote #f))
                                          (quote #f))
                                      (#%app
                                       for-loop
                                       acc
                                       (#%app unsafe-fx+ pos inc))
                                      acc))
                                acc))
                          acc))))
     for-loop)
   (quote 0)
   start))
```

PR #981, opened 2020-10-09
rename Exist to Some

capfredf:
The other day, the original name suddenly looked off to me....
I should have chosen Some in the first place ....


capfredf:
@samth this can be merged too

PR #982, opened 2020-10-09
no user-input Imp

capfredf:


yjqww6:
It seems you are in trouble with a recent racketcs problem https://github.com/racket/racket/issues/3397 .

capfredf:
@yjqww6 Thank you for the information. Got "nonrecoverable invalid memory reference" on Racket/CS 7.8.0.11.  On Racket/CS 7.9.0.2, the integrated tests simply hangs, but running the tests without `xvfb-run` works fine. 

I will try your minimum command and see what happens.

capfredf:
@samth tests passed as well.

PR #988, opened 2020-11-03
remove an expression for debugging

### LINKED ISSUE #987: Bug when use existential functions without arguments.
7.9 cs
### What program did you run?
```
#lang typed/racket

(: a (-> (Some (a) (-> a)) Void))

(define (a sizeable)
  (sizeable)
  (void))
```
type checks.
```
racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt:85:7: match-define: no matching clause for '()
```
---
BTW: With an argument, it works.
```
#lang typed/racket

(: a (-> (Some (a) (-> Any a)) Void))

(define (a sizeable)
  (sizeable 'a)
  (void))
```

capfredf:
fix #987


capfredf:
Oh my stupid bug......
I will add more unittests as sanity checks

PR #993, opened 2020-11-09
remove type anns for parameters in λs generated from kw/opt plambdas 

### LINKED ISSUE #940: #:forall and optional keyword arguments
Welcome to DrRacket, version 7.7.0.7 [3m].

### What program did you run?
```
#lang typed/racket/base

(struct (A) foo ([a : (Listof A)]))


(define #:forall (A) (bar [a : (Listof A)]
                          #:test1 [b : (foo A)(foo '())]
                          #:test2 [c : String]) : (foo A)
  (foo a))
```

type-check
```
. Type Checker: parse error in type;
 type name `A' is unbound in: A
. Type Checker: type mismatch
  expected: (Listof (Pairof Struct-Type-Property Any))
  given: (List (List* Struct-Type-Property 'bar (-> Any (Listof Error) Any))) in: (#%module-begin (define-values (bar) (-lambda #:forall (A) (a #:test1 (b (foo (quote ()))) #:test2 c) (ann (foo a) (foo A)))))
. Type Checker: Summary: 2 errors encountered in:
  A
  (#%module-begin (define-values (bar) (-lambda #:forall (A) (a #:test1 (b (foo (quote ()))) #:test2 c) (ann (foo a) (foo A)))))
```
------------------
using (: bar (All (A) (->* ... etc works

capfredf:
Optional or keyword lambdas is expanded to more lambdas where
syntax objects for the original paramaters are reused. This is problematic when
those parameters are typed with bound type variables in polymorphic functions,
because in generated lambdas those variables would become unbound.

This patch fixes the problem by throwing the types away before typechecking
the generated lambdas.

closes #940

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims.rkt
@@ -684,7 +684,22 @@ the typed racket language.
        (if (attribute return.type)
            #`(ann last-e #,(attribute return.type))
            #'last-e))
-     (define d (syntax/loc stx (λ formals.erased e ... last-e*)))
+
+     ;; if the lambda form being checked is a polymorphic function, tag its
+     ;; parameters with property `from-lambda'. 
+     (define (maybe-set-from-lambda type-vars formals)
+       (cond
+         [(and type-vars (stx-list? formals))
+          (stx-map (lambda (x)
+                     (from-plambda-property x #t))
+                   formals)]
+         [type-vars (from-plambda-property formals #t)]
+         [else formals]))
+
+     (define d (with-syntax ([erase-formals (maybe-set-from-lambda
+                                             (attribute vars.type-vars)
+                                             (attribute formals.erased))])
+                 (syntax/loc stx (λ erase-formals e ... last-e*))))
      (define d/prop
        (if (attribute formals.kw-property)
            (kw-lambda-property d (attribute formals.kw-property))
--- a/typed-racket-lib/typed-racket/private/syntax-properties.rkt
+++ b/typed-racket-lib/typed-racket/private/syntax-properties.rkt
@@ -69,6 +69,8 @@
   (struct-info struct-info)
   (opt-lambda opt-lambda)
   (kw-lambda kw-lambda)
+  ; mark syntax objects for formal parameters in polymorphic lambdas forms.
+  (from-plambda from-plambda)
   (tail-position typechecker:called-in-tail-position #:mark)
   (tr:class tr:class #:mark)
   (tr:class:top-level tr:class:top-level)
--- a/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
@@ -43,14 +43,28 @@
 ;; (i.e. (append positional (or id '())) but syntax)
 (struct formals (positional rest syntax) #:transparent)
 
-(define (make-formals stx)
+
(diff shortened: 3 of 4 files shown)


typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -43,14 +43,18 @@
 ;; (i.e. (append positional (or id '())) but syntax)
 (struct formals (positional rest syntax) #:transparent)
 
-(define (make-formals stx)
+(define (make-formals stx [not-in-poly #t])
+  (define (maybe-remove a)
+    (if (and not-in-poly (from-plambda-property a))
+        (syntax-property-remove a 'type-label) a))
```
samth, line 49:
```suggestion
        (syntax-property-remove a 'type-label)
        a))
```

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -43,14 +43,18 @@
 ;; (i.e. (append positional (or id '())) but syntax)
 (struct formals (positional rest syntax) #:transparent)
 
-(define (make-formals stx)
+(define (make-formals stx [not-in-poly #t])
```
samth, line 46: This needs a comment explaining what's going on with `not-in-poly`

typed-racket-lib/typed-racket/base-env/prims.rkt
```diff
@@ -684,7 +684,18 @@ the typed racket language.
        (if (attribute return.type)
            #`(ann last-e #,(attribute return.type))
            #'last-e))
-     (define d (syntax/loc stx (λ formals.erased e ... last-e*)))
+
+     (define (maybe-set-original-poly type-vars formals)
```
samth, line 688: this function needs a comment explaining what it does

capfredf:
ping @samth

typed-racket-lib/typed-racket/private/syntax-properties.rkt
```diff
@@ -69,6 +69,8 @@
   (struct-info struct-info)
   (opt-lambda opt-lambda)
   (kw-lambda kw-lambda)
+  ; marks syntax objects in formal parameters in polymorhpic lambdas forms.
```
samth, line 72:
```suggestion
  ; marks syntax objects in formal parameters in polymorphic lambdas forms.
```

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -43,14 +43,23 @@
 ;; (i.e. (append positional (or id '())) but syntax)
 (struct formals (positional rest syntax) #:transparent)
 
-(define (make-formals stx)
+;; When not-in-poly is true, make-formals is invoked in tc/mono-lambda/type for
+;; checking non-polymorhic functions. In some scenarios, those functions are
```
samth, line 47: spelling here too.

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
samth: I think what's going on here is that we want to remove some information that might have out-of-scope type variables in particular cases. But the comment should explain why.

samth (changes requested):

capfredf:
The third CI run failed due to the flaky random testing

capfredf:
ping @samth

PR #996, opened 2020-11-13
put bindings of structs' extra constructors in deftbl in the main type-check

### LINKED ISSUE #601: unsoundness with 'define-struct' in TR
6.10.1.2--2017-09-23(-/f) [3m].

### What program did you run?

```racket
#lang racket

(module class-helpers typed/racket

  (provide look-in-struct
           (struct-out my-struct))

  ;; PROBLEM SPECIFIC TO 'define-struct':
  (define-struct my-struct ([a : Natural]))

  (: look-in-struct (my-struct -> Natural))
  (define (look-in-struct s)
    (my-struct-a s))

  )

(require 'class-helpers)

;; SHOULD FAIL: not a natural...
(define info (my-struct "not a natural..."))

;; RETURNS STRING, NOT NATURAL
(look-in-struct info)
```

a runtime error indicating that you can't create a "my-struct" with a string.

Instead, the structure is cheerfully created, and the function that's supposed to return a Natural actually returns a string. You can easily use this to cause a seg fault in racket; I have that code, too.

Note that this problem is specific to `define-struct`. Replacing it with `struct` solves the problem, and produces a program that signals the runtime error as expected.

Nope, it just does the unsound thing.

capfredf:
closes #601

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
@@ -364,27 +364,33 @@
 
   (define extra-constructor (struct-names-extra-constructor names))
 
+  (define constructor-type (poly-wrapper (->* all-fields poly-base)))
+  
   (define constructor-binding
     (make-def-binding (struct-names-constructor names)
-                      (poly-wrapper (->* all-fields poly-base))))
-  (define constructor-bindings
-    (cons constructor-binding
-          (if extra-constructor
-              (list (make-def-binding extra-constructor
-                                      (poly-wrapper (->* all-fields poly-base))))
-              null)))
+                      constructor-type))
+
+  (define-values (constructor-bindings bindings^)
+    (let ([extra-constructor-bindings
+           (if extra-constructor (list (make-def-binding extra-constructor
+                                                         constructor-type))
+               null)])
+      (values (cons constructor-binding extra-constructor-bindings)
+              (append (cons (make-def-struct-stx-binding (struct-names-type-name names)
+                                                         si
+                                                         (def-binding-ty constructor-binding))
+                            extra-constructor-bindings)
+                      bindings))))
 
   (for ([b (in-list (append constructor-bindings bindings))])
     (register-type (binding-name b) (def-binding-ty b)))
-
+    
   (append
-    (if (free-identifier=? (struct-names-type-name names)
-                           (struct-names-constructor names))
+   (if (free-identifier=? (struct-names-type-name names)
+                          (struct-names-constructor names))
       null
       (list constructor-binding))
-   (cons
-     (make-def-struct-stx-binding (struct-names-type-name names) si (def-binding-ty constructor-binding))
-     bindings)))
+   bindings^))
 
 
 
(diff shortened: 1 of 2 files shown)


typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -367,24 +367,32 @@
   (define constructor-binding
     (make-def-binding (struct-names-constructor names)
                       (poly-wrapper (->* all-fields poly-base))))
+
+  (define extra-constructor-binding
+    (if extra-constructor
```
samth, line 372: maybe use `and` here

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -367,24 +367,32 @@
...
+        (make-def-binding extra-constructor
+                          (poly-wrapper (->* all-fields poly-base)))
```
samth, line 374: Can we define this type just once and use it also in `constructor-binding`?

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -367,24 +367,32 @@
...
+   (if (free-identifier=? (struct-names-type-name names)
+                          (struct-names-constructor names))
       null
       (list constructor-binding))
    (cons
-     (make-def-struct-stx-binding (struct-names-type-name names) si (def-binding-ty constructor-binding))
-     bindings)))
+    (make-def-struct-stx-binding (struct-names-type-name names) si (def-binding-ty constructor-binding))
+    (if extra-constructor
+        (cons (make-def-struct-stx-binding extra-constructor si (def-binding-ty extra-constructor-binding))
```
samth, line 393: we could also use that type here instead of extracting it from the `def-binding`.

samth (approved):

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
capfredf: Yes. Also I don't think using `(make-def-struct-stx-binding extra-constructor ...)` makes sense here, because the field of def-struct-stx-binding is the name of the structure.  So I rewrote my changes

samth (approved):

capfredf:
Ping @samth

capfredf:
Ping @samth

typed-racket-lib/typed-racket/typecheck/tc-structs.rkt
```diff
@@ -364,27 +364,33 @@
...
+
+  (define-values (constructor-bindings bindings^)
+    (let ([extra-constructor-bindings
+           (if extra-constructor (list (make-def-binding extra-constructor
+                                                         constructor-type))
+               null)])
+      (values (cons constructor-binding extra-constructor-bindings)
+              (append (cons (make-def-struct-stx-binding (struct-names-type-name names)
+                                                         si
+                                                         (def-binding-ty constructor-binding))
```
samth, line 381:
```suggestion
                                                         constructor-type)
```

samth (approved):

capfredf:
Ping @samth

PR #1001, opened 2020-11-29
Custom port

wargrey:
Now we can make robust custom input ports in typed racket directly.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -1346,6 +1346,8 @@
 [exn:srclocs? (-> Univ B : (-has-struct-property prop:exn:srclocs))]
 [exn:srclocs-accessor (-> Univ (-lst Univ))] ;TODO
 
+[srcloc->string (-> -Srcloc -String)]
+
 ;; Section 10.3 (Delayed Evaluation)
 [promise? (make-pred-ty (-Promise Univ))]
 [force (-poly (a) (->acc (list (-Promise a)) a (list -force)))]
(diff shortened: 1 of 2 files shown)


typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -1982,20 +1982,20 @@
 
 ;; Section 13.1.9
 [make-input-port
- (let ([specials-func (-> (-opt -Integer) (-opt -Integer) (-opt -Integer) (-opt -Integer) Univ)])
+ (let ([specials-func (-> (-opt -PosInt) (-opt -Nat) (-opt -PosInt) (-opt -Nat) Univ)])
    (->opt Univ
           (Un (-> -Bytes (Un -Nat (-val eof) specials-func (-evt Univ)))
               -Input-Port)
-          (Un (-> -Bytes -Nat (-opt (-evt Univ))
+          (Un (-> -Bytes -Nat (-opt (-mu x (-evt x)))
```
capfredf, line 1989: It seems `-poly` should be used instead of `-mu`

capfredf: Also I think quantification over the whole function type better captures the behavior of the function

wargrey: Hi, @capfredf 

> (All (a)
      (case->
       (->* (Any Input-Port Input-Port (-> Any))
            ((U (-> (Evtof a)) False)
             (U (-> Positive-Integer (Evtof a) (Evtof Any) Any) False)
             (U (-> (values
                     (U Exact-Positive-Integer False)
                     (U Exact-Nonnegative-Integer False)
                     (U Exact-Positive-Integer False)))
                False)
             (-> Any)
             (U (-> (U Exact-Positive-Integer False)) False Integer Port)
             (U (case-> (-> (U block none) Any) (-> (U False block none)))
                False))
            Input-Port)
       (->* (Any
             (-> Bytes
                 (U (Rec
                     x
                     (Evtof (U EOF Exact-Nonnegative-Integer Input-Port x)))
                    EOF
                    Exact-Nonnegative-Integer))
             False
             (-> Any))
            (False
             False
             (U (-> (values
                     (U Exact-Positive-Integer False)
                     (U Exact-Nonnegative-Integer False)
                     (U Exact-Positive-Integer False)))
                False)
             (-> Any)
             (U (-> (U Exact-Positive-Integer False)) False Integer Port)
             (U (case-> (-> (U block none) Any) (-> (U False block none)))
                False))
            Input-Port)
       (->* (Any
             (-> Bytes
                 (U (Rec
                     x
                     (Evtof
                      (U (-> (U Exact-Positive-Integer False)
                             (U Exact-Nonnegative-Integer False)
                             (U Exact-Positive-Integer False)
                             (U Exact-Nonnegative-Integer False)
                             Any)
                         EOF
                         Exact-Nonnegative-Integer
                         Input-Port
                         x)))
                    Port-Reader-Datum))
             (-> Bytes
                 Nonnegative-Integer
                 (U (Evtof a) False)
                 (U (Evtof Any) False Port-Reader-Datum))
             (-> Any))
            ((U (-> (Evtof a)) False)
             (U (-> Positive-Integer (Evtof a) (Evtof Any) Any) False)
             (U (-> (values
                     (U Exact-Positive-Integer False)
                     (U Exact-Nonnegative-Integer False)
                     (U Exact-Positive-Integer False)))
                False)
             (-> Any)
             (U (-> (U Exact-Positive-Integer False)) False Integer Port)
             (U (case-> (-> (U block none) Any) (-> (U False block none)))
                False))
            Input-Port)))

does this look right?
1. I added two cases for the constraints on dependent arguments: 1) if `peek` is `#false`, the specials-func, `port-progress-evt` and `commit` are not allowed; 2) if `read-in` is an input-port, so is the `peek`.

2. I am not sure what is the actual use of `-poly`, I used the parameter on all `progress-evt`-related arguments, and loosened their types of input arguments to `(Evtof Any)`.

wargrey: I meant other APIs who require a `progress-evt` argument.

capfredf: I am not sure why we need recursive types for this function. Can you expand on that?

wargrey: No particular reasons.
If I understand it correctly, an event whose result is itself can be annotated with a recursive type, and the `progress-evt` is such a kind of the event.

Personally, I favor strict types for input arguments despite many exceptions.

Nonetheless, in this situation, neither `(Evtof Any)` nor `(Rec x (Evtof x))` are actual guarantees to ensure the correctness of the input arguments.

Besides, for all primitive APIs that require a `progress-evt` instance as their input arguments, their internal implementation is beyond the scope of typed racket. Thus, I agree that `(Evtof Any)` is a good choice as long as we keep their type annotations consistent to pass the event instance smoothly.

wargrey: For `make-input-port`, the `read-in` may return an event instance, whose result might be one of the other three possible non-event values or a recursive event like itself.  Should we just use `(Evtof Any)` here?

capfredf: Thanks for your explanation. `(Rec x (Evtof X))` is indeed a better choice for things related to `progress-evt`

wargrey: Okay, I updated this pull request with more type annotations.

`collection-path` remains the same since it has been deprecated.

PR #1008, opened 2020-12-11
fix the type of sort

capfredf:


samth:
This looks right but the new test doesn't pass.

jessealama:
I recently bumped into this false positive type checking (that is, type checking passed but it should have failed). If I make the change here to my local TR, I get the expected type check error.

dannypsnl:
I think this one failed because it raises two errors
```
Type Checker: Polymorphic function `sort' could not be applied to arguments:
...
Type Checker: Polymorphic function with keywords sort7 could not be applied to arguments:
...
```
a simple fix is changing to`(exn-pred "Type Checker: Summary: 2 errors encountered")`, but I have no idea if need to reduce to only report one error.

samth:
It would be great to fix this before the release -- @capfredf do you want me to do that?

capfredf:
Ah, I totally this pending pull request. I will be working on those failed tests later today.

PR #1016, opened 2021-01-12
fix a bug in typechecking partially type-annotated lambdas

### LINKED ISSUE #1011: fails to catch ill-typed property values that are partially annotated
7.9.0.10 on c71e6289af3f1c6e4498e3503519d95f889cdcb6

### What program did you run?
```racket
(struct root ([p : Integer] [q : Integer])
    #:property prop:custom-write 
    (λ ([me : Integer] [port : Output-Port] mode) : Void
         (fprintf port "[[~a ~a]]" (root-p me) (root-q me))))
```

A type error

-----------------
If the parameter `mode` in the lambda is correctly annotated with `(U Boolean 1 0)`, then the typechecker reports a type error expectedly

capfredf:
When using an expected arrow type to check a lambda with not every parameter
annotated, the existing annotation was thrown away, and the lambda was treated
as non-annotated one.

closes #1011

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
@@ -137,7 +137,8 @@
           (cond
             [(andmap type-annotation arg-list)
              (get-types arg-list #:default Univ)]
-            [(zero? extra-arg-count) arg-tys]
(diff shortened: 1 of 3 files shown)


capfredf:
blocked by https://github.com/racket/typed-ra1020

samth (approved):

PR #1017, opened 2021-01-12
remove travis

capfredf:


samth:
Actually, even thought I merged this it's not right. We want to
1. Have PLT_TR_CONTRACTS set in a different build than the usual one
2. Set it when compiling Typed Racket, otherwise it will have no effect.

capfredf:
That was what I thought initially: i.e, set the PLT_TR_CONTRACTS before `raco setup`. But the way the variable was set in travis was more close to the current change. 

So the correct solution seems like:
1. PLT_TR_CONTRACTS = 1 raco setup typed typed-racket typed-scheme
2. run unittests
3. raco setup typed typed-racket typed-scheme 
4. run other test suites

samth:
No, we want to do this as a matrix, the way the travis configuration did.

PR #1018, opened 2021-01-12
fix github actions

capfredf:


.github/workflows/ci.yml
```diff
@@ -24,14 +25,29 @@ jobs:
...
              sudo raco pkg update -i --auto --no-setup source-syntax/ typed-racket-lib/ typed-racket-more/ typed-racket-compatibility/ typed-racket-doc/ typed-racket/ typed-racket-test/
-             sudo raco setup --check-pkg-deps typed typed-racket typed-racket-test typed-scheme
-      - run: PLT_TR_CONTRACTS=1 racket -l typed-racket-test -- --unit
+      - run: sudo raco setup --check-pkg-deps typed typed-racket typed-racket-test typed-scheme
+        if: ${{ !matrix.enable-contracts }}
+      - run: sudo raco setup --check-pkg-deps typed typed-racket typed-racket-test typed-scheme
+        if: ${{ matrix.enable-contracts }}
+        env:
+          PLT_TR_CONTRACTS: 1
+      - run: racket -l typed-racket-test -- --unit
```
samth, line 34: I think this now only runs when contracts are enabled.

capfredf: This was a little weird to me at first. According to the old travis configuration, unittests only ran when contracts were enabled. see: https://travis-ci.org/github/racket/typed-racket/builds/753582875

samth: I think we should do it in both cases.

PR #1019, opened 2021-01-13
update the type annotation of `current-memory-use`

wargrey:


sorawee:
Is the current one wrong? If so, it would be nice to add a test case to `typed-racket-test/succeed/memory.rkt`.

samth:
I also don't understand what's different about the new type compared with the old one.

wargrey:
Ah, I am sorry, someone updated it before me in a recent commit.

PR #1024, opened 2021-01-21
parse `All`: check if the underlying symbols of type vars are duplicated

### LINKED ISSUE #1014: type variables are unhygiene
7.9
### What program did you run?
```racket
#lang typed/racket
(define-syntax-rule (def id vars ...)
  (begin
    (: id (All (vars ... a) (-> vars ... a a)))
    (define (id vars ... a) a)))

(def f a)
(def g x)
```
then in repl:
```racket
> (f 1 "2")
- : (U One String)
"2"
> (g 1 "2")
- : String
"2"
```
@Syntacticlosure told me about this problem.
 If I understand correctly, type variables are treated as symbols currently, which is not intuitive. 
At least, it should report an error when duplicated type variables appear in `All`.

capfredf:
closes #1014

### DIFF
--- a/typed-racket-lib/typed-racket/private/parse-type.rkt
+++ b/typed-racket-lib/typed-racket/private/parse-type.rkt
@@ -199,20 +199,20 @@
 (define (parse-all-type stx)
   (syntax-parse stx
     [(:All^ (vars:id ... v:id dd:ddd) . t:omit-parens)
-     (define maybe-dup (check-duplicate-identifier (syntax->list #'(vars ... v))))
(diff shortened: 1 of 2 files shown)


capfredf:
The error report would not be very good. For example, the error for the test file shows `(All (vars ... a) (-> vars ... a a))` has the duplicate variable `a`, which is confusing at first glance.

samth:
I think we should merge this so as to fix the unsoundness, but we can revisit the error reporting.

samth (approved):

Issue #1025, opened 2021-01-21
Support `racket/hash`


samth:
Functions from `racket/hash`, in particular `hash-union` and `hash-union!`, should have types.

PR #1029, opened 2021-01-22
Fix problems reported by @dzoep.

samth:


typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -1085,6 +1089,10 @@
 [make-immutable-custom-hash (->opt (-> Univ Univ Univ) (-> Univ -Nat) [(-> Univ -Nat)] Univ)]
 [make-weak-custom-hash (->opt (-> Univ Univ Univ) (-> Univ -Nat) [(-> Univ -Nat)] Univ)]
 
+
+[hash-union (-poly (a b) (->* (list (-Immutable-HT a b))  (-HT a b) (-Immutable-HT a b)))]
+[hash-union! (-poly (a b) (->* (list (-Mutable-HT a b))  (-HT a b) -Void))]
```
sorawee, line 1094: `hash-intersect` too?

samth: This was just aiming at #1025, which is just those two functions.

samth: In fact, there is no `hash-intersect`

sorawee: See https://github.com/racket/racket/pull/3310

samth: Ah, that's what I get for looking at docs.racket-lang.org. Unfortunately there's a typo in the docs ...

PR #1035, opened 2021-01-25
fix the types for procedure-rename and procedure->method

### LINKED ISSUE #1033: How to use `procedure-rename` to rename a procedure?
7.9[cs]

### What program did you run?
```
(define add (procedure-rename + 'add))
(displayln (add 1 2 3))
```

`6`

```
Type Checker: cannot apply a function with unknown arity;
 function `add' has type Procedure which cannot be applied
  in: (add 1 2 3)
```

capfredf:
close #1033

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -1179,8 +1179,8 @@
 [procedure? (make-pred-ty top-func)]
 [compose (-poly (a b c) (-> (-> b c) (-> a b) (-> a c)))]
 [compose1 (-poly (a b c) (-> (-> b c) (-> a b) (-> a c)))]
-[procedure-rename (-> top-func -Symbol top-func)]
-[procedure->method (-> top-func top-func)]
+[procedure-rename (-poly (a) (-> (-Inter top-func a) -Symbol a))]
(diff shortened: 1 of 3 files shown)


sorawee:
I guess it's not possible to fix `procedure-reduce-arity` in the same way, right?

samth:
That definitely needs a special case in the type checker, rather than just a type.

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -1179,8 +1179,8 @@
 [procedure? (make-pred-ty top-func)]
 [compose (-poly (a b c) (-> (-> b c) (-> a b) (-> a c)))]
 [compose1 (-poly (a b c) (-> (-> b c) (-> a b) (-> a c)))]
-[procedure-rename (-> top-func -Symbol top-func)]
-[procedure->method (-> top-func top-func)]
+[procedure-rename (-poly (a) (-> (-Inter top-func a) -Symbol (-Inter top-func a)))]
```
samth, line 1182: Why do we need the intersection on the result type?

capfredf: Actually, we don't. I'm pushing a fix on that.

samth (approved):

PR #1037, opened 2021-01-28
update some tests to pass on windows.

bdeket:
The following 3 test fail (for me) on windows:

- [fail/pr13289.rkt](https://github.com/racket/typed-racket/blob/master/typed-racket-test/fail/pr13289.rkt)
- [optimizer/test/known-length-vector-op.rkt](https://github.com/racket/typed-racket/blob/master/typed-racket-test/optimizer/tests/known-length-vector-op.rkt)
- [succeed/simple-kw-ap.rkt](https://github.com/racket/typed-racket/blob/master/typed-racket-test/succeed/simple-kw-app.rkt)

This PR solves this. But at least for the third test, I'm not convinced my approach is the best solution. The title of the test suggests what is being tested is kw-app but it uses the file system. So maybe a better approach is to use a kw function that doesn't need the filesystem. Or if it is the filesystem being tested, use a file from the typed-racket-test package, that is sure to be present.

### DIFF
--- a/typed-racket-test/fail/pr13289.rkt
+++ b/typed-racket-test/fail/pr13289.rkt
@@ -1,5 +1,5 @@
 #;
-(exn-pred #rx"fail/pr13289.rkt:9:10:.*in: Natural")
+(exn-pred #rx"fail(/|\\\\)pr13289.rkt:9:10:.*in: Natural")
 #lang typed/racket
 
 ;; This test ensures that the error message for misuse of
--- a/typed-racket-test/fail/pr13289.rkt
+++ b/typed-racket-test/fail/pr13289.rkt
@@ -1,5 +1,5 @@
 #;
-(exn-pred #rx"fail(/|\\\\)pr13289.rkt:9:10:.*in: Natural")
+(exn-pred #rx"fail.pr13289.rkt:9:10:.*in: Natural")
 #lang typed/racket
 
 ;; This test ensures that the error message for misuse of
--- a/typed-racket-test/optimizer/run.rkt
+++ b/typed-racket-test/optimizer/run.rkt
@@ -35,7 +35,9 @@
           ;; making program output identical on RacketCS is often worthwhile,
           ;; but optimization logs are too fragile in some cases
           (check-equal? (list (set-subtract log expected-log) (set-subtract expected-log log)) (list (list) (list))))
-        (check-equal? (regexp-split "\n" output) (regexp-split "\n" expected-output))))))
+        (check-equal? (regexp-split "\n" output) (regexp-split "\n" (if (eq? (system-type 'os) 'windows)
+                                                                        (regexp-replace* "\r" expected-output "")
+                                                                        expected-output)))))))
 
 
 (define-runtime-path tests-dir                "./tests")
--- a/typed-racket-test/optimizer/run.rkt
+++ b/typed-racket-test/optimizer/run.rkt
@@ -9,7 +9,7 @@
          generate-log tests-dir missed-optimizations-dir)
 
 (define (get-expected-results file)
-  (with-input-from-file file
+  (with-input-from-file file #:mode 'text
     (lambda () ; from the test file
       (read-line) ; skip the #;#;
       (values (for/list ((l (in-lines (open-input-string (read))))) l)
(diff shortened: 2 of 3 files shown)


samth:
Thanks for this. I think using a file from typed-racket-test is the right solution for the third one. For the second one, why does only that test fail? Your change is quite general, but if that change is needed I would expect a lot of tests to fail.

bdeket:
Re (2). I don't know exactly, but I guess that at some point the known-length-vector-op.rkt file has been edited on a windows computer. git replaces \r\n with \n when it commits files, but I guess it only puts \r\n back if at some point it has had this to begin with.
If I open other files in the same folder I only see the \n ending character. Only for the mentioned file the \r\n shows up.

I went this more general solution to prevent a future problem if another file is committed/saved with \r\n line endings.

typed-racket-test/fail/pr13289.rkt
```diff
@@ -1,5 +1,5 @@
 #;
-(exn-pred #rx"fail/pr13289.rkt:9:10:.*in: Natural")
+(exn-pred #rx"fail(/|\\\\)pr13289.rkt:9:10:.*in: Natural")
```
capfredf, line 2:
```
(exn-pred (regexp (format "~a:9:10:.*in: Natural" (build-path "fail" "pr13289.rkt"))))
```
This might be better

typed-racket-test/optimizer/run.rkt
```diff
@@ -35,7 +35,9 @@
           ;; making program output identical on RacketCS is often worthwhile,
           ;; but optimization logs are too fragile in some cases
           (check-equal? (list (set-subtract log expected-log) (set-subtract expected-log log)) (list (list) (list))))
-        (check-equal? (regexp-split "\n" output) (regexp-split "\n" expected-output))))))
```
capfredf, line 38: I would do 
```
(let ([linebreak (if (eq? (system-type 'os) 'windows) "\r\n" "\n"])
  (check-equal? (regexp-split linebreak output) (regexp-split linebreak expected-output))))))
```

capfredf: But I am wondering if there is a better way to get the platform-specific linebreak in Racket.

bdeket: I tried something similar with "\r?\n"
However, this did not work. The last element still had an "\r". I'm guessing that the way `expected-output` is constructed, the last \n gets stripped.

samth: I think we should just fix the file and then not make this change. Our files should just be expected to have Unix line endings.

shhyou: I used port->lines

shhyou: If the tests are reading local checkout files, some git configuration will automatically convert Unix line breaks to Windows line breaks.

The xml tests had that issue before 
https://github.com/racket/racket/blob/master/pkgs/racket-test/tests/xml/test.rkt#L91

If #:mode is 'text, \r\n should've been converted to \n though.

typed-racket-test/fail/pr13289.rkt
bdeket: Unfortunatly it doesn't work because then you get `#rx"(fail\\pr132..."` instead of `#rx"(fail\\\\pr132..."`

shhyou: How about:
```
(string-append
 (regexp-quote (format "~a:9:10:" (build-path "fail" "pr13289.rkt")))
 ".*in: Natural")
```
output: `"fail/pr13289\\.rkt:9:10:.*in: Natural"`

samth: How about we avoid checking for the path separator entirely? Something like `"fail.*:9:10:.*in: Natural"`.

typed-racket-test/optimizer/run.rkt
bdeket: Thanks @shhyou, that was the right solution.

typed-racket-test/fail/pr13289.rkt
bdeket: I don't know about dropping the filename. If the test is for checking the right source location, is the filename not more important than that there was the word "fail" somewhere in the message.
Is the `(/|\\\\)` so bad, if so I can change it to @shhyou's solution.

samth: Sorry, I didn't mean to drop the filename there. I'm suggesting `"fail.*pr13289.rkt:9:10:.*in: Natural

bdeket: oh right. And we can use just `.` since `/` and `\\` are both only one caracter.

PR #1040, opened 2021-02-08
Unify types for quotient, remainder and quotient/remainder

capfredf:
use the same spec to generate types for them.

closes #1039

PR #1052, opened 2021-03-03
Repair types for quotient/remainder.

samth:
Should fix the lambda-sh package. Fixes changes in #1040.

cc @capfredf @wargray


capfredf:
LGTM

Issue #1055, opened 2021-03-13
different replies to the pattern matching `(p ,_ ...) `(p . ,_)


wnsena:
Racket version 8.0 [cs]

#### Why does it work
``` racket
#lang typed/racket

(define-type Term (Listof (U Symbol Term)))

(: get-tokens (-> Term (Listof Symbol)))
(define (get-tokens term)
  (match term
    [`()                       `()]
    [`(,(? list? tk) . ,tks)   `(,@(get-tokens tk) ,@(get-tokens tks))]
    [`(,(? symbol? tk) . ,tks) `(,tk ,@(get-tokens tks))]))
````

#### And it doesn't work?
``` racket
#lang typed/racket

(define-type Term (Listof (U Symbol Term)))

(: get-tokens (-> Term (Listof Symbol)))
(define (get-tokens term)
  (match term
    [`()                         `()]
    [`(,(? list? tk) ,tks ...)   `(,@(get-tokens tk) ,@(get-tokens tks))]
    [`(,(? symbol? tk) ,tks ...) `(,tk ,@(get-tokens tks))]))
````

#### Error
``` racket
. Type Checker: type mismatch
  expected: Term
  given: (Listof Any) in: tks
. Type Checker: type mismatch
  expected: Term
  given: (U Symbol Term) in: tks
. Type Checker: Summary: 2 errors encountered in:
  tks
  tks
```

maueroats:
I have no solution, but wrote this to give more context in case someone tries to work on it later.

I guess you know you are matching a [quasipattern](https://docs.racket-lang.org/reference/match.html?q=match#%28form._%28%28lib._racket%2Fmatch..rkt%29._match%29%29)? I think this report shows the type inferencing for quasipatterns behaving in a different way from "equivalent" non-quasipatterns. (As far as I understand it.)

The good: the type system is making the desired deductions when applied to non-quasipatterns `thing ...` (see code below), and also when applied to the form `. last-qp` in a quasipattern (first example in the report above).

The bad: the type system is not making the desired deductions when applied to the quasipattern `qp ...` (second example in the code above). The type system can only figure out that this is type `(Listof Any)` but in context above it should be to be of type `(Listof Term)`. 

I'm going to give up on it here.

```racket
(: get-tokens (-> Term (Listof Symbol)))
(define (get-tokens term)
  (match term
    [`()                           `()]
    [(list (? list? tk) tks ...)   `(,@(get-tokens tk) ,@(get-tokens tks))]
[...]

PR #1057, opened 2021-03-16
Add a match example with an annotation and ellipsis (...)

maueroats:
Document how type annotations help the expanded code for match type check. Documents the explanation provided by samth in issue #1055.

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
@@ -244,7 +244,17 @@ anticipate with their pattern matching, e.g.:
   (eval:error
    (size 42))]
 
+Patterns involving an ellipsis @racket[...] for repetition may generate a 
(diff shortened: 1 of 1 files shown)


typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
```diff
@@ -244,7 +244,17 @@ anticipate with their pattern matching, e.g.:
...
+Patterns involving an ellipsis @racket[...] for repetition may generate a 
+@racket[for] loop that requires annotations on variables to type check. 
+The (deliberately obscure) code below does not type check without the type 
+annotation on the match pattern variable @racket[c].
 
+@examples[#:label #f #:eval the-eval
+  (: without-10 : (Listof Integer) -> (Listof Integer))
+  (define (without-10 lst)
+    (match lst
+      [`(10 ,#{c : (Listof Integer)} ...)   c]))
```
samth, line 256: If you include a check like `(? number?)` then this annotation will still be needed after I fix that bug.

PR #1058, opened 2021-03-16
Match typing docs - show use of annotations with ellipsis (...)

maueroats:
Update of #1057


typed-racket-doc/typed-racket/scribblings/guide/caveats.scrbl
```diff
@@ -244,7 +244,17 @@ anticipate with their pattern matching, e.g.:
   (eval:error
    (size 42))]
 
+Patterns involving an ellipsis @racket[...] for repetition may generate a 
+@racket[for] loop that requires annotations on variables to type check. 
+The (deliberately obscure) code below does not type check without the type 
+annotation on the match pattern variable @racket[c].
 
+@examples[#:label #f #:eval the-eval
+  (: without-10 : (Listof Integer) -> (Listof Integer))
```
sorawee, line 253: The name `without-10` is kinda weird, now that the first element is not 10.

maueroats: Name improved to `skip-a-num`, which is kind of casual.
I also consided `rest-nums` and variations on it (`rest-nums-only`, `rest-numeric`, `rest-intonly`).
I don't have a real preference or guiding principle for the choice.

samth:
Sorry, I wasn't quite clear on what I meant. Here's the kind of program I'm suggesting:

```
(: suml : (Listof Any) -> Number)
(define (suml l)
   (match l
     [(list (? number? #{c : (Listof Number)}) ...) (apply + c)]
     [_ 0]))
```

maueroats:
This should be a draft until the code is tested with the recent typed racket fixes by @samth.

samth:
I changed `match` so you should be able to see the original example working in the latest snapshot.

samth:
Is this ready?

maueroats:
"Real Work" is busy, I will ping you when this is clean. Soon.

maueroats:
The following code runs in Racket but generates an error in Scribble. Any help understanding why? @samth 
```racket
@examples[#:label #f #:eval the-eval
  (: do-nothing (-> (Listof Integer) (Listof Integer)))
  (define (do-nothing lst)
    (match lst
      [(list (? number? #{c : (Listof Integer)}) ...)   c]))]
```

samth:
The problem is that the typed/racket reader extends the racket reader to support that syntax. I think you probably want to use codeblock and specify the language explicitly.

maueroats:
Replaced by #1070

PR #1065, opened 2021-03-22
Add `for/foldr` and `for*/foldr`.

NoahStoryM:


samth:
This looks great, but is it possible to combine the implementation of `for/fold` and `for/foldr`? They look almost identical.

NoahStoryM:
> This looks great, but is it possible to combine the implementation of `for/fold` and `for/foldr`? They look almost identical.

Yes, I redefined them with `define-syntaxes`.

samth:
Excellent! Is it possible to further combine the `for/fold[r]` and `for*/fold[r]` versions?

NoahStoryM:
> Excellent! Is it possible to further combine the `for/fold[r]` and `for*/fold[r]` versions?

I'm not sure how to further combine them. Maybe I should define `define-for[*]-variant/fold-style` to abstract them?

capfredf:
This looks nice. BTW, at the first glance I think maybe we can very likely abstract over the similarities of all `for` structures

NoahStoryM:
I am not very familiar with the use of macro, I am not sure if I can combine all `for` structures well..

capfredf:
@NoahStoryM No worries. We can do that later. 
@samth I think we can send another PR to combine for*/foldr, for/foldr, for*/fold, for/fold, for/lists, for*/lists and etc.

capfredf:
Also you should add entries for `for/foldr` and `for/fold` to [the reference](https://github.com/racket/typed-racket/blob/master/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl#L227)

NoahStoryM:
I have added them to the [reference](https://github.com/racket/typed-racket/pull/1065/commits/d6edace93db25337ee17e0b18880681f33dd9a34).

Is this all right? I think you'll further combine all `for`, so I don't define the macro for abstracting `for[*]/lists` and `for[*]/fold[r]`.

PR #1070, opened 2021-04-15
Adding typing to match when (? number? c) is involved

maueroats:
Add a match example with an annotation and ...
Document how type annotations help the expanded code for match type check.


maueroats:
Replaces #1058, sorry I don't know how to clean up that one.

maueroats:
@samth Clean commit with working code example for the match with typing docs.

samth:
Thanks!

PR #1072, opened 2021-04-16
Make Typed Racket implementation structs authentic.

samth:


capfredf:
Should we also make other structs authentic, like those in `static-contracts/`?

samth:
Yes, I think so.

PR #1075, opened 2021-04-18
Update the type of tab-panel%

### LINKED ISSUE #1074: Racket GUI tab-panel% has added a few new features.
This is just a bug to indicate that the signature of typed racket `tab-panel%` is out of date

Version 8.0 cs

### What program did you run?

Here's the interactive thing I ran (errors included), interactive prompt was `racket -I typed/racket`

```racket
> (require typed/racket/gui)
> (define f (new frame%
                 [label "test frame"]
                 [width 500]
                 [height 500]
                 [alignment '(center top)]))

> (define tab-elm
    (new tab-panel%
         [choices '("one" "two" "three")]
         [parent f]
         [style '(can-reorder can-close no-border)]
         [callback (lambda (panel event)
                     (displayln (format "panel ~a\nevent ~a"
                                        panel event))
                                )
                   ]
         ))
; readline-input::307: Type Checker: type mismatch
;   expected: (Listof (U 'deleted 'no-border))
;   given: (List 'can-reorder 'can-close 'no-border)
;   in: (quote (can-reorder can-close no-border))
; [,bt for context]
>
```
(please forgive the ugly parens, but I'm leaving them just in case of data inconsistancy)

This exact same code runs just fine in normal racket with `racket/gui` instead of `typed/racket/gui`. This should allow not only the new styles but the related functions.

The [related functions][tab-panel-docs] are thus:
- `on-reorder`
- `on-close-request`

The relevant (and now partially resolved) issues for improoving this panel are [this][racket-gui-89] and [this][racket-gui-184]

[racket-gui-89]: https://github.com/racket/gui/issues/89
[racket-gui-184]: https://github.com/racket/gui/issues/184
[tab-panel-docs]: https://docs.racket-lang.org/gui/tab-panel_.html

See above interactive prompt.

Lazerbeak12345:
closes #1074 

Can someone review this? The two new functions are intended to be augmented, and I'm not 100% sure if that's how the signature would look.


Lazerbeak12345:
I can squash this if needed, but I figured you could do it with just as much (if not less) effort.

typed-racket-more/typed/racket/private/gui-types.rkt
```diff
@@ -1592,6 +1592,8 @@
                [stretchable-height Any #:optional])
          [append (String -> Void)]
          [delete (Integer -> Void)]
+         [on-reorder (Listof Exact-Nonnegative-Integer)]
```
capfredf, line 1595: according to https://docs.racket-lang.org/gui/tab-panel_.html?q=on-reorder%20#%28meth._%28%28%28lib._mred%2Fmain..rkt%29._tab-panel~25%29._on-reorder%29%29, *on-reorder* is a *method* that takes a list of exact-nonnegative-integer? and returns void. So is *on-close-request*

That's why the test failed.

capfredf, line 1595:
```suggestion
         [on-reorder ((Listof Exact-Nonnegative-Integer) -> Void)]
```

typed-racket-more/typed/racket/private/gui-types.rkt
```diff
@@ -1592,6 +1592,8 @@
...
+         [on-close-request Exact-Nonnegative-Integer]
```
capfredf, line 1596:
```suggestion
         [on-close-request (Exact-Nonnegative-Integer -> Void)]
```

capfredf (changes requested):

typed-racket-more/typed/racket/private/gui-types.rkt
Lazerbeak12345: Ah! Right, I can't believe I didn't notice that! Glad to have another pair of eyes on this.

PR #1077, opened 2021-04-19
fix the constructor expr in the exported structure's struct-info

capfredf:


samth:
Can you write a test for this?

typed-racket-lib/typed-racket/typecheck/def-binding.rkt
```diff
@@ -116,17 +116,29 @@
...
 
      (define/with-syntax (constr* type-desc* pred* super* accs* ...)
        (for/list ([i (in-list (cons constr-new-id new-ids))])
          (and (identifier? i) #`(quote-syntax #,i))))
 
+     (define/with-syntax constr-in-si*
+       (if (free-identifier=? new-id constr-new-id)
+           new-id
+           constr-new-id))
+
```
samth, line 131: Why are we defining this twice?

capfredf: good catch

typed-racket-test/succeed/test-struct-info.rkt
```diff
@@ -0,0 +1,39 @@
...
+  (check-constr kiwi kiwi)
+  (check-constr apple make-apple)
+  (check-constr pear pear)
+  (check-constr dragon-fruit make-dragon-fruit)
+  (check-constr DragonFruit make-dragon-fruit)
+  (check-constr PEAR pear))
+
+(require 'typed)
+(require 'test)
+(check-constr apple make-apple)
```
samth, line 35: I feel like this test could be simpler/more direct. For example, is `(object-name apple)` always `'make-apple` here? Or can we just check that everything works correctly as a procedure?

capfredf: Do you mean we use `object-name` at phase 0? my understanding is 
1. if v is simply a structure type identifier and not a constructor name, then we will see racket/typed racket complains identifier for static struct-type information cannot be used as an expression before `(object-name v)` run
2. `object-name` doesn't return the constructor name. E.g `(object-name (make-apple))` will return `'apple`
3. If I understand you correctly, everything already almost worked correctly as a procedure before this patch. Previously we used `export-id` as the constructor name. In this case, the struct info of `(struct apple () #:constructor-name make-apple)`  incorrectly had some form of #'apple as the constructor. But we found no errors from running TR's tests, buiding packages and running their tests until `delay-pure` which directly used the struct info of typed structure types.

samth: I thinking of this code:

```
#lang racket

(module test racket/base
  (require  (for-syntax racket/base syntax/parse racket/struct-info racket/match))

  (define-syntax (check-constr stx)
    (syntax-parse stx
      [(_ sname:id)
       (define si (extract-struct-info (syntax-local-value #'sname (lambda () #f))))
       (match-define (list* struct-desc constr rst) si)
       constr]))
  (provide check-constr))
(module t typed/racket/base
  (require (submod ".." test) typed/rackunit)
  (struct kiwi ())
  (struct apple () #:constructor-name make-apple)
  (struct pear () #:type-name PEAR)
  (struct dragon-fruit () #:type-name DragonFruit #:constructor-name make-dragon-fruit)

  (check-equal? (check-constr apple) 'make-apple)

  (provide (struct-out apple)
           (struct-out kiwi)
           (struct-out PEAR)
           (struct-out DragonFruit)))

(require 'test 't rackunit)
(check-equal? (check-constr apple) 'make-apple)
```

capfredf: Ah, this is better

typed-racket-test/succeed/test-struct-info.rkt
```diff
@@ -0,0 +1,40 @@
+#lang racket/base
+(module test racket/base
+  (require (for-syntax racket/base syntax/parse racket/struct-info racket/match))
+
+  (define-syntax (get-constr stx)
+    (syntax-parse stx
+      [(_ sname:id)
+       (define si (extract-struct-info (syntax-local-value #'sname (lambda () #f))))
+       (match-define (list* struct-desc constr rst) si)
+       #`'#,(syntax-e constr)]))
```
samth, line 10: Why `syntax-e`? Can't you just use `constr` here?

samth: I guess you are trying to see the _name_ of the binding that is specified, but I think we shouldn't care about that, and should care about it having the right name.

capfredf: then for `check-equal?`, we would be using its runtime value, which is a procedure

samth (approved):
Good when tests pass.

PR #1078, opened 2021-04-21
fix a bug where re-exported struct names via require/typed/provide don't have struct infomation

### LINKED ISSUE #837: Untyped code requiring struct from `require/typed`

`7.3--2019-06-14`

### What program did you run?

```rkt
(module a racket/base
  (struct foo ())
  (provide (struct-out foo)))

(module b typed/racket/base
  (require/typed (submod ".." a) [#:struct foo ()])
  (provide (struct-out foo)))

(module c racket/base
  (require (submod ".." b))
  (provide (struct-out foo)))
```

No error.

```
struct-out: identifier is not bound to struct type information in: foo
```

capfredf:
closes #837 
supersedes #926

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims-contract.rkt
@@ -39,11 +39,11 @@
          #'(begin (provide (rename-out [names id] ...))
                   (define-syntax (names stx) (id stx)) ...))]))
   (def require/opaque-type
-        require-typed-signature
-        require-typed-struct-legacy
-        require-typed-struct
-        require/typed-legacy require/typed require/typed/provide
-        require-typed-struct/provide make-predicate define-predicate)
+       require-typed-signature
+       require-typed-struct-legacy
+       require-typed-struct
+       require/typed-legacy require/typed require/typed/provide
+       require-typed-struct/provide make-predicate define-predicate)
 
   ;; Expand `cast` to a `core-cast` with an extra `#%expression` in order
   ;; to prevent the contract generation pass from executing too early
@@ -448,7 +448,7 @@
                       ;The actual identifier bound to the constructor
                       [real-maker (if (syntax-e #'id-is-ctor?) #'internal-maker #'maker-name)]
                       [extra-maker (and (attribute input-maker.extra)
-                                        (not (bound-identifier=? #'make-name #'nm))
+                                        (not (bound-identifier=? #'maker-name #'nm))
                                         #'maker-name)]
                       [type (if (stx-null? #'(tvar ...))
                                 #'type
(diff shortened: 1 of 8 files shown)


bennn:
Hi @capfredf , I just ran into this issue again (with @llazarek). 

Do you remember what's the status here? In particular, are there any majors problems that would prevent a fix?

capfredf:
@bennn It has been a while, so my memory is a bit vague. If I recall correctly, there are some issues with imported structs with `#:constructor-name` specified via `require/typed`. The real bigger issue that stopped me from proceeding with this PR was https://github.com/racket/macro-debugger/issues/34, which made debugging the issue much harder.

bennn:
Ok, thanks! I may try to pick this up. (Hopefully one of the tests here illustrates the `#:constructor-name` problem, but if not that's a good hint to start with.)

capfredf:
@bennn Sam and I talked about this pr. Since this pr as-is is an improvement, we've decided to move ahead and try to get this merged into the release of 8.6. The I will work on the fix/support for `extra-constructor-name` later

typed-racket-lib/typed-racket/typecheck/def-binding.rkt
```diff
@@ -109,18 +109,20 @@
      (define sname-is-constructor? (and (or extra-constr-name (free-identifier=? sname constructor-name)) #t))
      (define constr (or extra-constr-name constr^))
      (define type-is-sname? (free-identifier=? tname internal-id))
+
      ;; Here, we recursively handle all of the identifiers referenced
      ;; in this static struct info.
      (define-values (constr-defn constr-export-defn constr-new-id constr-aliases)
        (cond
-         [(not (identifier? constr))
```
capfredf, line 116: Revert

samth:
Is this done?

capfredf:
I made a mistake when I rebased the commits. It should be fixed now.

typed-racket-lib/typed-racket/base-env/prims-contract.rkt
```diff
@@ -448,7 +448,7 @@
                       ;The actual identifier bound to the constructor
                       [real-maker (if (syntax-e #'id-is-ctor?) #'internal-maker #'maker-name)]
                       [extra-maker (and (attribute input-maker.extra)
-                                        (not (bound-identifier=? #'make-name #'nm))
+                                        (not (bound-identifier=? #'maker-name #'nm))
```
samth, line 451: Is this the only substantive change, other than removing the `type-only` field everywhere?

typed-racket-lib/typed-racket/typecheck/def-binding.rkt
```diff
@@ -107,8 +107,9 @@
              (mk-ignored-quad e))))
 
      (define sname-is-constructor? (and (or extra-constr-name (free-identifier=? sname constructor-name)) #t))
-     (define constr (or extra-constr-name constr^))
+     (define constr (or extra-constr-name constructor-name))
```
capfredf, line 110: @samth the other meaningful change is here. The constructor name from the repackaged struct info by `request/typed` are not actually exported as the constructor, so we use the actual constructor name received by `dtsi*`.

capfredf: It is very likely that we can simplify the code for require/typed for structs, but we should do that in this PR

PR #1084, opened 2021-05-04
Fix make-log-receiver type signature.

SnapCracklePopGone:
Extend type signature of make-log-receiver to allow n-ary logging topics and log levels.  Current type signature supports only a single logging topic.

This change is more restrictive than the documented behavior of the untyped make-log-receiver which allows for defaulting the topic to #f.  

This fix allows for a typical usage of make-log-receiver to properly type check. e.g.
(make-log-receiver (current-logger) 'info #f 'debug 'my/lib/logger)


samth:
The tests fail here: https://github.com/racket/typed-racket/pull/1084/checks?check_run_id=2500914122#step:9:5, I think because the arguments were previously optional.

capfredf:
@samth The type is more in line with the specification, although it is not entirely right.

PR #1089, opened 2021-05-07
fix struct-related bugs:

capfredf:
1. fix a bug where the construct-name wasn't contracted.
2. fix a bug where the struct info was created with the struct name instead of
the extra constructor name when it was supplied


capfredf:
I realized the two issues when working on #1078. I decided spin the bugfix off of #1078 because safely removing `#:type-only` in `dtsi*` and other places for `require/typed` involves handling a lot of indirection correctly and this bugfix is unrelated to the refactoring

samth (approved):

PR #1095, opened 2021-05-12
Use new setup-racket features.

samth:


samth:
This is not yet using the correct package catalog, I think some changes are probably needed by @bogdanp.

PR #1096, opened 2021-05-12
fix internal contract errors

### LINKED ISSUE #1073: internal contract violation
installing [mox](https://github.com/wargrey/mox) with PLT_TR_CONTRACTS enabled reveals a contract violation:
```
-is-type: contract violation
  expected: Type?
  given: #f
  in: the 2nd argument of
      (->
       (or/c integer? name-ref/c OptObject?)
       Type?
       Prop?)
  contract from: (function -is-type)
  blaming: <pkgs>/typed-racket-lib/typed-racket/rep/prop-rep.rkt
  at: <pkgs>/typed-racket-lib/typed-racket/rep/prop-rep.rkt:47.23
  context...:
   /Users/capfredf/code/racket/racket/collects/racket/contract/private/blame.rkt:346:0: raise-blame-error
   /Users/capfredf/code/racket/racket/collects/racket/contract/private/arrow-higher-order.rkt:379:33
   /Users/capfredf/code/racket-extra-pkgs/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-envops.rkt:35:8: loop
   /Users/capfredf/code/racket-extra-pkgs/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-if.rkt:13:0: tc/if-twoarm
   /Users/capfredf/code/racket/racket/collects/racket/contract/private/arrow-higher-order.rkt:375:33: ...row-higher-order.rkt:375:33
   /Users/capfredf/code/racket/racket/collects/racket/contract/private/arrow-higher-order.rkt:375:33
   .../private/runtime.rkt:89:23: fail-handler141
   /Users/capfredf/code/racket-extra-pkgs/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt:98:0: tc-expr/check
   /Users/capfredf/code/racket/racket/collects/racket/contract/private/arrow-higher-order.rkt:375:33
   /Users/capfredf/code/racket-extra-pkgs/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-let-unit.rkt:103:0: check-let-body33
   /Users/capfredf/code/racket/racket/collects/racket/contract/private/arrow-higher-order.rkt:375:33: ...row-higher-order.rkt:375:33
   /Users/capfredf/code/racket/racket/collects/racket/contract/private/arrow-higher-order.rkt:375:33
   .../private/runtime.rkt:89:23: fail-handler141
   /Users/capfredf/code/racket-extra-pkgs/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt:98:0: tc-expr/check
   /Users/capfredf/code/racket/racket/collects/racket/contract/private/arrow-higher-order.rkt:375:33
   /Users/capfredf/code/racket-extra-pkgs/typed-racket/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt:95:0: tc-lambda-body62
   ...
```

capfredf:
closes #1073

### DIFF
--- a/typed-racket-lib/typed-racket/rep/prop-rep.rkt
+++ b/typed-racket-lib/typed-racket/rep/prop-rep.rkt
@@ -155,4 +155,7 @@
 (def-prop AndProp ([ps (listof (or/c OrProp? TypeProp? NotTypeProp? LeqProp?))])
   [#:frees (f) (combine-frees (map f ps))]
   [#:fmap (f) (apply -and (map f ps))]
-  [#:for-each (f) (for-each f ps)])
+  [#:for-each (f) (for-each f ps)]
+  [#:custom-constructor/contract
+   (-> (listof (or/c TrueProp? OrProp? TypeProp? NotTypeProp? LeqProp?)) AndProp?)
+   (make-AndProp (filter (lambda (x) (not (TrueProp? x))) ps))])
--- a/typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
@@ -81,12 +81,13 @@
                                              (env-set-id-type Γ x new-t*)
                                              obj
                                              pt)]
-                       [else
-                        (loop ps
-                              (cons (-is-type obj (path-type pes new-t*)) atoms)
-                              negs
-                              (append new-props new)
-                              (env-set-id-type Γ x new-t*))])]))]
+                       [(path-type pes new-t*) => (lambda (pt)
+                                                    (loop ps
+                                                          (cons (-is-type obj pt) atoms)
+                                                          negs
+                                                          (append new-props new)
+                                                          (env-set-id-type Γ x new-t*)))]
+                       [else #f])]))]
                [(TypeProp: obj pt)
                 (update-obj-pos-type new Γ obj pt)]
                ;; process negative info _after_ positive info so we don't miss anything!
--- a/typed-racket-test/succeed/internal-contract-check1.rkt
+++ b/typed-racket-test/succeed/internal-contract-check1.rkt
@@ -1,0 +1,8 @@
+#lang typed/racket/base
+(struct foo ([a : Any]))
+(define-type Hello (Pairof foo (Listof foo)))
+(define-predicate hello? Hello)
+(define #:forall (T) (hello [v : T])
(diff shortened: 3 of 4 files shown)


samth:
Can we add a test for this?

capfredf:
Yes, I am working on it, but I have some trouble reproducing the error.

capfredf:
While trying to writing a minimal example to reproduce an error with the [master](c9bff1536dbeae641796fbf57d37f0e6e90afee8) of TR, I noticed a very weird behavior:
```
#lang typed/racket/base

(struct foo ([a : Any]))
(define (hello [v : (Pairof Any Any)])
  (if (and (foo? (car v)) (number? (foo-a (car v))))
      (add1 (foo-a (car v)))
      0))
```
This program only type-checks 7 or 8 out of 10 times.

capfredf:
I know what is happening. Before checking `(add1 (foo-a v))`, we have a conjunction of propositions `v is foo` and `(foo-a v) is number`. The conjunction is essentially a list of those two propositions and the order of their appearances decide the result. The creation of the list is done through [iteration of hash-maps](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/types/prop-ops.rkt#L245). The iteration order is consistent with BC, but with CS it is not.

samth:
Great work. I was suspicious that it was about hash maps.

Options:
1. Make the order not matter, by handling the conjunction better.
2. Fix the order by sorting.

capfredf:
@samth The problem is like  "we know `v is foo` so `(foo-a v)` is `Any`, hence we should reject the program"
vs "we know (foo-a v) is number let's add one to it!"
I think we should as you said, fix the conjunction better. In this case, `(foo-a v) is number` should not exist, because it conflicts with `v is foo`. 
Moreover, do we[ have path elements for structure types](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/typecheck/tc-structs.rkt#L296) only when they are polymorphic? If so, should we remove path elements when structures types are monomorhpic?

samth:
It seems like we could just keep both of them, and then we go to use them, `N @ (foo-a v)` should be available, and the program should work.

capfredf:
The inconsistency shown in #1098 made me rethink about reproducing the contract error in #1073. 

Running the newly committed `typed-racket-test/succeed/internal-contract-check2.rkt` several times with contracts enabled should expose the contract error. 

cc @samth

samth (approved):

PR #1099, opened 2021-05-22
fix the order of paths to objects in the typechecking process

### LINKED ISSUE #1098: inconsistent typechecking results
7.6 or later (CS)
### What program did you run?

```
#lang typed/racket/base

(struct foo ([a : Any]))
(define (hello [v : Any])
  (if (and (foo? v) (number? (foo-a v)))
      (add1 (foo-a v))
      0))
```

This program only type-checks 7 or 8 out of 10 times. The type checker should have always reported the same error as it does when running with BC versions

capfredf:
closes #1098


typed-racket-lib/typed-racket/typecheck/tc-envops.rkt
```diff
@@ -29,7 +29,20 @@
...
+
+     (define atoms (if atoms^
+                       ;; Fix the order of paths to the same object.
+                       ;; Objects without path elements always come last
+                       (sort atoms^ (lambda (x y)
+                                      (match* (x y)
+                                        [((TypeProp: (Path: pes1 (? identifier? var1)) _)
+                                          (TypeProp: (Path: pes2 (? identifier? var2)) _))
+                                         #:when (equal? var1 var2)
+                                         (not (null? pes1))]
```
samth, line 42: Shouldn't this use a comparison that's more useful in more cases, like `(< (length pes1 pes2))`, or something like that? In general I worry about a comparison function that isn't an order.

capfredf: my first attempt was indeed using `(< (length pes 1 pes2))`, but I was not so sure it was completely right. 
I will write some tests to make myself confident about the general solution.

typed-racket-lib/typed-racket/types/path-type.rkt
```diff
@@ -51,8 +56,10 @@
       ;; struct ops (non-prefab)
       [((Struct: _ _ flds _ _ _ _) (cons (StructPE: struct-ty idx) rst))
        #:when  (subtype t struct-ty)
-       (match-let ([(fld: ft _ _) (list-ref flds idx)])
-         (path-type rst ft (hash)))]
+       (handle-struct rst flds idx)]
+
+      [((Univ:) (cons (StructPE: struct-ty idx) rst))
```
samth, line 61: I think this should be something a little more general, so that it handles `(U Struct String)` instead of just `Any`.

capfredf: Ah Yes! Thanks!

capfredf:
@samth I am thinking [the program](https://github.com/racket/typed-racket/pull/1099/files#diff-d91214bc2b485389d1f1f86dfe0f274d4f0c4adbf8d15567f9324c8bd54fb453) should type check.
```racket
(struct bar ([a : Any]))

(define (five [v : Any])
  (if (bar? v)
      (let ([a (bar-a v)])
        (if (number? a)
            (add1 (bar-a v))
            42))
      0)) ;;  always work

(define (five+ [v : Any])
  (if (and (bar? v) (number? (bar-a v)))
      (add1 (bar-a v))
      0)) ;; has the inconsistent error
```
my sense is that making `five+` not type check but five type check would be hard for people to reason about.

typed-racket-lib/typed-racket/types/update.rkt
```diff
@@ -47,28 +47,46 @@
...
           ;; Note: this updates fields even if they are not polymorphic.
           ;; Because subtyping is nominal and accessor functions do not
           ;; reflect this, this behavior is unobservable except when a
           ;; variable aliases the field in a let binding
-          (match-define-values (lhs (cons (fld: ty acc-id #f) rhs)) (split-at flds idx))
-          (match (update ty rst)
-            [(Bottom:) -Bottom]
-            [ty (let ([flds (append lhs (cons (make-fld ty acc-id #f) rhs))])
-                  (make-Struct nm par flds proc poly pred props))])]
+          (let/cc fail (Struct-update t flds
```
samth, line 56: This can be `let/ec`.

samth (commented):
Can you give a (significanlty) longer comment in the commit message explaining what's going on here, and maybe also in the file?

typed-racket-lib/typed-racket/types/update.rkt
```diff
@@ -47,28 +47,46 @@
          ;; struct ops
          ;; (NOTE: we assume path elements to mutable fields
          ;;        are never created)
-         [((Struct: nm par flds proc poly pred props)
-           (StructPE: s idx))
+         [((? Struct? t) (StructPE: s idx))
```
capfredf, line 50: This is incorrect. Still looking for a solution to what I want

typed-racket-doc/typed-racket/scribblings/guide/occurrence.scrbl
```diff
@@ -193,6 +193,24 @@ interactions, which may include mutations. It is possible to work
 around this by moving the variable inside of a module or into a local
 binding form like @racket[let].
 
+@section{Access to structure fields}
+
+Occurrence typing can work with accessors to immutable structure fields which
+are polymorphic or have type @racket[Any]:
```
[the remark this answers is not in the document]
capfredf: Yes, it should also work for (U apple something) or a subtype of apple

capfredf:
@samth I ditched the original change to path-type.rkt, which turns out to be unnecessary given the way we sort path elements. Instead, I made a small change https://github.com/racket/typed-racket/pull/1099/files#diff-f0608f18ee591476ebe281bbde62dcc5547ccee33c7919deed382e237b63fe1aR53

samth (approved):

PR #1107, opened 2021-06-07
Types for `vector-sort` and `vector-sort!`.

samth:


typed-racket-test/unit-tests/typecheck-tests.rkt
```diff
@@ -5324,6 +5324,10 @@
          (hash 0 1 2 3 4 5 6 7 8 0)
          (-Immutable-HT -Byte -Byte))
 
+       (tc-e
+        (regexp-match* #rx"foo" "foobar" #:match-select car #:gap-select? #t)
```
capfredf, line 5328: should we add an integrated test for this? `tc-e` won't evaluate the expression.  If we gave a wrong type for `regex-match` like we had done before, I think this unit test wouldn't help either.

samth: Yes, that sounds like a good idea.

capfredf (approved):

PR #1109, opened 2021-06-10
improve the true proposition from the type of equal/eqv/eq

capfredf:
for (equal? a b), when the type of a has an overlap with that of b,
i.e the meet of those two types are not Bottom, put the overlap in the true
proposition.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-eq.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-eq.rkt
@@ -3,6 +3,7 @@
 (require "../../utils/utils.rkt"
          "signatures.rkt"
          "utils.rkt"
+         (only-in (infer infer) intersect)
          syntax/parse syntax/stx racket/match racket/unsafe/undefined
          (typecheck signatures tc-funapp)
          (types abbrev prop-ops utils match-expanders)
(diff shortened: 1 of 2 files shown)


typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-eq.rkt
```diff
@@ -63,6 +64,23 @@
      (ret -Boolean (-PS (-is-type o (-val val)) (-not-type o (-val val))))]
     [((tc-result1: (Val-able: (? ok? val))) (tc-result1: t _ o))
      (ret -Boolean (-PS (-is-type o (-val val)) (-not-type o (-val val))))]
+
+    ;; In this case, try to find there is an overlap between t1 and t2
+    [((tc-result1: t1 _ o1) (tc-result1: t2 _ o2))
+     (=> fail)
+
+     (when (ormap (lambda (a) (id=? comparator a))
+                  (list #'member #'memv #'memq))
```
samth, line 73: Are these supposed to be `equal?` etc?

capfredf: My intention was to not deal with `member` `memv` `memq` for lists in this case, but to handle other binary comparators.

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-eq.rkt
```diff
@@ -63,6 +64,23 @@
      (ret -Boolean (-PS (-is-type o (-val val)) (-not-type o (-val val))))]
     [((tc-result1: (Val-able: (? ok? val))) (tc-result1: t _ o))
      (ret -Boolean (-PS (-is-type o (-val val)) (-not-type o (-val val))))]
+
+    ;; In this case, try to find there is an overlap between t1 and t2
+    [((tc-result1: t1 _ o1) (tc-result1: t2 _ o2))
```
capfredf, line 69: use app + #:when

samth (approved):

typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-eq.rkt
capfredf: okay, in (app expr pattern), the expr can't reference any variables bound in preceding patterns.

PR #1111, opened 2021-06-11
Lazerbeak12345 add tab panel aguments

Lazerbeak12345:
A while ago (in pr #1075)  I added the `on-reorder` and `on-close-request` functions to the `Tab-Panel%` type, but I at the time didn't know how `augment` worked so I didn't actually add them correctly. This should fix this problem and allow these events to be augmented.


capfredf:
can you add a test for this?

Lazerbeak12345:
Yeah sure. Where would that test go?

capfredf:
you can put your test file under `typed-racket-test/succeed/`

Lazerbeak12345:
I noticed that there is also `typed-racket-test/gui/succeed`. As this is a test of something in `typed/racket/gui` should this test I'll write actually go there?

capfredf:
Ah, yes, that is indeed better

Lazerbeak12345:
It's possible that I didn't write the test correctly, but I'm pretty sure that this should be accurate, or at least very close.

capfredf:
LGTM

capfredf (approved):

Lazerbeak12345:
Sorry about that wait, but I'm 100% sure if that test worked then we're good. :)

typed-racket-test/gui/succeed/test-tab-augments.rkt
```diff
@@ -0,0 +1,26 @@
...
+      (on-close-request-callback index))))
+(define tabs : (Instance Tab-Panel-With-Callbacks%)
+  (new tab-panel-with-callbacks%
+       [choices '("one" "two" "three")]
+       [parent frame]
+       [style '(can-reorder can-close no-border)]
+       [on-reorder-callback (lambda (tab-list)
+                              (displayln (format "on-reorder ~v" tab-list)))]
+       [on-close-request-callback
+        (lambda (tabs)
```
Lazerbeak12345, line 25: This is a bit of a nitpick as this variable name probably doesn't matter that much, but should I change this to something more correct, such as `index`?

capfredf: Thanks, that's a good idea! I will change it in my next pull request.

capfredf: Fixed it in #1113

PR #1114, opened 2021-06-15
Remove requirers

capfredf:


samth (commented):
Yay!

capfredf:
This pr is blocked by racket/rackunit#143

samth:
I think we should first remove all the uses, but keep the definitions, because other people (besides the two you found) might use them. Then we can try to find them and fix them gradually.

PR #1116, opened 2021-06-16
make the test `gui-contract.rkt` able to exit when running on mac

capfredf:
see https://github.com/racket/racket/issues/3873 for matthew's detailed explanation


capfredf:
cc @samth

PR #1117, opened 2021-06-17
use existential type results for method extraction

capfredf:
In this PR, existential type results are added to better annotate method extraction.
a property accessor previously annotated `(Some (x) (-> (Has-Struct-Property prop) (-> x Number) : x)` is now annotated `(-> (Has-Struct-Property prop) (Some (-> x Number) : #:+ x))`, i.e. the existential quantification has been pushed downward.

## Notes on implementation
-  An existential type `Result` are `Result` with at least one existential quantifier. It has a constructor and destructor, and there is a function `instantiate-result` used to turn de brunji indexes to names in an existential result when the type checker is checking application of a single arrow function. 
- `subresult*` was changed to reason about subtyping between a (-> Self ....) and a (Some (x) (-> x ....) : x), and two existential type results.
- added some helper functions like `abstract/instantiate-propset` and some tests
- relevant parts of documentations were updated. 
- refactored `type-contract.rkt` and `tc-funapp.rkt`
- The changes are almost backward-compatible with old `Some` types, except one point: In the example shown above, previously `x` in proposition position (`: x`), was symmetricial, whereas it is now symmetricial (`: #: x`). Thus how structure properties definition is checked has been [adjusted](https://github.com/racket/typed-racket/pull/1117/files#diff-516cab1093154106286b0cff961a6e54c2468f64a452f70f4bb8a47b73e82df3R583). If this change broke existing packages, we could either fix those packages or make `tc/make-struct-type-property` account for both old `some` types and new existential type results.
- fixed several bugs

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
@@ -993,7 +993,7 @@ prefab types with the (implicitly quoted) prefab-key
 
   @ex[(: prop:foo (Struct-Property (-> Self Number)))
       (: foo-pred (-> Any Boolean : (Has-Struct-Property prop:foo)))
-      (: foo-accessor (Some (X) (-> (Has-Struct-Property prop:foo) (-> X Number) : X)))
+      (: foo-accessor (-> (Has-Struct-Property prop:foo) (Some (X) (-> X Number) : X)))
       (define-values (prop:foo foo-pred foo-accessor) (make-struct-type-property 'foo))
      ]
   @history[#:added "1.10"]}
--- a/typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
@@ -654,6 +654,7 @@ functions and continuation mark functions.
                      mandatory-kw
                      opt-kw]
                 [rng type
+                     (code:line (Some (a ...) type : #:+ proposition))
                      (Values type ...)]
                 [mandatory-kw (code:line keyword type)]
                 [opt-kw [keyword type]]
@@ -746,6 +747,14 @@ functions and continuation mark functions.
 
   @ex[(: add2 (Number -> Number))
       (define (add2 n) (+ n 2))]
+
+  @margin-note{Currently, because explicit packing operations for existential types are
+  not supported, existential type results are only used to annotate accessors
+  for @racket[Struct-Property]}
+  @emph{(Some (a ...) type : #:+ proposition)} for @racket[rng] specifies an
+  @deftech[#:key "Some"]{existential type result}, where the type variables @racket[a ...] may appear
+  in @racket[type] and @racket[opt-proposition]. Unpacking the existential type
+  result is done automatically while checking application of the function.
 }
 
 @;; This is a trick to get a reference to ->* in another manual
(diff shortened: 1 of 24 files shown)


typed-racket-doc/typed-racket/scribblings/reference/types.scrbl
```diff
@@ -746,6 +747,15 @@ functions and continuation mark functions.
 
   @ex[(: add2 (Number -> Number))
       (define (add2 n) (+ n 2))]
+
+  @racket[(Some (a ...) type : opt-proposition)] for @racket[rng] specifies an
+  existential type result, where the type variables @racket[a ...] may appear
+  in @racket[type] and @racket[opt-proposition]. Unpacking the existential type
+  result is done automatically while checking application of the function. Note
```
capfredf, line 754: use marginal-note

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -1434,7 +1437,40 @@
...
 
-
+(define/cond-contract (abstract-type-in-prop initial names-to-abstract)
+  (-> Prop? (or/c symbol? (listof symbol?)) Prop?)
+  (match initial
+    [(? TypeProp? prop) (TypeProp-update prop type
+                                         (lambda (old-ty)
+                                           (abstract-type old-ty names-to-abstract)))]
+    [tp #:when (or (equal? -tt tp) (equal? -ff tp)) tp]
+  ;; TODO finish the function for other types of props
```
samth, line 1447: What is missing here?

typed-racket-lib/typed-racket/rep/type-rep.rkt
```diff
@@ -1434,7 +1437,40 @@
...
+                            (abstract-type-in-prop old-p- names-to-abstract)))))
+
+(define/cond-contract (instantiate-type-in-prop initial images)
+  (-> Prop? (or/c Type? (listof Type?)) Prop?)
+  (match initial
+    [(? TypeProp? prop) (TypeProp-update prop type
+                                         (lambda (old-ty)
+                                           (instantiate-type old-ty images)))]
+    [tp #:when (or (equal? -tt tp) (equal? -ff tp)) tp]
+    ;; TODO finish the function for other types of props
```
samth, line 1465: And here.

samth (approved):

typed-racket-lib/typed-racket/rep/type-rep.rkt
capfredf: Other prop reps like `AndProp` and `OrProp`, but I am planning to do that in future PRs, because dealing with those kinds of props was not a priority in this PR.  

This function and the one above are more of stopgaps, because the way `abstract-*/instantiate-*` is currently organized is okay but could be improved. For example, `instantiate-obj` and `abstract-obj` are put in `type-rep.rkt`. I think it would be more natural to put them in `object-rep.rkt`. Apparently there would be some dependency issues(e.g, `abstract-obj` needs to know some type reps, but `type-rep.rkt` also requires `object-rep.rkt`) , which could be solved by `lazy-require`. But my sense is we could doing instantiation or abstraction via generic interfaces given that the current implementation of Rep structures.

typed-racket-lib/typed-racket/rep/type-rep.rkt
capfredf: see above

PR #1119, opened 2021-06-28
Add `index-where` from `racket/list`.

### LINKED ISSUE #1118: index-where missing from racket/list

8.1

### What program did you run?

````
#lang typed/racket/base
(require racket/list)
(index-where '("a" "b") (lambda (x) (and (string? x) (string=? x "b"))))
````

Either a type error should have been omitted, or the intended computation should have produced the expected answer `1`, as it does in `#lang racket/base`.

````
busted.rkt:3:1: Type Checker: missing type for identifier;
  identifier: index-where
  from module: racket/list
  in: index-where
  ...
````

### PR DESCRIPTION
**Pull request by jessealama**
closes #1118

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -696,11 +696,11 @@
                  (cl->* ((-lst a) Univ . -> . (-opt -Index))
                         ((-lst a) b (-> a b Univ)
                            . -> . (-opt -Index))))]
-[index-where (-poly (a b)
-                    (Univ (-lst a)
-                          (-> a Univ)
-                          . -> .
-                          (-opt -Index)))]
+[index-where (-poly (a)
+                    ((-lst a)
+                     (-> a Univ)
+                     . -> .
+                     (-opt -Index)))]
 
 [list? (make-pred-ty (-lst Univ))]
 [list (-poly (a) (->* '() a (-lst a)))]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: typed-racket-lib/typed-racket/base-env/base-env.rkt [hunk 1]
```diff
@@ -696,6 +696,11 @@
                  (cl->* ((-lst a) Univ . -> . (-opt -Index))
                         ((-lst a) b (-> a b Univ)
                            . -> . (-opt -Index))))]
+[index-where (-poly (a b)
+                    (Univ (-lst a)
```
samth, line 700: What is the first argument of type Univ for?

typed-racket-lib/typed-racket/base-env/base-env.rkt
```diff
@@ -696,6 +696,11 @@
                  (cl->* ((-lst a) Univ . -> . (-opt -Index))
                         ((-lst a) b (-> a b Univ)
                            . -> . (-opt -Index))))]
+[index-where (-poly (a b)
```
capfredf, line 699: looks like we don't need `b` in the binder

typed-racket-lib/typed-racket/base-env/base-env.rkt
jessealama: oops, not necessary

typed-racket-lib/typed-racket/base-env/base-env.rkt
jessealama: right, thanks -- dropped

samth:
Thanks

PR #1126, opened 2021-07-15
fix contract for 2 numeric scs

bennn:
swap the nonnegative? and nonpositive? predicates


samth (approved):

typed-racket-test/unit-tests/contract-tests.rkt
```diff
@@ -925,4 +925,5 @@
                (make-channel)
                #:typed
                #:msg "higher-order value passed as `Any`")
+   (t-int -NonNegInexactReal void 2.0 #:untyped)
```
capfredf, line 928: do we need a case for `Nonpositive-Single-Flonum`, something like `(t-int -NonNegFlonum void -42.42 #:untyped)`?

bennn: I didn't think it was worth the trouble of trying to make a single flonum. (Check the platform, set `(read-single-flonum #t)`, proceed)

Issue #1135, opened 2021-08-15
`(MPairof String Null)` should not be considered a subtype of `(MListof String)`


NoahStoryM:
### What version of Racket are you using?
`v8.2`

### What program did you run?
Program 1:
```
#lang typed/racket

(: m0 (MPairof String Null))
(define m0 (mcons "world" '()))

(: m1 (MPairof Number (MPairof Symbol (MPairof String Null))))
(define m1 (mcons 123 (mcons 'hello m0)))


(displayln m0)
(displayln m1)
```

Program 2:
```
#lang typed/racket

(: m0 (MPairof String Null))
(define m0 (mcons "world" '()))

(: m1 (MPairof Symbol (MPairof String Null)))
(define m1 (mcons 'hello m0))

(: m2 (MPairof Symbol (MPairof String Null)))
(define m2 (mcons 'hello (mcons "world" '())))

(: m3 (MPairof Symbol (MPairof String Null)))
(define m3 (mcons 'hello (mcons "world" (ann '() Null))))

(: m4 (MPairof Symbol (MPairof String Null)))
(define m4 (mcons 'hello (ann (mcons "world" '()) (MPairof String Null))))


(displayln m0)
(displayln m1)
(displayln m2)
(displayln m3)
```

### What should have happened?
Printing values. 

### If you got an error message, please include it here.
Program 1:
```
Type Checker: Polymorphic function `mcons' could not be applied to arguments:
Argument 1:
  Expected: a
  Given:    Positive-Byte
Argument 2:
  Expected: b
  Given:    (MPairof Symbol (MListof String))

Result type:     (MPairof a b)
Expected result: (MPairof Number (MPairof Symbol (MPairof String Null)))

  in: (mcons 123 (mcons (quote hello) m0))
```

`(MPairof String Null)` (`m0`'s type) is regarded as a subtype of `(MListof String)`.
```
> (: m0 (MPairof String Null))
> (define m0 (mcons "world" '()))
> m0
- : (MListof String) [more precisely: (MPairof String Null)]
(mcons "world" '())

> (: m1 (MListof String))
> (define m1 (mcons "world" '()))
> m1
- : (MListof String)
(mcons "world" '())

> (: m2 (MListof String))
> (define m2 (mcons "world" '()))
> m2
- : (MListof String)
(mcons "world" '())

> (set-mcdr! m1 m2)
> m1
- : (MListof String)
(mcons "world" (mcons "world" '()))

> (set-mcdr! m0 m2)
[...]

PR #1139, opened 2021-08-21
fix variance of mpair

capfredf:
Since Null is a subtype of all types of sequences (including mutable lists), it
should be treated differently in subtyping comparison between mpairs.

related to #1135

### DIFF
--- a/typed-racket-test/unit-tests/subtype-tests.rkt
+++ b/typed-racket-test/unit-tests/subtype-tests.rkt
@@ -253,9 +253,13 @@
     [FAIL (-pair -String (-lst -Symbol)) (-seq -String)]
     [FAIL (-pair -String (-vec -String)) (-seq -String)]
     [-Null (-seq -String)]
+    [-Null (-mlst -String)]
     [(-mpair -String -Null) (-seq -String)]
+    [(-mpair -String -Null) (-mlst -String)]
+    [(-mpair -Null -Null) (-mpair (-seq -String) (-seq -String))]
     [(-mlst -String) (-seq -String)]
     [(-mpair -String (-mlst -String)) (-seq -String)]
+    [FAIL (-mpair (-seq -String) (-seq -String)) (-mpair -Null -Null)]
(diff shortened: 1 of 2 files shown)


typed-racket-test/unit-tests/subtype-tests.rkt
```diff
@@ -253,9 +253,13 @@
     [FAIL (-pair -String (-lst -Symbol)) (-seq -String)]
     [FAIL (-pair -String (-vec -String)) (-seq -String)]
     [-Null (-seq -String)]
+    [-Null (-mlst -String)]
     [(-mpair -String -Null) (-seq -String)]
+    [(-mpair -String -Null) (-mlst -String)]
```
samth, line 258: This subtyping is incorrect -- you could mutate the cdr to be a list and then discover that a value of type Null was actually a pair.

capfredf: Did you mean something like this:
```
(: m0 (MPairof String Null))
(define m0 (mcons "hello" '()))
(set-mcdr! m0 (mcons "hello" null))
((inst set-mcdr! String Null) m0 (mcons "Hello" null))
```
Both `set-mcdr! ` expressions are rejected by TR with/without this patch.

My changes is intended to ensure a sequence can be substituted with a null, but not the other way around.

AlexKnauth: What about
```racket
(: m0 (MPairof String Null))
(define m0 (mcons "hello" '()))
((inst set-mcdr! String (MListof String)) m0 (mcons "Hello" null))
```
?
This program should not typecheck, and in master it correctly fails with an error message that `(MPairof String Null)` is not a subtype of `(MListof String)`, and that is how it should be.

`Null` can be a subtype of `(MListof String)`, and it already is, but `MPairof` still needs to be invariant without any `Null` special case.

capfredf: Yes, you are right. M* should always be invariant.

PR #1140, opened 2021-08-27
fix and refactor print type result at top level

capfredf:
1. only subtypes of Number are generalized at top level
2. move the code to tc-result

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/toplevel-trampoline.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/toplevel-trampoline.rkt
@@ -8,8 +8,7 @@
 ;; take over and keep head local-expanding until `begin` forms are exhausted,
 ;; at which point the syntax can be fully-expanded and checked normally.
 
-(require "../utils/utils.rkt"
-         syntax/parse
+(require syntax/parse
          "../private/syntax-properties.rkt"
          (for-template racket/base))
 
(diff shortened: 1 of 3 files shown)


typed-racket-lib/typed-racket/typecheck/toplevel-trampoline.rkt
```diff
@@ -157,28 +158,37 @@
...
-                           (format "\n[more precisely: ~a]"
-                                   (pretty-format-rep (make-Values (map -result tcs))
-                                                       #:indent 17))]
-                          [else (format " [more precisely: ~a]" (cons 'Values tcs))])
+                          [supertype?
+                           (cond
+                             [indented?
+                              (format "\n[more precisely: ~a]"
+                                      (pretty-format-rep (make-Values (map -result tcs))
+                                                         #:indent 17))]
```
samth, line 189: Can 17 be calculated?

typed-racket-lib/typed-racket/typecheck/toplevel-trampoline.rkt
```diff
@@ -157,28 +158,37 @@
             ;; before printing them.
             (define tc (cleanup-type t))
             (define tg (generalize tc))
```
samth, line 160: Should these happen in the other order?

typed-racket-test/unit-tests/interactive-tests.rkt
```diff
@@ -163,6 +163,18 @@
...
       (:print-type (+ 1 1)))
+
+    (test-form #rx"more precisely: Positive-Byte"
+               10)
+
+    (test-form #rx".*(?!more precisely)"
+               (begin (: a-pair (MPairof String Null))
+                      (define a-pair (mcons "123456" '()))
+                      a-pair))
+    (test-form #rx".*\\)\\s*(?!\\[more precisely)"
```
samth, line 174: Probably this test should specifically rule out mentioning `Mutable-Vectorof`.

typed-racket-lib/typed-racket/typecheck/toplevel-trampoline.rkt
capfredf: `cleanup-type` is intended for function types and generalize doesn't deal with function types, so I don't think the order matters in this case.

samth (approved):

PR #1147, opened 2021-10-07
test/fail: correct 'exn:pred' to 'exn-pred'

bennn:


samth:
Wow that seems like a silly mistake now.

PR #1150, opened 2021-10-10
Update make-readtable type

eutro:
This PR fixes the type of `make-readtable`, to match more closely the contract in untyped Racket.

[Untyped `make-readtable`](https://docs.racket-lang.org/reference/readtables.html#%28def._%28%28quote._~23~25kernel%29._make-readtable%29%29) can take `#false` as its readtable parameter, which is currently not permitted by the typed version.

This causes issues when trying to use `make-readtable` with `(current-readtable)`, which has a type of `(U False Read-Table)` and is thus rejected by the type checker.

Even the contract currently in this PR is incorrect, since it doesn't reflect the variadic nature of `make-readtable`.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -2352,9 +2352,9 @@
       (make-Rest
        (list (-opt -Char)
              (Un (one-of/c 'terminating-macro 'non-terminating-macro 'dispatch-macro) -Char)
-             (-opt (Un (one-of/c (-> -Char -Input-Port (-opt -PosInt) (-opt -Nat)
-                                     (-opt -PosInt) (-opt -Nat) Univ)
-                                 -Read-Table)))))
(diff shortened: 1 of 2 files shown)


eutro:
Upon further inspection, there's even more issues than just that. The most appropriate type, I believe, would be
```racket
- : (->* ((U False Read-Table))
         #:rest-star
         ((U Char False)
          (U 'dispatch-macro 'non-terminating-macro 'terminating-macro Char)
          (U (-> Char Input-Port (U Exact-Positive-Integer False) (U Exact-Nonnegative-Integer False) (U Exact-Positive-Integer False) (U Exact-Nonnegative-Integer False) Any)
             False
             Read-Table))
         Read-Table)
```
since this matches most closely the contract specified in the docs for untyped `make-readtable`.

To contrast with the original type (before this PR), the new type is significantly more permissive, allowing
- False as the first argument
- Any number of arguments one more than a multiple of three (rather than just 1, 4 or 7 arguments as before)
- A False or Read-Table "action" argument.

PR #1156, opened 2021-10-29
add flag to disable static contract optimization

bennn:
Add a switch to turn off contract optimization.

(SCV-CR needs to disable the optimizations to do its analysis, and people like Shriram might want to show the full contracts for teaching)

This switch _could_ piggyback on the `PLT_TR_NO_OPTIMIZE` / `#:no-optimize` flag, but then code that needs to disable those expression-level optimizations for safety is going to be slower than it needs to be at boundaries. So, it seems better to make a new switch.

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/optimization.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/optimization.scrbl
@@ -27,3 +27,13 @@ is insufficiently powerful to access @racketmodname[racket/unsafe/ops],
 for example when executing in a sandbox (see @secref["Sandboxed_Evaluation"
 #:doc '(lib "scribblings/reference/reference.scrbl")]). This prevents untrusted
 code from accessing these operations by exploiting errors in the type system.
+
+
+@section{Contract Optimization}
+
+Typed Racket generates contracts for its exports to protect them against
+untyped code.
+By default, these contracts do not check that typed code obeys the types.
+If you want to generate contracts that check both sides equally (for analysis,
+for teaching, etc.) then set the environment variable
+@envvar{PLT_TR_NO_CONTRACT_OPTIMIZE} to any value and recompile.
--- a/typed-racket-lib/typed-racket/static-contracts/instantiate.rkt
+++ b/typed-racket-lib/typed-racket/static-contracts/instantiate.rkt
@@ -40,6 +40,8 @@
            compute-recursive-kinds
            instantiate/inner))
 
+(define no-optimize-sc? (and (getenv "PLT_TR_NO_CONTRACT_OPTIMIZE") #t))
+
 (define (instantiate/optimize sc fail [kind 'impersonator] #:cache [cache #f] #:trusted-positive [trusted-positive #f] #:trusted-negative [trusted-negative #f])
   (define recursive-kinds
     (with-handlers [(exn:fail:constraint-failure?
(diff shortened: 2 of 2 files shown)


typed-racket-doc/typed-racket/scribblings/reference/optimization.scrbl
```diff
@@ -27,3 +27,13 @@ is insufficiently powerful to access @racketmodname[racket/unsafe/ops],
...
+
+
+@section{Contract Optimization}
+
+Typed Racket generates contracts for its exports to protect them against
+untyped code.
+By default, these contracts do not check that typed code obeys the types.
+If you want to generate contracts that check both sides equally (for analysis,
+for teaching, etc.) then set the environment variable
+@envvar{PLT_TR_NO_CONTRACT_OPTIMIZE} to any value and recompile.
```
capfredf, line 39: I think we might need to be more specific here. Simply recompiling the project that depends on TR won't take effect, as TR needs to be recompiled as well. IOW, i.e. `raco setup typed-racket typed <your-collection>`.

bennn: Why is this comment hidden? Should I reword things or is it ok?

I agree that it'd be good to be more specific ... we could easily say `recompile both Typed Racket and your code` ... but all environment variables have the same problem and maybe this page of the docs needs a bigger change.

capfredf: I wanted to run some experiment to further verify my comment so I hid it. And I agree that this is a common problem regarding environment variables.  maybe we need to have an entry (e.g. via deftech) for `recompile`.

bennn:
Just realized, the [GTP benchmarks[(https://docs.racket-lang.org/gtp-benchmarks/index.html) could use this flag to test absolute worst case performance.

(Absolute worst case = no optimize + every boundary has full contracts => overapprox of what boundary checks happen in the real typed/untyped configurations.)

Rebased. Let's see if the tests pass.

bennn:
@jackfirth any idea why resyntax is upset with the changes to [typed-racket-doc/typed-racket/scribblings/reference/optimization.scrbl](https://github.com/racket/typed-racket/pull/1156/files#diff-b598309bfddffcedb159fac50a63e8648d93bfea8d75c2633cd6151cbdea0caa)?
```
Run xvfb-run racket -l- resyntax/cli analyze --local-git-repository . "origin/${GITHUB_BASE_REF}" --output-as-github-review --output-to-file ./resyntax-review.json
lex-line: contract violation
  expected: a git file name line (starting with '+++ b/') or a hunk range line (starting with '@@')
  given: "@@ -29,0 +30,10 @@ code from accessing these operations by exploiting errors in the type system."
  context...:
   /home/runner/.local/share/racket/snapshot/pkgs/resyntax/private/git.rkt:2[7](https://github.com/racket/typed-racket/actions/runs/4255739023/jobs/7403736358#step:6:8):0: git-diff-modified-lines
  ....
```

<https://github.com/racket/typed-racket/actions/runs/4255739023/jobs/7403736358>

jackfirth:
There's something wrong with the regexes in [here](https://github.com/jackfirth/resyntax/blob/master/private/git.rkt#L43) but I don't have time to dig in to it more right now.

jackfirth:
Figured out the issue. Sometimes, git will try to guess the name of the function that a hunk is in. When it does, the hunk header will look like `@@ -4,6 +5,10 @@ some_function_name` instead of `@@ -4,6 +5,10 @@`. The regexes in that `lex-line` function assume that there's nothing after the second `@@`. And scribble files look close enough to C files (if you squint while inebriated) that git tries to guess function names.

jackfirth:
@bennn I fixed the issue in https://github.com/jackfirth/resyntax/pull/172. Once the package server updates and the resyntax job in this repository is downloading and installing the latest version of resyntax, the failure should go away.

bennn:
Wow ... git never ceases to amaze. Thanks!!

PR #1163, opened 2021-11-08
Document: Fix the definition of `defmodule/incl`.

NoahStoryM:
In the old definition, if we search for `JSExpr` in the `search manuals`, we will get: 
`JSExpr  provided from typed/racket/base, typed/racket`

instead of
`JSExpr  provided from typed/json`

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/libraries.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/libraries.scrbl
@@ -52,10 +52,10 @@ The following libraries are included with Typed Racket in the
 @racketfont{typed} collection:
 
 @(define-syntax-rule @defmodule/incl[name rest ...]
-   (list
-      (section #:style '(hidden toc-hidden unnumbered)
-               (string-append "Typed for " (symbol->string 'name)))
-      @defmodule[name rest ...]))
+   (begin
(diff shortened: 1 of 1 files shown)


samth:
I don't think this is the right fix -- it's equivalent to just taking out the `section`.

NoahStoryM:
I'm not sure about the purpose of introducing `section` in `defmodule/incl`, but I tried deleting `#:style` in these two definitions to ensure that the section is valid: 

```
@(define-syntax-rule @defmodule/incl[name rest ...]
    (list
     (section (string-append "Typed for "(symbol->string'name)))
     @defmodule[name rest ...]))

```

```
@(define-syntax-rule @defmodule/incl[name rest ...]
    (begin
      (section (string-append "Typed for "(symbol->string'name)))
      @defmodule[name rest ...]))

```

It seems that these two definitions both make the document generate the corresponding section. Have I missed something? I am not very familiar with scribble.

NoahStoryM:
By the way, can we consider adding `@section{Typed for the @racketfont{typed} collection}` and using `(subsection #:style'(hidden) (symbol->string'name))` instead of `(section #:style'(hidden toc-hidden unnumbered))`? I think it might be better to list these typed collections in the catalog, just like we list Types in `Type Reference`.

PR #1184, opened 2022-01-12
split out tests that rely on external libraries (e.g. math)

capfredf:
This change speeds up building typed-racket-test so that we can have fast a
feedback loop from the core test suites.

On my ryzen 5850u pro (8c/16t), the build time reduces from 1m30s to 22s


typed-racket-test/racketcs-eval-server.rkt
```diff
@@ -2,6 +2,7 @@
 
 (require racket/sandbox)
 
+(eprintf "hello world hihihihihihihi ~n ~n")
```
samth, line 5: Should this be removed?

PR #1185, opened 2022-01-14
typed-racket-more/typed: add xml.rkt & xml/path.rkt

xgqt:
Closes: racket#1183
Signed-off-by: Maciej Barć <xgqt@gentoo.org>


xgqt:
Not implemented:
- `xml/plist`
- `Se-Path` should have Keyword only at the and (optionally)
- `Valid-Char` type for entity and XExpr - now as `Positive-Index`
- `syntax:read-xml` & `syntax:read-xml/element`
- xml's `exn`s

typed-racket-more/typed/xml.rkt
```diff
@@ -0,0 +1,102 @@
+#lang typed/racket/base
+
+(provide (all-defined-out))
+
+
+(define-type XExpr-Attribute
+  (List Symbol String)
+  )
```
samth, line 8: Can you fix the `)` on their own lines?

xgqt: sure, I will follow your style, any more formatting fixes?

capfredf:
@xgqt are you planning to add types for what you listed above in this pull request? otherwise, I am happy to merge it.

xgqt:
> @xgqt are you planning to add types for what you listed above in this pull request? otherwise, I am happy to merge it.

I havent had time to resolve it. If you say that it is good as-is then go ahead and merge, maybe it will get refined later.

LiberalArtist:
I wonder if there's a solution to `permissive-xexprs` similar to `json-null` in <https://github.com/racket/typed-racket/pull/771>. (But I've never really experimented with `permissive-xexprs`.)

samth:
I think `permissive-xexprs` is a mistake, and is fundamentally untypable. So I would be fine with a solution that just makes it always `#f` when typed.

xgqt:
@capfredf Could you review?

PR #1187, opened 2022-01-17
support running unit-tests in parallel

capfredf:
This changes enabled test files to run on multiple workers(places). Due to the
overhead of loading typed racket and racket, using all available threads/cores
does not alway yield the best performace. It look like the best number of
workers for parallism is four.

For reference, on my ryzen 5850u pro, the unit tests take 18 seconds to run with
the change. By contrast, 36 seconds were needed to finish the unit tests before.

more fine-grained parallelization, i.e. support for running test-cases/suites in
parallel, should be implemented in rackunit,

### DIFF
--- a/typed-racket-test/main.rkt
+++ b/typed-racket-test/main.rkt
@@ -199,13 +199,19 @@
    suite))
 
 
-(define (go tests) (test/gui tests))
+(define (go tests)
+  (if (null? tests)
+      0
+      (test/gui (make-test-suite "Typed Racket Tests"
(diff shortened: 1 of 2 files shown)


capfredf:
just found out that the github ci has two threads. the unit tests has sped up. Yay

samth:
This feels like it's half of a new test runner that works in parallel. Is that right?

capfredf:
I am not sure about what you mean by "half of a new test runner"

Issue #1192, opened 2022-02-15
Annotating a rest argument on a function with an optional argument is unsound


lexi-lambda:
This program is patently incorrect and should be rejected:

```racket
#lang typed/racket

(define (bad [opt #f] . [vs : (Vectorof Any) *]) : Any
  (vector-ref vs 0))

(bad)
```

However, it is not:

```
$ racket bad.rkt
invalid memory reference.  Some debugging context lost
  context...:
   body of "/tmp/bad.rkt"
```

The bug is in the type assigned to `vs` in the body of `bad`. It should have type `(Listof (Vectorof Any))`, since it’s a rest argument, but in fact it is assigned just `(Vectorof Any)`. This is wildly unsound.

capfredf:
duplicate of #862

PR #1193, opened 2022-02-16
Fix rest args

### LINKED ISSUE #862: Unsound function types with rest and opt arguments
7.4
### What program did you run?
```racket
#lang typed/racket

(define (f [a : Integer 1] b : Integer *)
  (cons a b))

(add1 (cdr (f 1 2)))
```
a type error
```
  add1: contract violation
  expected: number?
  given: '(2)
```

capfredf:
close #862, #1192

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
@@ -96,7 +96,7 @@
 ;;                 and returns the resulting Arrow?.
 ;; arg-names: The identifiers of the positional args
 ;; arg-types: The types of the positional args
-;; rest-arg+type: Either #f for no rest argument or (cons rest-id rest-type)
+;; rest-arg+type: Either #f for no rest argument or (list rest-id rest-type (Listof rest-type))
 ;;                where rest-id is the identifier of the rest arg,
 ;;                and (ListOf rest-type) is the type that identifier would
 ;;                have in the function body
@@ -263,6 +263,7 @@
 (define/cond-contract (tc/opt-lambda-clause arg-list body aux-table)
   (-> (listof identifier?) syntax? free-id-table?
       (listof Arrow?))
+
   ;; arg-types: Listof[Type?]
   (define arg-types
     (for/list ([a (in-list arg-list)])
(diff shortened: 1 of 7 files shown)


samth:
Can you explain more about what's going on here, and why this approach is the right one? What's the expansion of this, and could we just pattern match on that?

capfredf:
For example,
```
```
expands to  
```
(define-values
 (foo)
 (let-values (((foo)
               (lambda (x1 new-rest)
                 (let-values (((x) (if '#f '#f x1)))
                   (let-values (((y) new-rest))
                     (#%app + '10 (#%app list-ref y '0)))))))
   (case-lambda (() (#%app foo '#f null)) ((x1 . y) (#%app foo x1 y)))))

```
TR checks the RHS of the outmost let bindings and ignores its body. 

The problem we are dealing with boils down to two subproblems:
1. In the lambda at the RHS, `y` in the second let binding has a type label--`Integer`, which essentially makes the binding annotated incorrectly. 
2. In the body of the lambda, the `new-rest` should be given the type `Listof Integer` instead of `Integer`. However, the argument types in the resulting arrow type of the lambda needs to remain `Integer`, because TR later converts the arrow type into an arrow type by treating the last argument type of the former as the type for the rest argument of the latter.  

I can't think of a better way to fix the first sub-problem without ignoring the type label. As to the second, I think the current approach (cf. https://github.com/racket/typed-racket/pull/1193/files#diff-627049fc095268de250612cb67386bb0ae63b0e0f56875dc4f846fb8c626a933R283) is right on the target. 

pattern matching on a syntax of this pattern is indeed a good idea, i.e. use this pattern instead of https://github.com/racket/typed-racket/blob/3274d419dd2700cf27fd083e0ab106c7da99f802/typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt#L340. Then I can add another function to check an optional lambda clause with a rest argument and leave the `tc/opt-lambda-clause` unchanged, or flag to `tc/opt-lambda-clause`.

capfredf:
@samth some updates on the implementation of the better approach:

https://github.com/capfredf/typed-racket/commit/5e2d611aebfa6d31fee4f84e193727ae42a4f2c8 enables the type checker to gather the type information for all types parameters and then passes to corresponding parameters of new lambdas in expanded code. It seems this approach indeed more intuitive, code is simpler (e.g.  `opt-unconvert` is no longer needed) and it works fines with optional lambdas. 

However, since `tc/opt-lambda-clause` is used to check lambdas with keywords, we will need to do a similar "preprocessing" for   a keyword lambda in the case where it is handled in `tc-expr-unit.rkt`, but the preprocessing will be more complicated than that for an optional lambda due to the complexity of the expanded form of a keyword lambda.

My sense is you guys chose a `shortcut` to check those two types of lambdas for a reason.  I'm not sure if the better approach would be worth the effort. If not, maybe we should go with this pull request.

typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt
```diff
@@ -276,7 +277,32 @@
...
+
+      ;; we use `arg-types` to make the arrow consistent with the original type
+      ;; annotation, i.e. `new-rest` needs to have the same type as the original
+      ;; function's the rest parameter. In order to check the function body, we
+      ;; duplicate `arg-types` with the last element changed from `ty` to
+      ;; `(Listof ty`)
+      (define arg-types^ (list-update arg-types (sub1 (length
+                                                       arg-types))
+                                      (lambda (ty) (make-Listof ty))))
+      (-Arrow
```
capfredf, line 297: use tc-lambda-body and substitute arg-types^ for the arguments types

PR #1200, opened 2022-03-01
repair apply and checking property values

### LINKED ISSUE #691: Problems with define-struct/exec
As originally reported on the [email list](https://groups.google.com/d/topic/racket-users/7hL-zpOdaT0/discussion), there are some problems using applicable structures created with `define-struct/exec`. For example, the following program fails to typecheck:
```racket
#lang typed/racket

(define-struct/exec adder ([name : Symbol])
  [(λ (this n) (add1 n)) : (-> adder Number Number)])

(define v
  (adder 'v))

(apply v '(41))
```
with the error "Type Checker: Type of argument to apply is not a function type: 
adder". 

I thought that I might need to add a type annotation, but I also am confused about how to write the type of `v` in the above program. If I change the above program to one that does typecheck:
```racket
#lang typed/racket

(define-struct/exec adder ([name : Symbol])
  [(λ (this n) (add1 n)) : (-> adder Number Number)])

(define v
  (adder 'v))

(v 41)
```
the REPL prints the type of v as simply `adder`, but, hovering over the use of `v` in the last line, the tooltip in DrRacket shows the type as `(U (-> adder Number Number) adder)`. 

This seems to be wrong. That function type is the type of the value for `adder`'s prop:procedure, not the type of an `adder` instance when used as a function. Annotating `v` with `(: v (U (-> adder Number Number) adder))` causes the program to fail to typecheck.

But it isn't clear what type should be used to annotate `v`. Writing `(: v (U (-> Number Number) adder))` also causes typechecking to fail, as does experimenting with an intersection type or simply writing `(: v (-> Number Number))`.

capfredf:
closes #691


typed-racket-test/gh-issue-691.rkt
```diff
@@ -0,0 +1,14 @@
+#lang typed/racket/base
+
+(struct adder ([name : Symbol])
+  #:property prop:procedure
+  (λ ([this : adder] [n : Number]) : Number
+    (add1 n)))
+
```
capfredf, line 7: put the original code in the issue

typed-racket-lib/typed-racket/base-env/prims-struct.rkt
```diff
@@ -176,6 +177,25 @@
...
+       (syntax/loc val
+         (-> ty ... ret-ty))]
+      [(ann ((~or lambda λ lambda:) formals . rst) ty)
+       #'ty]
+      [_:exact-nonnegative-integer
+       val]
+      [(struct-field-index fld)
+       #'fld]
+      [_
+       (tc-error/stx val "expected: a nonnegative integer literal or an annotated lambda")])))
```
samth, line 197: Delay this error until `tc-struct` (consider programs in `typed/racket/no-check`).

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -413,6 +409,25 @@
               (~datum prefix-all-defined) (~datum prefix-all-defined-except)
               (~datum expand)))))
 
+;; finish registering struct in two steps with the second one being the return
+;; thunk, which can be invoked on demand.
+(define (register-struct-type-info! form-li)
```
samth, line 414: This could use a signature.

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -413,6 +409,25 @@
...
+  (for ([form (in-list form-li)])
+    (define name (name-of-struct form))
+    (define tvars (type-vars-of-struct form))
+    (register-resolved-type-alias name (make-Name name (length tvars) #t))
+    (register-type-name name)
+    (add-constant-variance! name tvars))
+
+  ;; the resulting thunk finishes the rest work
+  (lambda ()
+    (for*/list ([form (in-list form-li)]
```
samth, line 425: Maybe using promises can avoid the two list iterations?

typed-racket-lib/typed-racket/base-env/prims-struct.rkt
```diff
@@ -176,6 +177,25 @@
                              (second (build-struct-names #'nm.name null #t #t))))
                 . opts))]))
 
+(define-for-syntax (maybe-get-prop-proc-ty prop-names prop-vals)
```
samth, line 180: Add a signature

typed-racket-lib/typed-racket/base-env/prims-struct.rkt
```diff
@@ -176,6 +177,25 @@
...
+  (for/first ([nm (in-list prop-names)]
```
samth, line 181: What if there are two occurrences of `prop:procedure`?

capfredf:
@samth I did some tricks to support lambdas without annotations for prop:procedure.  When the annotation on the first parameter is omitted, its type will be considered the structure type or `(struct-type-name A B ...)` in presence of polymorphic structure types.
For other parameters, `Any` will be used. 

see https://github.com/racket/typed-racket/pull/1200/files#diff-24395f27e353164bc5dc89a7b790a5b540cdb5ea6d392d7b75965f50b77bcc58R97

support for keyword parameters were added. In particular, polymorphic lambda with monomorphic keyword types was enabled by https://github.com/racket/typed-racket/pull/1200/files#diff-380c101f40583beaf0f3ad2820b1e2f9b9d649257da8caf48b323b1fceb25a88R463

the documentation was updated as well.

typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt
```diff
@@ -231,6 +233,19 @@
...
 
+
+      [(define-values
+         (lifted/7 lifted/8 lifted/9 lifted/10 lifted/11)
+         (#%plain-app
+          make-struct-type
+          a
+          b ;; struct:keyword-procedure/arity-error
+          . rst))
+       ;; why does the below not work?
```
capfredf, line 244: fix me!

typed-racket-lib/typed-racket/base-env/prims-struct.rkt
```diff
@@ -176,10 +198,55 @@
...
+                                       #,(default-any (attribute ret-ty.type)))
+                                val)))
+     ;; add `ann` to the property value. This ensures we have an expected type
+     ;; and therefore the annotation on the receiver can be omitted when
+     ;; checking the value in pass2.
+     (values (with-type* val ty-stx) ty-stx)]
+    [(ann _ ty)
+     (values val (st-proc-ty-property #'ty val))]
+    [_
+     (values val val)]))
```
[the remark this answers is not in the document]
capfredf: The problem is that given `tc/structs` is basically the first things happens during the whole type-checking process, the type checker wouldn't be able to know an identifier's type at the point when it is trying to get the type of the property value expression. 

For example, 

```
(: func (-> foo String))
(define (func me)
   "42"))

(struct bar () #:property prop:procedure func))
```

we can even imagine there are possibilities where the type of `func` refers to other user-defined types and/or type constructors.

PR #1215, opened 2022-03-27
Add types for `indexes-of` and `indexes-where`.

NoahStoryM:
`length` just returns the length of a list, it should have nothing to do with the type of the list.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -673,7 +673,7 @@
                (cl->*
                 (->* (list (-pair a (-lst a))) (-lst a) (-pair a (-lst a)))
                 (->* (list) (-lst a) (-lst a))))]
-[length (-poly (a) (-> (-lst a) -Index))]
+[length (-> (-lst Univ) -Index)]
 [memq (-poly (a) (-> Univ (-lst a) (-opt (-ne-lst a))))]
 [memv (-poly (a) (-> Univ (-lst a) (-opt (-ne-lst a))))]
 [memf (-poly (a) ((a . -> . Univ) (-lst a) . -> . (-opt (-ne-lst a))))]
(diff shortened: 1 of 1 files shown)


capfredf:
I would say `(All (a) (-> (Listof a) Index))` is 1. more accurate 2. in line with types for other list functions, though `(-> (Listof Any) Index)` won't be wrong for `length`. As to typed-untyped interaction, I need to sleep on it. By the way, do you have any programs that the polymorphic type gets in your way?

I'll take another look at the types for `indexes-of` and `indexes-where` later this week. Thanks

NoahStoryM:
I think that `length` should be in line with types for length functions.

For example, I want to implement a generic function for finding the length of a data structure:

```
#lang typed/racket


(: make-length (All (A) [-> (pred A) [-> A Index] [-> Any * Index]]))
(define make-length
   (λ (pred len)
     (λ args
       (match args
         [`(,arg) #:when (pred arg) (len arg)]))))


(make-length bytes?  bytes-length)
(make-length string? string-length)
(make-length vector? vector-length)
(make-length list?   length)

```

Although that `(inst length Any)` can be used to solve the problem of this code, but I compared the types of `length` and `vector-length`, I can't get the meaning of defining `length` as a polymorphic type -- After all, it just returns the length of a list, it has nothing to do with the specific type of the list, I think this is also the reason why the argument type of `vector-length` is `VectorTop`.

capfredf:
That seems a valid point. Note that in your example only vector and list are containers. Since length has been assumed to have a polymorhic type, for back -compatibility, I think we should keept it that way. `(All (a) (-> (Listof Any) Index)` looks a bit weird to me
`a` does not appear in the body and the whole type is actually mormorphic.

NoahStoryM:
I also feel that `(All (A) (-> (Listof Any) Index))` looks weird. I thought maybe TR will give warnings for these usages in the future, and eventually TR may no longer be compatible with them. So that `(All (A) (-> (Listof Any) Index))` is a solution that allows me to avoid `(inst length Any)` now.

But it won't be a big problem for me to change the code back if that day comes. I agree to keep the current type of `length`.

samth:
@capfredf are you ok with this now?

capfredf:
lgtm

PR #1219, opened 2022-04-06
Add types for some raise functions.

NoahStoryM:
I am a little confused when I read code in `base-env-indexing-abs.rkt`.

In the definition of `indexing`, a total of four types are used to represent indexes: `-Integer`, `-Nat`, `-Index` and `index-type` (as `-Integer` in `base-env-indexing.rkt`). And many functions actually use `exact-nonnegative-integer?` to check their arguments, but TR use `index-type` (aka `Integer`) instead of `Natural` (or `Index`) to check them.

I'm not sure why these functions involving index are treated separately in `base-env-indexing-abs.rkt`, in order to be consistent with the types of other functions, I use `index-type` to define the types for these `raise` functions.

Will there be any compatibility problem if I use more precise types instead of `index-type` for all functions in `indexing`? And in addition to vector related functions, should I use `Natural` or `Index` to represent index?

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-env-indexing-abs.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env-indexing-abs.rkt
@@ -4,7 +4,8 @@
  (for-template racket/base racket/list racket/unsafe/ops racket/flonum racket/extflonum racket/fixnum)
  "../utils/tc-utils.rkt"
  (rename-in "../types/abbrev.rkt"  [-Boolean B] [-Symbol Sym])
- (rename-in "../types/numeric-tower.rkt" [-Number N]))
+ (rename-in "../types/numeric-tower.rkt" [-Number N])
+ (only-in "base-structs.rkt" -Arity-At-Least))
 
 (provide indexing)
 
@@ -239,7 +240,7 @@
    [flvector (->* (list) -Flonum -FlVector)]
    [make-flvector (cl->* (-> index-type -FlVector)
                          (-> index-type -Flonum -FlVector))]
-   
+
    [shared-flvector (->* (list) -Flonum -FlVector)]
    [make-shared-flvector (cl->* (-> index-type -FlVector)
                                 (-> index-type -Flonum -FlVector))]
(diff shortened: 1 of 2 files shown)


NoahStoryM:
I see, thanks for your reply!

NoahStoryM:
I noticed that `raise` use `(Un -Flat -Exn)` instead of `Univ` as its argument type, is there any special reason here?

samth:
Consider the following program:

```
#lang racket 
(module m1 typed/racket
   (provide f)
   (define (g [x : Integer]) (+ 1 x))
   (define (f) (raise g)))

(require 'm1)
(with-handlers ([void (lambda (v) (v "x"))]) (f))
```

This would, if the type of `raise` allowed it, pass `g` to a context where it was called with the wrong argument.

An alternative fix would be to introduce a contract barrier between raise and handling, as is done with some other continuation operators. That would reject more values, though, since it would have to use `any-wrap/c`.

NoahStoryM:
I see. I wonder if we can add more types (`-Bytes`, `(-ivec -Flat)` and `(-Immutable-HT -Flat -Flat)`) to `-Flat`?

samth:
Yes, I think adding those should be ok, but why are you `raise`ing values that are not exceptions?

NoahStoryM:
Well, I'm not sure if it's a good style:

I want to use `raise` as `return` in C language, and when the function returns a value, output something according to the type of the value. `let/cc` should be able to do something similar, but I think `with-handlers` is more flexible.

sorawee:
What do you mean by flexible? 

`let/cc` is theoretically more expressive than exception, so in this sense it is more "flexible", and Typed Racket supports `let/cc`, too.

NoahStoryM:
For example:
```
#lang racket/base

(define (test0)
  (with-handlers ([string? (λ (res) (displayln "Return String") res)]
                  [pair? (λ (res) (displayln "Return Pair") res)])
    (raise "Hello")))

(define (test1)
  (define res
    (let/cc return
      (return "Hello")))
  (cond [(string? res) (displayln "Return String") res]
        [(pair? res) (displayln "Return Pair") res]))

(displayln (test0))
(displayln (test1))

```

I think `test0` is more readable than `test1`, and `test1` need more Type information to work on TR.

sorawee:
We have macros precisely to help with that readability problem.

```
#lang typed/racket

(define-syntax-rule (with-my-handler ret ([p? c] ...)
                      body ...)
  (let ([v (let/cc ret : Any
             body ...)])
    (cond
      [(p? v) (c v)]
      ...)))


(define (test2)
  (with-my-handler
      return ([string? (λ (res) (displayln "Return String") res)]
              [pair? (λ (res) (displayln "Return Pair") res)])
    (return "Hello")))

(test2)
```

The reason why I dislike the implementation via `with-handlers` is that `with-handlers` can unintentionally catch values that you don't want to catch.

```
;; your code
(with-handlers ([number? ...do something...])
  (a-function-from-another-library))

;; other people's code
(define (a-function-from-another-library)
  (raise 1))
```

When other people `(raise 1)`, they probably didn't intend for your code to catch it. But it does.

With `with-my-handler`, you are the one who introduces the name `return`, so it's impossible for other libraries to hijack into your `return`.

If you don't want to write `return` explicitly in `with-my-handler`, you can do that too. See syntax parameters.

NoahStoryM:
> The reason why I dislike the implementation via with-handlers is that with-handlers can unintentionally catch values that you don't want to catch.

This is a problem to me, thanks for your suggestion!

Issue #1223, opened 2022-04-13
`typed/racket/no-check` does not export `racket/base` at phase level 1.


NoahStoryM:
### What version of Racket are you using?
`v8.4 [cs]`

### What program did you run?
```
#lang typed/racket/no-check

(define-syntax (case-λ stx)
  (syntax-case stx ()
    [(_ [formals body ...] ...)
     #'(case-lambda [formals body ...] ...)]))

```

### What should have happened?
No error.

### If you got an error message, please include it here.
```
draft.rkt:6:3: syntax-case: unbound identifier;
 also, no #%app syntax transformer is bound in the transformer phase
  at: syntax-case
  in: (syntax-case stx () ((_ (formals body ...) ...) (syntax (case-lambda (formals body ...) ...))))
  location...:
   draft.rkt:6:3

```

PR #1224, opened 2022-04-13
Let `typed/racket/no-check` export `racket/base` at phase level 1.

NoahStoryM:
Closes https://github.com/racket/typed-racket/issues/1223.

### DIFF
--- a/typed-racket-compatibility/typed-scheme/no-check.rkt
+++ b/typed-racket-compatibility/typed-scheme/no-check.rkt
@@ -1,5 +1,5 @@
 #lang typed-racket/minimal
 
 (require racket/require typed/private/no-check-helper
-         (subtract-in typed/scheme/base typed/private/no-check-helper))
-(provide (all-from-out typed/scheme/base typed/private/no-check-helper))
+         (subtract-in typed-scheme typed/private/no-check-helper))
+(provide (all-from-out typed-scheme typed/private/no-check-helper))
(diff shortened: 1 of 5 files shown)


samth:
Can you say something (preferably in a comment in the files) about why you need the additional changes beyond the first one?

capfredf:
`racket/base` isn't included at phase level 1 by default, either

NoahStoryM:
On the first commit I simply let `typed/racket/no-check` and `typed/scheme/no-check` provide `(for-syntax (all-from-out racket/base))` and `(for-syntax (all-from-out scheme/base))` respectively, but then I found 2 problems:

1. `typed/scheme/no-check.rkt` needs to require `(for-syntax (all-from-out scheme/base))` first, which conflicts with `typed/racket/no-check` (because `typed/racket/no-check` provides `(for-syntax (all-from-out racket/base))` now).

2. `typed/racket/no-check/base` and `typed/scheme/no-check/base` should not provide the corresponding `base` at phase level 1, so they cannot be implemented by requiring and providing `typed/racket/no-check`.

The simplest and most straightforward solution I can think of is to make these languages implement the corresponding `no-check` version directly through `typed/private/no-check-helper` (like `typed/racket/no-check`), which made additional changes beyond the first one.

samth:
This change broke the "rosetta" (see https://plt.cs.northwestern.edu/pkg-build/server/built/fail/rosetta.txt) because apparently `typed/racket/base/no-check` used to provide the contents of (at least) `racket/list`. 

Here's an example:

```
#lang typed/racket/base/no-check

drop
```

samth:
In fact, I think all of `racket` was provided, or at least `racket/list`, `racket/math`, `racket/function`, and `racket/port`.

PR #1225, opened 2022-04-14
enable proposition propagation when checking module-level exprs

### LINKED ISSUE #782: No type detected after (unless (list? x) (error))
7.0

### What program did you run?
    #lang typed/racket

    (define  x (read))
    (unless (list? x)
      (error "not a list" x))
    (length x) ;<-- type error here

I think it should typecheck, because after the check it's sure that `x` is a list.  

This works if `x` is defined in a `let` or if `x` is the argument of a function, so probably the problem is that `x` is a module variable.

> Type Checker: Polymorphic function `length' could not be applied to arguments:
> Argument 1:
>   Expected: (Listof a)
>   Given:    Any
> 
> Result type:     Index
> Expected result: AnyValues
>  in: (length x)

capfredf:
close #782

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/signatures.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/signatures.rkt
@@ -15,6 +15,10 @@
    [cond-contracted tc-expr/t (syntax? . -> . Type?)]
    [cond-contracted tc-expr/t* (syntax? . -> . (values Type? boolean?))]
    [cond-contracted single-value ((syntax?) ((or/c tc-results/c #f)) . ->* . full-tc-results/c)]
+   [cond-contracted tc-expr-seq (->* ((listof syntax?) (-> syntax? tc-results/c full-tc-results/c))
+                                     (tc-results/c
+                                      #:unreachable-op (-> (listof syntax?) void?))
+                                     full-tc-results/c)]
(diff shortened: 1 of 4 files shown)


typed-racket-lib/typed-racket/typecheck/signatures.rkt
```diff
@@ -15,6 +15,10 @@
    [cond-contracted tc-expr/t (syntax? . -> . Type?)]
    [cond-contracted tc-expr/t* (syntax? . -> . (values Type? boolean?))]
    [cond-contracted single-value ((syntax?) ((or/c tc-results/c #f)) . ->* . full-tc-results/c)]
+   [cond-contracted tc-sequent-exprs (->* ((listof syntax?) (-> syntax? tc-results/c full-tc-results/c))
```
samth, line 18: I think "sequent" is not a good term here; it means something else in logic.

capfredf: I agree. Do you have some better names in mind? Actually, I was scrambling with this name. (At first, I called it `tc-exprs-in-sequence-with-propositions-from-preceding-exprs`)

samth: `tc-expr-sequence`?

PR #1230, opened 2022-04-21
print out slow tests after 1 min

capfredf:


typed-racket-test/send-places.rkt
```diff
@@ -10,8 +10,25 @@
 (define-values (enq-ch deq-ch) (place-channel))
 (define (start-workers)
   (when (places)
-    (for ([i (places)])
-      (start-worker deq-ch i))))
+    (define ws (for/list ([i (places)])
+                 (start-worker deq-ch i)))
+    (place/context me
+      (let loop ([tests (set)])
+        (define val (apply sync/timeout 120 ws))
```
capfredf, line 17: use a parameter for the `timeout` and make a command line option for it

PR #1231, opened 2022-04-24
Mention assert and ann in the docs of cast.

scolobb:
`cast` protects the values by two contracts, which is often too much of a performance penalty.  In most cases, `assert`, occurrence typing, or `ann` should be used.

This PR is a follow-up of [this discussion](https://racket.discourse.group/t/managing-cast-performance-penalty/905).

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
@@ -614,7 +614,13 @@ protect higher-order uses of the value.
 @ex[
 ((cast (lambda ([s : String]) s) (Any -> Any)) "hello")
 (eval:error ((cast (lambda ([s : String]) s) (Any -> Any)) 5))
-]}
+]
+
+@racket[cast] is mainly useful in the cases in which a thorough check is
+required, in particular when using higher-order or mutable values that cannot
+be fully verified by a flat predicate, or when a good error message from the
+contract system is worth the performance penalty.  In most other cases
+occurrence typing or @racket[assert] should be preferred.}
 
 @defform*[[(inst e t ...)
            (inst e t ... t ooo bound)]]{


typed-racket-doc/typed-racket/scribblings/reference/special-forms.scrbl
```diff
@@ -614,7 +614,13 @@ protect higher-order uses of the value.
...
 ((cast (lambda ([s : String]) s) (Any -> Any)) "hello")
 (eval:error ((cast (lambda ([s : String]) s) (Any -> Any)) 5))
-]}
+]
+
+@racket[cast] is mainly useful in the cases in which a thorough check is
+required, in particular when using higher-order or mutable values that cannot
+be fully verified by a flat predicate, or when a good error message from the
+contract system is worth the performance penalty.  In most other cases,
+occurrence typing, @racket[assert], or @racket[ann] should be preferred.}
```
samth, line 623: I find it a bit odd to mention `ann` here, since it's doing something very different. But I'm open to it depending on what you think.

scolobb: I put it in because it was mentioned in the discussion on Racket Discourse :smile: From my experience, `ann` is indeed doing something rather unrelated, so I would rather agree to remove the mention of `ann` here.

jbclements:
I'm not over the moon about the use of the word "thorough" here; the check performed by an assert is no less thorough than that performed by a cast. I see that the commit has already gone in, though, so I'll just post this comment and let it go.

scolobb:
I see @jbclements . I can still force-push a fix removing "thorough" to this branch or open a new PR if you feel that it's a good idea.  (I don't have an opinion.)

alex-hhh:
I also think that the word "thorough" is confusing -- I have no idea what a "thorough check" is, or a "non-thorough check" is.  How about simply stating what the problem is and directing the user to `assert`.  Something like:

> `cast` will wrap the value in a contract which will affect the runtime performance of reading and updating the value.  This is needed when `e` is a complex data type, such as a hash table, however, when the type of the value can be checked using a simple predicate, consider using `assert` instead.

scolobb:
@alex-hhh Thank you for the suggestion!  Is it better for me to make a separate PR or to push an update to this one?

scolobb:
Great, here is the new PR where people can further discuss: https://github.com/racket/typed-racket/pull/1232.

PR #1235, opened 2022-05-05
adjust assertion error message to use ~e

jbclements:
Per discussion on racket discourse, this pull request adjusts the error message associated with an assertion failure to use ~e rather than ~v

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/extra-procs.rkt
+++ b/typed-racket-lib/typed-racket/base-env/extra-procs.rkt
@@ -8,7 +8,7 @@
      #`(let ([val v])
          #,(syntax-property
             (syntax/loc stx
-              (or val (error (format "Assertion failed on ~v" val))))
+              (or val (error (format "Assertion failed on ~e" val))))
             'feature-profile:TR-dynamic-check #t))]
     [(assert v p)
      #`(let ([val  v]
@@ -17,7 +17,7 @@
             (quasisyntax/loc stx
               (if (pred val)
                   val
-                  (error (format "Assertion ~a failed on ~v" pred val))))
+                  (error (format "Assertion ~a failed on ~e" pred val))))
             'feature-profile:TR-dynamic-check #t))]))
 
 (define (defined? v) #t)


typed-racket-lib/typed-racket/base-env/extra-procs.rkt
```diff
@@ -17,7 +17,7 @@
             (quasisyntax/loc stx
               (if (pred val)
                   val
-                  (error (format "Assertion ~a failed on ~v" pred val))))
+                  (error (format "Assertion ~a failed on ~e" pred val))))
```
sorawee, line 20: Perhaps use `~e` for the predicate too? Otherwise, the same problem would occur for

```
#lang typed/racket

(struct x ([y : (Listof String)]) #:transparent
  #:property
  prop:procedure
  (λ (si x)
    #f))

(assert 1 (x (make-list 10000 "ouch")))
```

NoahStoryM: Would it be better if we use `(object-name pred)` here?

NoahStoryM:
I was wondering if we could extend `assert` so that we can verify that the argument doesn't satisfy the constraint?

For example:
```
Welcome to Racket v8.5 [cs].
> (require (for-syntax racket/base))
> (define-syntax (assert stx)
    (syntax-case stx (!)
      [(assert v) #'(assert v (! not))]
      [(assert v (! (! p))) #'(assert v p)]
      [(assert v (! p))
       #`(let ([val v]
               [pred p])
           #,(syntax-property
              (syntax/loc stx
                (if (pred val)
                    (error (format "Assertion (not/c ~a) failed on ~e" (object-name pred) val))
                    val))
              'feature-profile:TR-dynamic-check #t))]
      [(assert v p)
       #`(let ([val  v]
               [pred p])
           #,(syntax-property
              (quasisyntax/loc stx
                (if (pred val)
                    val
                    (error (format "Assertion ~a failed on ~e" (object-name pred) val))))
              'feature-profile:TR-dynamic-check #t))]))
> (assert 123 (! (! (! string?))))
123
> (assert 123 (! number?))
Assertion (not/c number?) failed on 123 [,bt for context]
> (assert #f)
Assertion (not/c not) failed on #f [,bt for context]
> 

```

typed-racket-lib/typed-racket/base-env/extra-procs.rkt
samth: ping @jbclements

capfredf: ping @jbclements

bennn:
Pushed with `~e` for the predicate in https://github.com/racket/typed-racket/commit/4165b6baa0c79e7391d990d76d74c483592145c1

PR #1247, opened 2022-06-20
Also use the correct tech term for byte strings

sorawee:


capfredf:
sam has fixed this issue in https://github.com/racket/typed-racket/commit/9d3264c97aa63b6a7163a219937b88a612add8ab

PR #1249, opened 2022-06-24
Provide everything from `racket` in `typed/racket/base/no-check`.

samth:
Unfortunately this was accidentally done previously, and fixing it
in #1224 unfortunately broke existing code.


NoahStoryM:
Hi @samth and @capfredf , apparently the behavior of `typed/racket/base/no-check` is inconsistent with the description in the documentation, is it necessary to clarify in the TR reference?

samth:
I don't really want to encourage people to rely on this mistake.

PR #1253, opened 2022-07-02
Fix resyntax permissions issue

jackfirth:
The current Resyntax workflow runs on `pull_request`, which gives it a github token with **read-only** access when run on pull requests from forks. Using `pull_request_target` instead gives it a read-write token and allows the action to submit pull request reviews for pull requests from forks.

### DIFF
--- a/.github/workflows/resyntax.yml
+++ b/.github/workflows/resyntax.yml
@@ -1,7 +1,9 @@
 name: Resyntax
 
 on:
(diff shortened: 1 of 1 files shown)


jackfirth:
Update: this didn't work. A better strategy is outlined [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). I'll give that approach a try next.

PR #1257, opened 2022-07-09
Split Resyntax integration into two workflows

jackfirth:
See the comment in `resyntax-analyze.yml` for why.


samth:
Don't you need to use `pull_request_target` somewhere?

jackfirth:
...oops. Yes I do. Will fix that in a sec.

jackfirth:
@samth wait no, actually I don't need to use `pull_request_target` anymore. Instead, I use `workflow_run`. The second workflow runs when the first one finishes, and the first one doesn't need `pull_request_target` because it no longer needs write access to the repository.

PR #1260, opened 2022-07-09
Fix Resyntax workflows

jackfirth:
Made some typos. Unfortunately, it's very hard to detect issues like that before merging pull requests. Here's hoping this doesn't require too much fiddling to get right.


sorawee:
FYI: https://github.com/nektos/act/ allows you to run GHA locally (I've been using it -- works really great). I don't know if it supports auto-reviewing features, but it should be able to do a basic sanity check.

PR #1263, opened 2022-07-13
contract: repair contract generation for type application

capfredf:
fix http://drdr.racket-lang.org/61098/cs/racket/share/pkgs/plot-test/plot/tests/PRs/42.rkt

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -17,6 +17,7 @@
  "../env/type-name-env.rkt"
  "../env/row-constraint-env.rkt"
  "../env/lexical-env.rkt"
+ "../env/type-constr-env.rkt"
 
  "../rep/core-rep.rkt"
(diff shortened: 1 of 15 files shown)


typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -412,17 +415,25 @@
...
-                                     (hash-set recursive-values
-                                               (cons name 'app)
-                                               (recursive-sc-use name*))))
-                             (recursive-sc-use name*))])]
+        (define constr (lookup-type-constructor name))
+        (cond
+          ;; when constr is a built-in or non-recursive user-defined type
+          ;; constructor, don't generate a recursive static contract
+          ;; for the resulting type.
+          [(not (and (user-defined-type-constr? constr)
```
samth, line 423: I would move this expression and the `lookup-type-constructor` call into a function like `simple-type-constructor?`.

samth (approved):

PR #1275, opened 2022-09-03
Fix Resyntax workflows

jackfirth:
There are two issues with the current workflows: the artifact upload step produces malformed JSON, and the artifact download step doesn't work because the action I'm using for it doesn't support downloading artifacts uploaded by other workflows. This pull request should hopefully fix both issues.


jackfirth:
The only remaining issue appears to be a permissions issue because this pull request modifies the Resyntax Review Submission workflow, and when that workflow runs from this fork it doesn't have the write permissions it needs. This should be fixed after I merge, since it only runs without write permissions in pull requests that edit that workflow.

PR #1282, opened 2022-09-10
type-contract: merge vector and hash types earlier

bennn:
Context:
  (Vector T) is implemented as the union of two types. Enforcing the
  type with two contracts is inefficient; we want one merged contract.
  Same for (HashTable K V), which is implemented as three types.

This commit replaces the old static-contract merging with a type merging.

The problem with static contract merging is that it happens after the "typed-side" optimization kicks in --- which meant that two types with equal components could end up as static contracts with unequal components. This was a huge practical problem in the GTP benchmark `dungeon` ... the fix lowers the runtime for one loop in the worst case down from 30 seconds to 1/2 second.

### DIFF
--- a/typed-racket-lib/typed-racket/private/type-contract.rkt
+++ b/typed-racket-lib/typed-racket/private/type-contract.rkt
@@ -500,15 +500,15 @@
             (apply or/sc (append other-scs (map t->sc (nbits->base-types nbits)))))]
        [(? Union? t)
         (match (normalize-type t)
-          [(HashTableTop:)
-           ;; NOTE: this is a special case to make `HashTableTop` produce a flat contract in casts.
-           ;; Without this case:
-           ;; - `HashTableTop` would make a chaperone contract
-           ;; - because `HashTableTop` is a union containing `(Immutable-HashTable Any Any)`
-           ;; - and `Any` makes a chaperone contract
-           (only-untyped hash?/sc)]
           [(Union-all-flat: elems)
-           (apply or/sc (merge-overlapping-scs (map t->sc elems)))]
+           ;; merge hash and vector types if they have equal components
+           (let*-values ([(hash-tys elems) (partition hash-type? elems)]
+                         [(hvec-tys elems) (partition heterogeneous-vector-type? elems)]
+                         [(vec-tys elems) (partition vector-type? elems)]
+                         [(hash-scs) (merge-hash-types hash-tys t->sc (only-untyped hash?/sc))]
+                         [(hvec-scs) (merge-heterogeneous-vector-types hvec-tys t->sc)]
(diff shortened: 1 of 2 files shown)


bennn:
Any suggestions for how to write regression tests? I'm thinking the unit tests for static contracts need a way to check expected output.

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -1435,116 +1435,88 @@
...
+    (list (hash/sc k0 v1))]
+   [(list-no-order
+      (Mutable-HashTable: k0 v0)
+      (Weak-HashTable: k1 v1))
+    #:when (and (equal? k0 k1) (equal? v0 v1))
+    (list (hash/sc k0 v1))]
+   [_
+    (map t->sc tys)]))
+
+(define (merge-vector-types tys t->sc top-sc)
```
capfredf, line 1503: I am not sure I am completely following the reasoning about all `merge-*` operations here. It looks to me that they only generate a merged static contract when the input `tys` is a list of two elements that meet some conditions.  

let's say that `tys` is `(list (Immutable-Vectorof Natural) (Immutable-Vectorof String) (Mutable-Vectorof Natural))`. Should we merge the static contract for  `(Immutable-Vectorof Natural)` and `(Mutable-Vectorof Natural))`.

capfredf:
> Any suggestions for how to write regression tests? I'm thinking the unit tests for static contracts need a way to check expected output.

we could extend `t/sc` in `static-contract-conversion-tests` so it can take an expected value.

typed-racket-lib/typed-racket/private/type-contract.rkt
bennn: That's a bug. These functions need editing.

They'd better be recursive too. If we change your example to `((IV Natural) (IV String) (MV Natural) (MV String))` the end result should be `((v natural) (v string))`

bennn: hooray for `group-by`

typed-racket-lib/typed-racket/private/type-contract.rkt
```diff
@@ -1435,116 +1440,103 @@
...
+
+(define (heterogeneous-vector-type? ty)
+  (match ty
+   [(or (Immutable-HeterogeneousVector: _)
+        (Mutable-HeterogeneousVector: _))
+    #true]
+   [_ #false]))
+
+(define (group-types-by-elems t->elems ts)
+  (group-by t->elems ts non-#f-and-equal?))
```
bennn, line 1469: Here & below, the code is more flexible that it needs to be. It lets `t->elems` return `#f` and calls the normal `(t->sc _)` if that happens.

If we raise an error instead of allowing `#f`, then we can delete this "group-types" function and one argument from `merge-types->scs`

Issue #1285, opened 2022-09-22
Incorrect signature for `extract-field`.


Lazerbeak12345:
### What version of Racket are you using?

racket-8.6.0.12

### What program did you run?

Program A:

```racket
#lang typed/racket/base
(require
  (only-in typed/net/url get-pure-port/headers netscape/string->url make-http-connection)
  (only-in typed/net/head extract-field))
(define theUrl (netscape/string->url "https://example.com"))
(define-values (port headers) (get-pure-port/headers theUrl #:connection (make-http-connection)))
(define location (extract-field #"location" headers))
(displayln location)
```

Program B is the same, but replace the bytestring `#"location"` with the string `"location"`

### What should have happened?

~~A should run, B should error in the exact way it does.~~

A should have a type error, B should run.

### If you got an error message, please include it here.

Error message for A: (should be a type error)

```
. . extract-field: contract violation
  expected: string field for string header
  given: #"location"
  argument position: 1st
  other arguments...:
```

Unwanted error for B

```
. Type Checker: type mismatch
  expected: Bytes
  given: String in: "location"
```

Lazerbeak12345:
Ah. I should also add, this isn't an issue in untyped racket, and that's why I reported it here instead of in the relevant networking package

sorawee:
I agree that this is a bug in Typed Racket, but I don't follow why program A and B should work the way you expect it to be.

In particular, program A results in the same error when I run in untyped Racket.

The issue, as I understand, is that `extract-field` can consume either string or byte, but both arguments must agree in the type. That is, either string + string or bytes + bytes, but not bytes + string or string + bytes.

So I think program B (string + string) should be allowed, and program A (bytes + string) should be disallowed.

Lazerbeak12345:
Ah, you're right. I don't know why I had that misunderstanding. Program B is the one to be desired.

Lazerbeak12345:
Corrected OP

Lazerbeak12345:
Here's the offending line:

https://github.com/racket/typed-racket/blob/4165b6baa0c79e7391d990d76d74c483592145c1/typed-racket-more/typed/net/head.rkt#L8

[Here's the docs for this function](https://docs.racket-lang.org/net/head.html#%28def._%28%28lib._net%2Fhead..rkt%29._extract-field%29%29)

This signature would be optimal, but it it "cannot be converted into a contract" because it "has two cases of arity 2"

```racket
[...]

PR #1287, opened 2022-09-27
Repair `net/head` types.

Lazerbeak12345:
Solution 2 for the previous message.

Fixes up a bunch of the types in this libarary. Just read the code lol.

---

_old message_:

Solution 1 for https://github.com/racket/typed-racket/issues/1285

Upsides:

- Fixes the problem

Downsides:

- `extract-field` remains inconsistent compared to the rest of the `net/head` exports.

### DIFF
--- a/typed-racket-more/typed/net/head.rkt
+++ b/typed-racket-more/typed/net/head.rkt
@@ -1,16 +1,23 @@
 #lang typed/racket/base
 
-(require typed/private/utils)
+(require typed/private/utils typed/racket/unsafe)
 
+(unsafe-require/typed/provide net/head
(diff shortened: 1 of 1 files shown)


samth:
We could just use `unsafe-require/typed` here to avoid the impossible-to-generate contract.

Lazerbeak12345:
Good idea. I'll write that approach when I have time.

While I'm there, I'll see if I can also rectify the rest of the functions.

Lazerbeak12345:
updated OP

Lazerbeak12345:
It's been a few months. Any plans here?

capfredf:
Sorry, I forgot to merge it! LGTM

PR #1288, opened 2022-10-19
Allow #:combine and #:combine/key

bdeket:
for `hash-union`, `hash-union!` and `hash-intersect`

See also: https://racket.discourse.group/t/type-checker-inference-for-polymorphic-keyword-functions-not-supported/145/5

maybe related: following failing test was recently added but I think the first hash should be `#hash()`. However I could be wrong about what is being tested
https://github.com/racket/typed-racket/blob/20acc4bc64907ccf0638c275dccec48461749979/typed-racket-test/fail/hash-union.rkt#L3


capfredf:
The test is to ensure that doing type inference for invoking `hash-union` with keyword arguments raises a type error instead of an internal error.

bdeket:
I just wanted to point out that the current test will always fail, because `hash-union` expects an immutable hash as first argument. which is of course different than an internal error.
Also thank you for running the tests again.

Issue #1293, opened 2022-11-26
Performance regression 8.6 -> 8.7


philnguyen:
### What version of Racket are you using?
8.7

### What program did you run?
Type-checking [philnguyen/soft-contract](https://github.com/philnguyen/soft-contract) from scratch
```
cd soft-contract/soft-contract
raco pkg install
rm compiled */compiled
time raco make -j 8 cmdline.rkt
```

### What should have happened?
Type-checking time should remain similar to 8.6

### If you got an error message, please include it here.
On a computer, with 8.6:
```
real	1m38.201s
user	4m14.191s
sys	0m31.089s
```
with 8.7:
```
real	5m4.759s
user	17m46.285s
sys	0m36.917s
```

While scv itself is dead software that nobody should use, the same regression might happen to other projects. Commit #1dde78d1 looks like the one that introduces the regression.

jbclements:
@bennn wow, that is a pretty big slowdown...

mfelleisen:
Wy are you @jbclements re-directing this to @bennn ? Shallow uses the exact same type-checker as the pre-existing Typed Racket; he wasn't supposed to change the type checker.

jbclements:
@mfelleisen  It appears to me that @philnguyen was suggesting that Ben's Optional & shallow types commit was responsible for this slowdown; I was just tagging him to make it more likely that he saw the post.

bennn:
Thanks for the report & tag. I don't know what this could be but I'll take a look.

bennn:
Some of the cost came from adding a [`PolyRow` case to `types/subtype.rkt`](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/types/subtype.rkt#L1133). We can probably remove that (not ideal); better  yet, maybe a fail-fast path will speed things up.

But that's only 150s of the total slowdown. Still have 600s left to track down.

samth:
We could presumably move that case further down, which should make it happen less.

bennn:
It was all from a `normalize-type` that checked whether Shallow could trust the results of a function.

With that removed, the PolyRow case doesn't make much difference. Keeping as-is vs. removing has about the same cost.

bennn:
https://github.com/racket/typed-racket/pull/1296

PR #1296, opened 2022-12-12
avoid calling normalize-type outside of shallow tr

bennn:
Fix the slowdown reported in https://github.com/racket/typed-racket/issues/1293

Saves about 600 seconds on my machine on this command:
```
rm compiled */compiled
time raco make -j 8 cmdline.rkt
```

- Before = `861s`
- After = `250s`

### DIFF
--- a/typed-racket-lib/typed-racket/utils/shallow-utils.rkt
+++ b/typed-racket-lib/typed-racket/utils/shallow-utils.rkt
@@ -11,14 +11,17 @@
   (submod typed-racket/rep/type-rep shallow-exports)
   typed-racket/types/tc-result
   typed-racket/types/type-table
-  typed-racket/types/union)
+  typed-racket/types/union
+  typed-racket/utils/tc-utils)
 
 
 (define (ArrowT+? arr)
   (and (Arrow? arr) (Arrow-rng-shallow-safe? arr)))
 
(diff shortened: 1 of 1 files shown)


bennn:
This fixes the regression, but there are two reasons we might want a different change:

1. Shallow still calls normalize-type. If soft-contract got ported to Shallow TR, it would be slow to typecheck.
2. Now that Deep and Optional don't call normalize type, Shallow might have a higher runtime cost for interacting with them. Take Optional for example. If Optional sends a function `f` to Shallow, then Shallow has to check that `f` is a function and **maybe** has to check the shape of results `(f x)`. It depends on whether `f` is marked as trusted. Removing the `normalize-type` might remove the mark in some common cases.

Ideally we should either remove `normalize-type` in all cases (because it doesn't save enough runtime checks to justify the compile-time costs) or use it carefully but consistently across Shallow, Deep, and Optional.

samth:
I think the real question is why is normalize-type so slow. Contract generation involves multiple passes over types, so adding this one shouldn't have been an enormous slowdown.

mfelleisen:
I brought this up with Ben before: Asumu spent a lot of time wresting with types because they become very large. Sharing as much as possible solved his problem. Is something similar going on here?

Issue #970, opened 2020-09-08
The `racket' language does not seem to have a type for this identifier; for stream->list


ailrk:
### What version of Racket are you using?
Racket v7.7
### What program did you run?
```racket
(stream->list (in-range 0 3))
```
### What should have happened?
return `'(0 1 2)`

### If you got an error message, please include it here.

```
> (stream->list (in-range 0 3))                                     
; readline-input:14:1: Type Checker: missing type for identifier;
; The `racket' language does not seem to have a type for this identifier;
;   please file a bug report identifier: stream->list
;   from module: (lib typed/racket)
;   in: stream->list
```

jackfirth:
In the meantime, you can use `sequence->list`.

Lazerbeak12345:
This stopgap won't work for a usecase I have in mind, since `(stream? empty-sequence)` returns `#f` and I'm typing something that relies on `racket/stream`. Just a note.

AlexKnauth:
Why wouldn't `sequence->list` work? All streams are sequences, right? So `(stream? empty-sequence)` shouldn't be necessary. `(sequence? empty-stream)` would be necessary instead, and that's fine because it's true.

Lazerbeak12345:
(we actually talked through this on discord. I'm just writing this here so others aren't confused)

The problem was defining a type for a function that takes a stream as an argument. Using the `Sequenceof` type should be sufficient as a workaround in the short-term.

Lazerbeak12345:
Me and several other discord users talked about this in the racket discord. https://discord.com/channels/571040468092321801/993032505127088209/1070776863900844102

I'll just use the `Sequenceof` type in place of `Streamof` for now.

PR #1300, opened 2023-01-28
Add `typed/racket/stream`

Lazerbeak12345:
I'd love feedback. Not finished yet, ofc.

See #970 and perhaps others.


Lazerbeak12345:
Arguably resyntax shouldn't need to run each time I edit the OP

PR #1307, opened 2023-02-06
Make the polymorphic struct error message a bit easier to understand.

Lazerbeak12345:


Lazerbeak12345:
It looks like the error for why these CI tests failed is to do with a 404 when grabbing the racket installer script during setup. Not related to the string I changed.

capfredf:
Thanks!

PR #1322, opened 2023-03-17
repair the type for input-port-append

### LINKED ISSUE #1320: input-port-append doesn't work in TR

8.7 and 8.8 - same outcome on both

### What program did you run?

```
(println (port->string (input-port-append #t (open-input-string "hello") (open-input-string "typed") (open-input-string "racket"))))
```

"hellotypedracket"

```
string:1:23: Type Checker: could not apply function;
 wrong number of arguments provided
  expected: 2
  given: 3
  in: (input-port-append #t (open-input-string "hello") (open-input-string "typed") (open-input-string "racket"))
  location...:
   string:1:23
  context...:
   /home/cadence/.racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-app-helper.rkt:32:0: tc/funapp1
   /home/cadence/.racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-funapp.rkt:71:0: tc/funapp
   .../private/runtime.rkt:80:24: temp676
   .../private/runtime.rkt:89:23: fail-handler114
[stack trace continues...]
```

### Further information

Works in regular racket.

The type in base-env.rkt is `[input-port-append  (->* (list Univ) -Input-Port -Input-Port)]`. I cloned the repo and messed around with the type but I wasn't able to make any progress towards fixing this. As far as I can tell (despite the unusual syntax in this file...) I think that type is already correct.

While the error message says "wrong number of arguments", I can't figure out where I'm calling a function incorrectly. Altering my calls to `input-port-append` and/or `open-input-string` to add or remove more arguments doesn't change the error message. It always says it expected 2, given 3.

capfredf:
closes #1320

### DIFF
--- a/typed-racket-test/succeed/gh-issue-1320.rkt
+++ b/typed-racket-test/succeed/gh-issue-1320.rkt
@@ -1,0 +1,6 @@
+#lang typed/racket/base
+(require racket/port)
+
+(input-port-append #t (open-input-string "hello") (open-input-string "typed") (open-input-string "racket"))
+(input-port-append #t (open-input-string "hello") (open-input-string "typed") (open-input-string "racket")
+                   #:name "hello typed racket")
--- a/typed-racket-lib/typed-racket/base-env/base-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-env.rkt
@@ -2137,7 +2137,7 @@
 [with-input-from-bytes (-poly (a) (-> -Bytes (-> a :T+ #f) a :T+ #f))]
 
 ;; Section 13.1.10.2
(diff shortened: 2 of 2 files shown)


typed-racket-test/succeed/gh-issue-1320.rkt
```diff
@@ -0,0 +1,3 @@
+#lang typed/racket/base
+
+(input-port-append #t (open-input-string "hello") (open-input-string "typed") (open-input-string "racket"))
```
samth, line 3: Maybe add a use of `#:name`?

PR #1325, opened 2023-03-26
test: uncomment math-untyped-id tests

bennn:
Once https://github.com/racket/math/pull/68 is merged, this should be good.

Before that, the CI will fail.


shhyou:
Does this change (tests) need to depend on a new version of the math package?

capfredf (approved):

Issue #1286, opened 2022-09-27
require/untypedcontract


bennn:

Rscho314:
Hi @bennn ,

I have posted some seemingly working code to add a language specification to `require/untyped-contract`. Two questions:

- I guess I should make the language spec argument optional, to avoid breaking existing code?
- Another idea would be fetching the language spec automatically from the imported module. Do you think that's worth pursuing?

Best,
Raoul

bennn:
Yes, language spec should be optional.

Fetching automatically seems too hard. It may have trouble with identifiers that get reprovided by TR languages (Shallow -> Deep -> require/untyped-contract).

bennn:
I'm confused now. Does it even make sense to use `require/untyped-contract` on a shallow module?

(It doesn't make sense for no-check because that language doesn't generate contracts at all.)

Maybe the right thing is just a better error message.

PR #1330, opened 2023-06-11
Patched require/untyped-contract to accept a language spec.

Rscho314:
Hi @bennn ,

This is a very modest proposal for #1286. I did not patch the docs yet, but I will do so if there is interest in this pull request. It seems to work with the following test code (changed language to typed/racket/no-check):

**server.rkt**

```
#lang typed/racket/no-check ;shallow ;optional ;no-check

(provide mynum check-array-shape)

(define mynum 42)

(define-type (MVec T) (Mutable-Vectorof T))

(define-type Indexes (MVec Index))
(define-type In-Indexes (U (MVec Integer) Indexes))

(: check-array-shape (In-Indexes (-> Nothing) -> Indexes))
(define (check-array-shape ds fail)
  (define dims (vector-length ds))
  (define: new-ds : Indexes (make-vector dims 0))
  (let loop ([#{i : Nonnegative-Fixnum} 0])
    (cond [(i . < . dims)
           (define di (vector-ref ds i))
           (cond [(index? di)  (vector-set! new-ds i di)
                               (loop (+ i 1))]
                 [else  (fail)])]
          [else  new-ds])))
```

**main.rkt**

```
#lang racket/base

(require typed/untyped-utils
         typed/racket/no-check
         (except-in "server.rkt"
                    check-array-shape))

(require/untyped-contract
 "server.rkt"
 'typed/racket/no-check
 [check-array-shape  #;((Mutable-Vectorof Integer) (-> Nothing) -> (Mutable-Vectorof Index))
                     ((Vectorof Integer) (-> Nothing) -> (Vectorof Index))])

(check-array-shape (vector 1 2 3) (lambda () (error 'die)))

```
Cheers!

### DIFF
--- a/typed-racket-lib/typed/untyped-utils.rkt
+++ b/typed-racket-lib/typed/untyped-utils.rkt
@@ -29,8 +29,8 @@
   (stx-map (lambda (id) ((make-syntax-introducer) id)) ids))
 
 (define-syntax (require/untyped-contract stx)
-  (syntax-parse stx #:literals (begin)
-    [(_ (begin form ...) from-module-spec:expr [name:id T:expr] ...)
+  (syntax-parse stx #:literals (begin quote)
+    [(_ (begin form ...) from-module-spec:expr (quote language-spec:id) [name:id T:expr] ...)
      (with-syntax* ([(typed-name ...)  (generate-temporaries #'(name ...))]
                     [(untyped-name ...)  (freshen #'(name ...))]
                     [(untyped2-name ...)  (generate-temporaries #'(name ...))]
@@ -39,10 +39,9 @@
                     [typed-module  (generate-temporary #'typed-module)]
                     [untyped-module  (generate-temporary #'untyped-module)]
                     [*racket/base (datum->syntax #'from-module-spec 'racket/base)]
-                    [*typed/racket/base (datum->syntax #'from-module-spec
-                                                       'typed/racket/base)]
-                    [*require (datum->syntax #'from-module-spec
-                                             'require)]
+                    [*typed/racket (datum->syntax #'from-module-spec (format-symbol "~a" (syntax-e #'language-spec)))]
+                    [*require (datum->syntax #'from-module-spec 'require)]
+                    [*language-spec (datum->syntax #'racket/base (format-symbol "~a" (syntax-e #'language-spec)))]
                     [from-module-spec-for-submod
                       (syntax-parse #'from-module-spec #:literals (submod)
                         [(submod (~and base (~or "." "..")) elem ...)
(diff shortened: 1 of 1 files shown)


bennn:
Tried running this on the example, and I'm getting an error about an `(only-in quote)`:

```
only-in: identifier `quote' not included in nested require spec
  at: "server.rkt"
  in: (only-in "server.rkt" (quote quote3) (check-array-shape check-array-shape4))
```

typed-racket-lib/typed/untyped-utils.rkt
```diff
@@ -39,19 +39,18 @@
                     [typed-module  (generate-temporary #'typed-module)]
                     [untyped-module  (generate-temporary #'untyped-module)]
                     [*racket/base (datum->syntax #'from-module-spec 'racket/base)]
-                    [*typed/racket/base (datum->syntax #'from-module-spec
-                                                       'typed/racket/base)]
-                    [*require (datum->syntax #'from-module-spec
-                                             'require)]
+                    [*typed/racket (datum->syntax #'from-module-spec (syntax-e #'language-spec))]
+                    [*require (datum->syntax #'from-module-spec 'require)]
+                    [*language-spec (datum->syntax #'racket/base (syntax-e #'language-spec))]
```
bennn, line 44: This was `typed/racket` before. I think it needs to be the language-spec.

Rscho314: @bennn Thanks for the feedback. I've changed the `datum->syntax` context to `#'_`. I'm not really up to good coding style conventions for Racket, so I hope I've done the right thing.

As for the error about an `(only-in quote)`, I agree the message could be improved, but from a cursory glance at the source, it would seem to require patching racket/collects/racket/private/reqprov.rkt (outside the `typed-racket` package). I'm also quite unsure of how to approach that. Somehow make `require/untyped-contract` a special case ?

Is there anything else I can do to improve this PR ?

bennn: how about changing the syntax from `typed/racket/shallow` to `#:lang typed/racket/shallow` --- from `:id` to keyword +`:expr` ? that should improve the error and make this easier to use later.

typed-racket-lib/typed/untyped-utils.rkt
```diff
@@ -29,46 +29,52 @@
   (stx-map (lambda (id) ((make-syntax-introducer) id)) ids))
 
 (define-syntax (require/untyped-contract stx)
-  (syntax-parse stx #:literals (begin)
-    [(_ (begin form ...) from-module-spec:expr [name:id T:expr] ...)
```
bennn, line 33: keep the indentation like it was

PR #1336, opened 2023-08-08
fix tests for the optimizer

capfredf:


samth:
Please squash and write a commit message that references the relevant racket commit.

PR #1339, opened 2023-10-16
add type for `prgexp-matcher` used internally in `match`

usaoc:
See racket/racket#4711 and racket/racket#4791.


sorawee:
Thanks for catching this!

typed-racket-lib/typed-racket/base-env/base-special-env.rkt
```diff
@@ -387,4 +387,9 @@
   [(make-template-identifier 'start-place 'racket/place)
    (-> -Symbol -Module-Path -Symbol (-opt -Input-Port) (-opt -Output-Port) (-opt -Output-Port) -Place)]
    [(make-template-identifier '-mlist 'compatibility/mlist)
-    (-poly (a) (->* (list) a (-mlst a)))])
+    (-poly (a) (->* (list) a (-mlst a)))]
+  ;; for `match`
+  ;; pregexp-matcher
+  [(make-template-identifier 'pregexp-matcher 'racket/match/runtime)
+   (-> (Un -PRegexp -String) -Symbol (-> (Un -String -Bytes) -Boolean))]
```
samth, line 394: This return type looks wrong; `regexp-match` and thus `pregexp-matcher` can produce more than just booleans.

samth:
Thanks for taking a look at this!

typed-racket-lib/typed-racket/base-env/base-special-env.rkt
usaoc: An update to the type was pushed.

capfredf (approved):

PR #1340, opened 2023-11-01
[Docs] Rename Opt to Maybe in guide

micahcantor:
When I was learning Typed Racket, I was confused by this section of the guide, which gives an example of defining a polymorphic data type. The example used is an optional type called `Opt`, defined analogously to `Maybe` in Haskell. Personally, this example was confusing to me since I then thought the the [Option](https://docs.racket-lang.org/ts-reference/type-ref.html#%28form._%28%28lib._typed-racket%2Fbase-env%2Fbase-types..rkt%29._.Option%29%29) type provided by the standard library was defined in the same way. However, that `(Option t)` type is actually defined as either `t` or `#f`, which aligns better with Racket conventions.

To avoid this confusion, and since `Maybe` from Haskell is already mentioned here, I recommend changing the name of `Opt` to `Maybe` in this example.  I also changed type parameter names to be upper case in this example, for consistency with the rest of the language. 

It also may be helpful to add a note afterward explaining that in a "real" program, for consistency the user should probably use `Option` rather than their own `Maybe` type, but maybe that's unnecessary, so I left it out here.

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/guide/types.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/guide/types.scrbl
@@ -303,9 +303,9 @@ an analog of the @tt{Maybe} type constructor from Haskell:
 @racketmod[
 typed/racket
 (struct Nothing ())
-(struct (a) Some ([v : a]))
+(struct (A) Some ([v : A]))
 
-(define-type (Maybe a) (U Nothing (Some a)))
+(define-type (Maybe A) (U Nothing (Some A)))
 
 (: find (-> Number (Listof Number) (Maybe Number)))
 (define (find v l)
@@ -320,7 +320,7 @@ a structure with no contents.
 The second definition
 
 @racketblock[
-(struct (a) Some ([v : a]))
+(struct (A) Some ([v : A]))
 ]
 
 creates a type constructor, @racket[Some], and defines a namesake structure with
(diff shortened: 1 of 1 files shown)


shhyou:
I think the current name is also a common one (e.g. [`'a option = NONE | SOME of 'a`](https://smlfamily.github.io/Basis/option.html#SIG:OPTION.option:TY) from SML and [`'a option = None | Some of 'a`](https://v2.ocaml.org/api/Option.html) from OCaml). Perhaps the guide could actually refer to SML/OCaml instead of Haskell.

Disambiguating `Opt` from the built-in `Option` does sound like a good idea.

micahcantor:
I agree, it's less that I think the guide needs to adopt Haskell's terminology, and more that as a newcomer to Typed Racket (but having experience with Haskell/OCaml) the guide was confusing since I didn't know if I should use `Option` as defined in the standard library or `Opt` as defined here.

sorawee:
I think we all agree that `Opt` is not a great name, because it's too similar to `Option`, which is a built-in type (that does something slightly different).

So let's use the name `Maybe`, and since you already rename `None` to `Nothing`, _maybe_ (see what I did here?) you want to rename `Some` to `Just` too?

micahcantor:
> _maybe_ (see what I did here?) you want to rename `Some` to `Just` too?

Oops, it's been a little while since I used Haskell instead of OCaml, I did mean that!

sorawee:
Thanks!

PR #1341, opened 2023-11-02
doc: add no_contract flag to index

bennn:


typed-racket-doc/typed-racket/scribblings/reference/optimization.scrbl
```diff
@@ -36,4 +36,5 @@ untyped code.
 By default, these contracts do not check that typed code obeys the types.
 If you want to generate contracts that check both sides equally (for analysis,
 for teaching, etc.) then set the environment variable
-@envvar{PLT_TR_NO_CONTRACT_OPTIMIZE} to any value and recompile.
+@index["PLT_TR_NO_CONTRACT_OPTIMIZE"]{@envvar{PLT_TR_NO_CONTRACT_OPTIMIZE}}
```
samth, line 39: Seems like this should be supported directly in `envvar`.

bennn: by default, or with an optional `#:index?` flag?

samth:
Probably optional because we don't want to index every place PATH is referenced.

samth:
I'm going to merge this, but I do think adding to `envvar` is a good idea.

bennn:
oh, thanks for merging!

I'll try to update envvar this month.

PR #1345, opened 2023-11-05
[Docs] Add section on custom propositions

micahcantor:
Section 5.2 of the guide describes how Typed Racket narrows types with propositions. However, I don't think this section makes it clear if and how users can define their own propositions from predicates. In 5.2, the proposition for the built-in type predicate `string?` is discussed. Then, in 5.2.1, it jumps to defining custom one-sided propositions, a slightly more advanced concept. I think it could use a section in between that makes it explicit how to create a proposition from a predicate.

Personally, I ran into this issue when I was learning Typed Racket: at the time I didn't understand from the guide how to define my own proposition. [I asked a question](https://stackoverflow.com/questions/69694263/how-to-assert-listof-string-predicate-in-typed-racket) on StackOverflow and got an answer, but I think this would be more clear with an example in the guide.

This PR adds such an example. I was a bit confused on whether to use `@examples` or `@racketblock` so I chose the latter which seemed simpler, but feel free to recommend changes to that, to the writing, or more generally to the structure/position of this section.

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/guide/occurrence.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/guide/occurrence.scrbl
@@ -84,6 +84,31 @@ the typechecker learns from the result of applying the function:
 Predicates for all built-in types are annotated with similar propositions
 that allow the type system to reason logically about predicate checks.
 
+@subsection{Custom Propositions}
+
+Although propositions are provided for all built-in type predicates,
+we may want to define propositions from our own predicates as well.
+For instance, consider the following predicate:
+
+@racketblock[
+(: (-> (Listof Any) Boolean))
+(define (listof-string? lst)
+  (andmap string? lst))
+]
+
+This function determines whether a given list contains only strings.
+With Typed Racket, we can use this predicate to refine the type of a given list from
+the general @racket[(Listof Any)] to the more specific @racket[(Listof String)].
+To do so, we must change the type of @racket[listof-string?] to be a proposition:
+
+@racketblock[
+(: (-> (Listof Any) Boolean : (Listof String)))
+]
+
+Here, we adjust the type annotation to include the logical proposition
+@racket[(Listof String)] after the second @racket[_:], from which the typechecker learns
+that this predicate narrows its result type to a list of strings.       
+
 @subsection{One-sided Propositions}
 
 Sometimes, a predicate may provide information when it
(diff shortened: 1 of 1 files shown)


sorawee:
I proposed a suggestion at https://github.com/micahcantor/typed-racket/pull/1.

micahcantor:
Looks good, I think adding that example makes it more clear why we would need to add the proposition

typed-racket-doc/typed-racket/scribblings/guide/occurrence.scrbl
```diff
@@ -84,6 +86,51 @@ the typechecker learns from the result of applying the function:
...
+    (andmap string? lst))
+]
+
+We then may wish to use this predicate to narrow a type in our main program:
+
+@examples[#:label #f #:eval the-eval*
+  (: main (-> (Listof Any) String))
+  (eval:error (define (main lst)
+                (cond
+                  [(listof-string? lst) (first lst)]
```
capfredf, line 110: Perhaps we need to mention `(andmap string? lst)` works here. After all, `listof-string` is just an eta-expansion of (curried) `(andmap string?). 

Also, on a possibly related note,  the reason `(: listof-string? (-> (Listof Any) Boolean))` does not work is that its positive and negative propositions are `always-true`, which is bigger than `lst is (Listof String)` or `lst is not (Listof String)`

micahcantor: To the first point, we could point out that the typechecker can infer proposition types for expressions that return booleans (in the untyped Racket sense).

sorawee: @capfredf does this look better? https://github.com/micahcantor/typed-racket/pull/2.

I actually kinda like the current version. I definitely don't want to get to how `(-> (Listof Any) Boolean)` technically has a proposition, since it's likely confusing to beginners who read it (keep in mind that this is the Typed Racket Guide). 

The narrative in the above PR works, but it kinda takes too long to set things up before getting to the point. Though that may be fine after all.

micahcantor: I think I prefer the current version too, looking at it side by side. Personally, the issue I had lied more in how to annotate the type of a proposition at the top level, and less with understanding that occurrence typing worked on more complex predicates.

capfredf: I think the current version looks good. Just in this particular example, `(andmap string? lst`) works directly. Perhaps, we can simply add something like `if we want to turn `(andmap string? x)` into a custom predicate, here is what we need to do:`

sorawee: I'll work on a margin note as a separate commit to address @capfredf's concern.

samth:
Thanks!

PR #1358, opened 2024-01-25
any-wrap: do best effort to preserve identity

### LINKED ISSUE #1357: `check-eq? (hasheq) (hasheq)` fails
`v8.10 [cs]`

### What program did you run?
```racket
#lang typed/racket/base
(require typed/rackunit)
(check-eq? (hasheq) (hasheq))
```

Test succeeds like with untyped `check-eq?`

```
FAILURE
name:       check-eq?
location:   racket-mode-repl:?:?
actual:     '#hasheq()
expected:   '#hasheq()
```

Intriguingly, `(check-eq? (hash) (hash))` is ok.

sorawee:
Fixes #1357

### DIFF
--- a/typed-racket-lib/typed-racket/utils/any-wrap.rkt
+++ b/typed-racket-lib/typed-racket/utils/any-wrap.rkt
@@ -126,13 +126,27 @@
        v]
       [(? unsafe-val?)
        (fail neg-party v)]
-      [(cons x y) (cons (any-wrap/traverse x neg-party seen/v) (any-wrap/traverse y neg-party seen/v))]
+      [(cons x y)
+       (define x* (any-wrap/traverse x neg-party seen/v))
+       (define y* (any-wrap/traverse y neg-party seen/v))
+       (cond
+         [(and (eq? x x*) (eq? y y*)) v]
+         [else (cons x* y*)])]
       [(? vector? (? immutable?))
        ;; fixme -- should have an immutable for/vector
-       (vector->immutable-vector
-        (for/vector #:length (vector-length v)
-          ([i (in-vector v)]) (any-wrap/traverse i neg-party seen/v)))]
-      [(? box? (? immutable?)) (box-immutable (any-wrap/traverse (unbox v) neg-party seen/v))]
+       (define all-eq? #t)
+       (define v*
+         (for/vector #:length (vector-length v) ([i (in-vector v)])
+           (define i* (any-wrap/traverse i neg-party seen/v))
+           (unless (eq? i i*)
+             (set! all-eq? #f))
+           i*))
+       (if all-eq? v (vector->immutable-vector v*))]
+      [(? box? (? immutable?))
+       (define v* (any-wrap/traverse (unbox v) neg-party seen/v))
+       (cond
+         [(eq? v* (unbox v)) v]
+         [else (box-immutable v*)])]
       [(? box?) (chaperone-box v
                                (lambda (v e)
                                  (with-contract-continuation-mark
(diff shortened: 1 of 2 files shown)


samth (approved):

samth:
This seems to have broken a bunch of `plot` tests: http://drdr.racket-lang.org/65758/cs/racket/share/pkgs/plot-test/plot/tests/plot3d-bsp-tests.rkt for example.

samth:
cc @alex-hhh

sorawee:
When I compare two traces, it looks like there's an extra

```
 (do-set-pen! ((0 0 0 1.0) 1 transparent round round #f))
 (do-set-brush! ((0 0 0 1.0) solid #f #f #f))
```

Old trace:

```
 (set-alpha 1)
 (do-set-pen! ((0 0 0 1.0) 1 solid round round #f))
 (do-set-brush! ((0 0 0 1.0) solid #f #f #f))




 (draw-ellipse 351.33333333333337 408.2808171026603 4 4)
 (set-alpha 1)
 (set-font (11 #f roman normal normal #f default #f aligned #hash()))
 (set-text-foreground (0 0 0 1.0))
 (set-text-foreground (255 255 255 1.0))
```

New trace:

```
 (set-alpha 1)
 (do-set-pen! ((0 0 0 1.0) 1 solid round round #f))
 (do-set-brush! ((0 0 0 1.0) solid #f #f #f))

 (do-set-pen! ((0 0 0 1.0) 1 transparent round round #f))
 (do-set-brush! ((0 0 0 1.0) solid #f #f #f))

 (draw-ellipse 351.33333333333337 408.2808171026603 4 4)
 (set-alpha 1)
 (set-font (11 #f roman normal normal #f default #f aligned #hash()))
 (set-text-foreground (0 0 0 1.0))
 (set-text-foreground (255 255 255 1.0))
```

sorawee:
Oh, and I confirm that this PR is the cause, specifically due to the change from:

```
[(cons x y) (cons (any-wrap/traverse x neg-party seen/v) (any-wrap/traverse y neg-party seen/v))]
```

to:

```
      [(cons x y)
       (define x* (any-wrap/traverse x neg-party seen/v))
       (define y* (any-wrap/traverse y neg-party seen/v))
       (cond
         [(and (eq? x x*) (eq? y y*)) v]
         [else (cons x* y*)])]
```

If I revert this particular change, the traces are now equal.

sorawee:
Found the cause: 

This line probably should use `equal?` rather than `eq?`.

https://github.com/racket/plot/blob/master/plot-lib/plot/private/common/plot-device.rkt#L549

sorawee:
@alex-hhh let me know if there's anything unclear. Due to https://github.com/racket/plot/issues/126, you are probably the best person to regenerate the fixed test data.

alex-hhh:
Yes, I am looking at this now (wasn't following this discussion, so I was reading through it).   This is indeed a bug in the plot package and I'll fix it.   I just want to check if it is worth changing `eq?` to `equal?` or simply remove the test, since it didn't do anything for so long...

sorawee:
I think the question is more about how important it is that `set-pen` and `set-brush` are called when `pen-color` and `brush-color` are the same:

```
                      (when (eq? pen-color brush-color)
                        (set-pen pen-color 1 'transparent)
                        (set-brush brush-color 'solid))
```

If it is important, I would fix it by changing `eq?` to `equal?`, and regenerate the tests. But if it's just an arbitrary decoration that is not important, you can try removing the whole `when`, which would make the failing tests passed (but might also make other tests failed).

alex-hhh:
That code effectively removes the outline around a glyph, when the pen (outline) and brush (fill) are the same color.  The visual effect of fixing the test would be that glyphs on the plot would appear smaller, especially if users specified a larger width for the outline.  This visual change might be considered a bug or backwards incompatible change...  

I'll need to look into this a bit mode, but I lean towards removing that code...

PR #1373, opened 2024-05-15
Add Unicode Aliases for `U` and `->*`.

NoahStoryM:
- Added `∪` as an alias for `U`.
- Added `→*` as an alias for `->*`.


samth:
Can you add some tests for this? Otherwise, it looks good.

samth:
Thanks!

PR #1377, opened 2024-06-03
doc: guide, show lambda return-type annotation and explain ann for ca…

bennn:
…se-lambda returns

- - -

prompted by a confused friend who knew about `(lambda args : TYPE e)` and got stuck trying to guess the equivalent for `case-lambda`

(another idea is to move the `ann` docs up, so they're first rather than last on this page)

### DIFF
--- a/typed-racket-doc/typed-racket/scribblings/guide/more.scrbl
+++ b/typed-racket-doc/typed-racket/scribblings/guide/more.scrbl
@@ -89,6 +89,10 @@ of @racket[Number]s.
 @racketblock[(lambda ([x : String] (unsyntax @tt["."]) [y : Number #,**]) (apply + y))]
 
 This function has the type @racket[(-> String Number #,** Number)].
+To specify the return type, add a type annotation after the arguments:
+
+@racketblock[(lambda ([x : String] (unsyntax @tt["."]) [y : Number #,**]) : (U Number String) (apply + y))]
+
 Functions defined by cases may also be annotated:
 
 @racketblock[(case-lambda [() 0]
@@ -96,6 +100,10 @@ Functions defined by cases may also be annotated:
 
 This function has the type
 @racket[(case-> (-> Number) (-> Number Number))].
+To specify the return type, either annotate the entire function or use the
+@seclink["Annotating_Expressions" #:doc '(lib "typed-racket/scribblings/ts-guide.scrbl")]{expression annotation form}
+(@racket[ann]) inside each case.
+
 
 @subsection{Annotating Single Variables}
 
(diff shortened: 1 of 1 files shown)


typed-racket-doc/typed-racket/scribblings/guide/more.scrbl
```diff
@@ -89,13 +89,21 @@ of @racket[Number]s.
...
+
 Functions defined by cases may also be annotated:
 
 @racketblock[(case-lambda [() 0]
                           [([x : Number]) x])]
 
 This function has the type
 @racket[(case-> (-> Number) (-> Number Number))].
+To specify the return type, either annotate the entire function or use the
+@seclink["Annotating_Expressions" #:doc '(lib "typed-racket/scribblings/ts-guide.scrbl")]{expression annotation form}
```
sorawee, line 104: I think `#:doc` can be elided here, since this current Scribble file is already within `typed-racket/scribblings/ts-guide.scrbl`?

typed-racket-doc/typed-racket/scribblings/guide/more.scrbl
```diff
@@ -89,13 +89,21 @@ of @racket[Number]s.
 @racketblock[(lambda ([x : String] (unsyntax @tt["."]) [y : Number #,**]) (apply + y))]
 
 This function has the type @racket[(-> String Number #,** Number)].
+To specify the return type, add a type annotation after the arguments:
```
sorawee, line 92: Maybe add the word "explicitly" here, so that it's not read as "if you don't specify the return type, then it doesn't exist"?

typed-racket-doc/typed-racket/scribblings/guide/more.scrbl
```diff
@@ -89,13 +89,21 @@ of @racket[Number]s.
...
+
+@racketblock[(lambda ([x : String] (unsyntax @tt["."]) [y : Number #,**]) : (U Number String) (apply + y))]
```
sorawee, line 94: Should we do `(U Number String)` here or just `Number`?

bennn: Number is boring. I want something other than the default.

typed-racket-doc/typed-racket/scribblings/guide/more.scrbl
bennn: That reading seems unlikely. There's an example function right above with no annotation and we say it returns `Number`.

samth (approved):

PR #1380, opened 2024-06-21
Handle keyword variants of file operations.

samth:
This is to support https://github.com/racket/racket/pull/5020


typed-racket-lib/typed-racket/base-env/base-special-env.rkt
```diff
@@ -5,6 +5,7 @@
 (require "../env/init-envs.rkt"
          "../types/abbrev.rkt"
          "../types/numeric-tower.rkt"
+	 (for-template (submod racket/file proc-names))
```
mflatt, line 8: Would using `make-template-identifier` (like other identifiers) be better? I worry about using `for-template`, because it can easily create a lot more module dependencies and instantiations than intended.

samth: The reason I'm not using `make-template-identifier` is that the identifier I want has the same name as the `define-syntax` binding so I don't think I can reliably pick out the right one.

mflatt: Latest approach makes sense!

Issue #1379, opened 2024-06-19
Since Racket 8.1, some nested lists pass type-checking even when they are invalid 


cloudrac3r:
### What version of Racket are you using?

8.13 [bc]. The issue also occurs in 8.13 [cs] and 8.1 [bc]. **This issue does not occur in 8.0 [bc].**

### What program did you run?

```racket
#lang typed/racket
(define-type Xexpr (List* Symbol (Listof (U String Xexpr))))

(define invalid 0.4)

(: result Xexpr)
(define result
  `(html
    (body
     (p ,(~a invalid)) ; works because it's a String which is allowed by Xexpr type
     (p ,invalid) ; expected a type mismatch because it's a Real which is not allowed by Xexpr type
     (p ,@((inst map Xexpr Xexpr) values (list 0.4))) ; expected a type mismatch because it splices (Listof Real) and Real is not allowed by Xexpr type
     )))

(display (banner))
(println result)
```

### What should have happened?

It should fail typechecking as per the inline comments.

### What actually happens?

It passes typechecking and runs even though the types are actually incorrect.

```
Welcome to Racket v8.13 [cs].
'(html (body (p "0.4") (p 0.4) (p 0.4)))
> (:print-type result)
Xexpr
```

This can cause runtime errors later on, which TR would normally prevent:

```
> (require typed/xml)
> (xexpr->xml result)
xexpr->xml: contract violation;
 Not an Xexpr. Expected a string, symbol, valid numeric entity, comment, processing instruction, or list, given 0.4
```

It also causes a runtime error when forcing a cast to the type that it already is:

```
> (cast result Xexpr)
\share\pkgs\typed-racket-lib\typed-racket\base-env\prims-contract.rkt:487:25: (cast for #f): broke its own contract
  promised: (or/c string? (list/c symbol? (recursive-contract g11 #:flat) (recursive-contract g11 #:flat) (recursive-contract g11 #:flat)) (list/c symbol? (recursive-contract g11 #:flat) (recursive-contract g11 #:flat)) (list/c symbol? (recursive-contract g11 #:flat)) (list/c symbol?))
  produced: '(html (body (p "0.4") (p 0.4) (p 0.4)))
  in: (recursive-contract
        -snipped for brevity- )
  contract from: cast
  blaming: cast
   (assuming the contract is correct)
```

### Investigation

[...]

PR #1382, opened 2024-06-27
Fix type-checking list elements

cloudrac3r:
This addresses a regression in commit 8d9cb49.

That commit uses (when (andmap not errs) (expected-but-got ...)) to suppress some errors when type-checking list elements. Before this, there would be two errors reported: one for the list element itself and one for the surrounding list. However, as per #1379 this change suppressed too much. In certain cases it wouldn't report any errors for an invalid list.

I couldn't figure out a way to report errors perfectly, so instead, I decided to revert the relevant change in 8d9cb49. This means list could report multiple errors again, but it's honestly not that bad. It shows the innermost location where the problem happened, as well as the wider context of the list, which can be helpful for locating and understanding the error. An example using the code from #1379:

```
sample.rkt:13:41: Type Checker: type mismatch
  expected: (Listof Xexpr)
  given: (List Positive-Float-No-NaN)
  in: (list invalid)
  - snip -
sample.rkt:9:2: Type Checker: type mismatch
  expected: Xexpr
  given: (List 'html (List 'body (List 'p String) (Pairof 'p Nothing)))
  in: (quasiquote (html (body (p (unquote (~a invalid))) (p (unquote-splicing ((inst map Xexpr Xexpr) values (list invalid)))))))
  - snip -
Type Checker: Summary: 2 errors encountered
  location...:
   sample.rkt:13:41
   sample.rkt:9:2
  - snip -
```

The performance fix from 8d9cb49 still works.

### DIFF
--- a/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-list.rkt
+++ b/typed-racket-lib/typed-racket/typecheck/tc-app/tc-app-list.rkt
@@ -130,17 +130,8 @@
                   (add-typeof-expr #'op-name (ret (->* arg-tys return-ty :T+ #t)))
                   (ret return-ty)))
               (or result
-                  ;; When arguments to the list function fail to typecheck, the
-                  ;; odds are those argument expressions are ill-typed.  In that
-                  ;; case, we should only report those type errors instead of the error
-                  ;; that the result type of application of list doesn't match
-                  ;; the expected one.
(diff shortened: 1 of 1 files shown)


samth:
Thanks, this looks like a reasonable approach.

sorawee:
@samth should we merge this?

PR #1414, opened 2024-11-26
Fix correspondence between location/c and Location in typed/racket

joergen7:
Checklist

- [x]  Bugfix
- [ ] Feature
- [x] tests included
- [ ] documentation

See also: Issue #1415 

In (untyped) racket, the XML collection uses a `source` struct whose fields `start` and `stop` follow the contract `location/c`. The `location/c` contract is defined as follows:

    (or/c location? symbol? #f)

However, in typed/racket the aforementioned fields are typed as `Location` which represents only the struct but neither symbols nor false.

This pull request extends the typing of `source`'s `start` and `stop` to entail both symbols and false typing them as

    (U Location Symbol False)


joergen7:
Two workflows await approval.

PR #1444, opened 2025-04-14
Improve definition of XExpr type

joergen7:
The type XExpr has been defined as follows:

    (define-type XExpr
      (U XML-Misc Cdata Positive-Index
         Number String Symbol
         (Pair Symbol (Pair (Listof XExpr-Attribute) (Listof XExpr)))
         (Pair Symbol (Listof XExpr))))

The above definition includes the superfluous type `Number` but is missing the type `Pcdata`.

The type `Number` is superfluous because so-called valid characters (`valid-char?`) are already covered by `Positive-Index`. `Pcdata` structs are covered by `xexpr?`, but are missing from the above type definition.


samth:
I worry a little about restricting the space of numbers like this. The change means that `(- 2 1)` no longer typechecks as `XExpr`.

joergen7:
I thought about this a bit longer and have come to the conclusion that Number is,in fact, too wide a definition for the type `XExpr`. I am aware that narrowing a type has practical implications that might break the workflow of some people that might have relied on the fact, that `Number` was part of the definition of `XExpr`. Thanks @samth for pointing at an example for such a case.

However, I would argue that it is, indeed necessary for the type `XExpr` to be congruent with its corresponding predicate `xexpr?`. I would expect that if a value types `XExpr`, then it also satisfies `xexpr?`. And this is not the case under the current definition. The counter-example, I am going with is `0`.

    #lang typed/racket/base

    (require typed/xml)

    (define xe : XExpr
      0)

    (xexpr? x)

produces `#f` which does not settle with me.

joergen7:
I understand. Thanks, @samth for your patience. The current change strictly widens the definition of `XExpr`.

PR #1460, opened 2025-06-27
(WIP) Use `define-deprecated-alias` for legacy aliases

jackfirth:
This is a (very) work-in-progress pull request to use the new `define-deprecated-alias` form to define the various aliases listed in [Legacy Forms](https://docs.racket-lang.org/ts-reference/Legacy_Forms.html). This will allow Resyntax to automatically migrate users of these forms to their recommended replacements.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/prims.rkt
+++ b/typed-racket-lib/typed-racket/base-env/prims.rkt
@@ -52,7 +52,7 @@ the typed racket language.
                      for/product:
                      for/sum:
                      for/vector:
-                     for/and:
+                     for*/and:
                      for*/first:
                      for*/fold:
                      for*/foldr:


jackfirth:
@samth These test failures confound me. Any idea what's going on here?

```
--------------------
Typechecker tests > tc-expr tests > 652 (case-lambda: (((a : Number) (b : Number)) (+ a b)))
ERROR
location:   <pkgs>/typed-racket-test/unit-tests/typecheck-tests.rkt:652:8

typed-racket-test/unit-tests/typecheck-tests.rkt:652:35: type declaration: must be used in a definition context
  in: :
--------------------
```

samth:
That's some of the most fragile code in TR. Is that test using the old alias or the new name?

PR #1499, opened 2026-03-02
Support Racket's qq-and-or.rkt -> core-syntax.rkt file rename

jesboat:
Typed Racket needs to be updated so it's aware of the `qq-append` function
regardless of whether that function is defined in the old place (qq-and-or.rkt)
or the new place (core-syntax.rkt). We can do this through a horrible hack, by
taking advantage of the fact that defines `qq-apend` is being moved wholesale,
so either the `racket/private/qq-and-or` xor `racket/private/core-syntax` will
be declared, and we can determine where `qq-append` is by checking which one.

Once the file rename can be reasonably assumed to be everywhere that somebody
would want to be installing TR's trunk branch, which is to say probably at most
a week after this is committed, we can remove the hack.

Cross-reference: https://github.com/racket/racket/pull/5457

# NOTE coordinated change in Racket

This change only makes sense if https://github.com/racket/racket/pull/5457 is going to be merged. The proper order for things is probably (1) get approval for both PRs, (2) merge the PR to Typed Racket and wait a couple hours for the package server to update, (3) merge this.

### DIFF
--- a/typed-racket-lib/typed-racket/base-env/base-special-env.rkt
+++ b/typed-racket-lib/typed-racket/base-env/base-special-env.rkt
@@ -18,6 +18,19 @@
      (module-path-index-join where #f)))
   (syntax-binding-set->syntax s what))
 
+; Like make-template-identifier, but given multiple modules and picks the
+; first module which appears to exist. This is a giant hack and should be
+; removed when no longer necessary, probably late April 2016 or so. See
+; https://github.com/racket/typed-racket/pull/1499
+(define (make-template-identifier* what where-potentials)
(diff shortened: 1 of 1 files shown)


samth:
Should we just merge this now?

jesboat:
> Should we just merge this now?

I'd been waiting (https://github.com/racket/racket/pull/5457#issuecomment-4083529856) for Matthew to confirm that the rename is still something we want to do before merging the support for it here, but I suppose there's no reason this couldn't be merged now, as it could always be reverted if the rename doesn't go forward.

